PR #6, opened 2018-07-17
Qmaps 388  zoom

Julien-laville:
peek 17-07-2018 15-54

Add zoom on bbox & click (scene / autocomplete)


nlehuby:
The `zoom to` in the autocomplet should have an offset to take into account the panel (just like we already did in the favorite `zoom to`)
zoom_to_offset

PR #8, opened 2018-08-01
Qmaps 404  poi panel

Julien-laville:


nlehuby:
The "service & info" text sometimes goes over the arrow:
screenshot_2018-08-02 qwant maps - beta

nlehuby:
It would be easier to understand if we prefix the beer panel with "beers" just like it is done for the internet access info:
strange:
screenshot_2018-08-02 qwant maps - beta 1
much better:
screenshot_2018-08-02 qwant maps - beta 2

nlehuby:
Can we add some tests ?
* click on the POI name in the panel
* click on the service & info block

nlehuby:
screenshot_2018-08-03 qwant maps - beta 1

a `:`after `Bières` is needed here

nlehuby:
The accessibility panel is sometimes empty (due to https://github.com/QwantResearch/idunn/pull/20)
wheelchair

src/panel/poi_panel.js
```diff
@@ -85,6 +85,17 @@ PoiPanel.prototype.setPoi = async function (poi) {
...
 
+PoiPanel.prototype.copy = function () {
+  let copyUrl = document.getElementById('share-url')
+  copyUrl.select()
+  document.execCommand('copy')
+  copyUrl.blur()
+}
+
+PoiPanel.prototype.center = function() {
+  fire('fit_map', this.poi, {sidePanelOffset : true})
```
amatissart: This does not respect the `ZOOM_BY_POI_TYPES` value.  
This is due to how the poi is initialized in `Poi.apiLoad`. The API does not currently return a `type` and the default zoom value is used. 
I think the type `"poi"` should be set by default when loading a Poi from the API.

src/views/poi_panel.dot
```diff
@@ -1,7 +1,7 @@
 <div class="poi_panel {{= this.active ? '' : 'poi_panel--hidden' }}">
   <div class="poi_panel__content">
   {{? this.poi }}
-    <div class="poi_panel__description_container">
+    <div class="poi_panel__description_container" {{= click(this.center, this) }}>
```
amatissart: There is currently no way for the user to notice that a click has an effect here.  
Maybe we could simply change the cursor on hover ?

src/panel/poi_panel.js
Julien-laville: fixed

src/views/poi_panel.dot
Julien-laville: fixed

PR #11, opened 2018-08-06
Qmaps 406  favorite

Julien-laville:


src/views/favorites_panel.dot
```diff
@@ -13,20 +12,19 @@
...
       {{= this.filterPanel.render() }}
-
       <ul>
         <li class="favorite_panel__item">
 
           <div class="favorite_panel__swipe_window">
             <div class="favorite_panel__swipe_flow--{{= this.favoritePois.length }}">
               {{~ this.favoritePois:poi:i }}
-                <div class="favorite_panel__swipe_element" id="{{= `favorite_item_${i}` }}">
+                <div class="favorite_panel__swipe_element" {{= click(this.go, this, poi) }} id="{{= `favorite_item_${i}` }}">
```
amatissart: `cursor: pointer;` would be nice here too

src/views/side_bar_panel.dot
```diff
@@ -2,7 +2,7 @@
 
   <div class="side_bar__logo"></div>
   <div class="side_bar__logo__limit"></div>
-  <div class="side_bar__result side_bar__item  {{= this.current === 0 ? 'side_bar__item--active': ''}}">
+  <div {{= click(this.toggleResult, this) }} class="side_bar__result side_bar__item {{= this.current === 0 ? 'side_bar__item--active': ''}}">
```
amatissart: Trigger "toggleResult" on this click is not enough, as the poi panel could be toggled via other actions.

src/views/favorites_panel.dot
Julien-laville: fixed

PR #14, opened 2018-08-21
Qmaps 407

Julien-laville:
Update storage id with dynamic version
Add helper to pre-store poi on test (tool.store)


nlehuby:
The test with the panel of the poi already in favorite needs to be updated too: https://github.com/QwantResearch/erdapfel/blob/master/tests/integration/tests/poi.js#L61

PR #15, opened 2018-08-23
Push i18n messages to transifex on merge on master

nlehuby:


.travis.yml
```diff
@@ -13,3 +14,20 @@ addons:
...
+    provider: script
+    skip_cleanup: true
+    script:
+        - npm run build
+        - npm run i18n
+        - pip install virtualenv
+        - virtualenv ~/env
+        - source ~/env/bin/activate
+        - pip install transifex-client
+        - sudo echo $'[https://www.transifex.com]\nhostname = https://www.transifex.com\nusername = '"$TRANSIFEX_API_USER"$'\npassword = '"$TRANSIFEX_API_KEY"$'\n' > ~/.transifexrc
```
Julien-laville: We should check a lightweight way to make the job, maybe check TransifleX api and call it with system call (curl) or simple script

amatissart: It looks good to me. Maybe we could use `sudo pip install` directly (without the virtualenv), but that's a detail.

amatissart:
What would the workflow look like exactly ?  
As far as I understand, when developing new features, the new translation keys are pushed to transifex on merge in "master". So we need to create a new PR with updates from transifex in a second phase ?

PR #21, opened 2018-09-14
add a test on adding a poi to the favorite

antoine-de:
Add one small test.

It is currently not that easy to add tests, I think we should add more a
straction to make it easier.

### DIFF
--- a/tests/integration/tests/poi.js
+++ b/tests/integration/tests/poi.js
@@ -2,7 +2,7 @@ const configBuilder = require('@qwant/nconf-builder')
 const config = configBuilder.get()
 const APP_URL = `http://localhost:${config.PORT}`
 const poiMock = require('../../__data__/poi')
-import {initBrowser, getText, wait, clearStore} from '../tools'
+import {initBrowser, getText, wait, clearStore, openFavoritePanel} from '../tools'
 
 let browser
 let page
@@ -63,7 +63,6 @@ test('load a poi already in my favorite from url', async () => {
   })
   await page.goto(`${APP_URL}/place/osm:node:2379542204@Musée_dOrsay#map=17.49/2.3261037/48.8605833`)
   let plainStar = await page.waitForSelector('.icon-icon_star-filled')
-  clearStore(page)
   expect(plainStar).not.toBeFalsy()
 })
 
@@ -94,7 +93,6 @@ test('update url after a favorite poi click', async () => {
   let location = await page.evaluate(() => {
     return document.location.href
   })
-  clearStore(page)
   expect(location).toMatch(/1@Mus%C3%A9e_dOrsay/)
 })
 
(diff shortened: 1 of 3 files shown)


nlehuby:
@antoine-de is developping on the frontend :heart_eyes_cat:

tests/integration/tests/poi.js
```diff
@@ -178,6 +165,50 @@ test('display details about the poi on a poi click', async () => {
...
+  // we check that the first favorite item is our poi
+  await toggleFavoritePanel(page)
+  let fav = await getFavorites(page)
+  expect(fav).toHaveLength(1)
+  expect(fav[0].title).toEqual("Musée d'Orsay")
+  expect(fav[0].desc).toEqual("musée")
+  expect(fav[0].icons).toContainEqual("icon-museum")
+
+  // we then reopen the poi panel and 'unstar' the poi.
+  await wait(100)
```
antoine-de: I don't understand why this wait is needded (the `toggleFavoritePanel` function wait `300` for the end of the animation) but If I remove it the test fail :grin:

tests/integration/tests/poi.js
```diff
@@ -178,6 +165,50 @@ test('display details about the poi on a poi click', async () => {
...
+  })
+  await page.click('#mock_poi')
+  await wait(300)
+}
+
+test('add a poi as favorite and find it back in the favorite menu', async () => {
+  expect.assertions(7)
+  await page.goto(APP_URL)
+
+  // we select a poi and 'star' it
```
Julien-laville: Should be on an atomic test

tests/integration/tests/poi.js
```diff
@@ -178,6 +165,50 @@ test('display details about the poi on a poi click', async () => {
...
+  await select_poi_1(page)
+  let poiPanel = await page.waitForSelector('.poi_panel__title')
+  expect(poiPanel).not.toBeFalsy()
+  await page.click('.poi_panel__actions__icon__store')
+
+  // we check that the first favorite item is our poi
+  await toggleFavoritePanel(page)
+  let fav = await getFavorites(page)
+  expect(fav).toHaveLength(1)
```
Julien-laville: maybe we can use only one comparison with deepEquality

tests/integration/tools.js
```diff
@@ -37,4 +37,24 @@ export function clearStore(page) {
...
+
+export async function toggleFavoritePanel(page) {
+  await page.waitForSelector('.icon-icon_star.side_bar__item__icon')
+  await page.click('.icon-icon_star.side_bar__item__icon')
+  await wait(300)
+}
+
+// Returns all favorites as an array
+// It needs the favorite panel to be open
+export async function getFavorites(page) {
```
Julien-laville: Maybe it's too soon but we should abstract some helper by context (Fav, Poi etc) 
This one is too specific to fit inside "tool"

tests/integration/tests/poi.js
antoine-de: I don't really know how to do it since I want to test the favorite add and the favorite deletion.

Do you see an easy way to do it ?

tests/integration/tests/poi.js
antoine-de: It's a bit complicated since we also want to test only that contains "icon-museum"

tests/integration/tools.js
antoine-de: so you want me to move this function to a `favorite_tools.js` ?

PR #22, opened 2018-09-18
Qmaps 419  metadata

Julien-laville:
Add og:meta to home page (description, logo, title, locale, image, site_name)


bin/middlewares/og_meta.js
```diff
@@ -0,0 +1,91 @@
+const request = require('request')
+const acceptLanguage = require('accept-language')
+const QueryError = require('../query_error')
+
+const metas = [
+  {name : 'site_name', content : 'Qwant map'},
```
amatissart: "Qwant map" :arrow_right: "Qwant Maps"

bin/middlewares/og_meta.js
```diff
@@ -0,0 +1,91 @@
...
+  {name : 'image', content : '/images/maps_opengraph.png'},
+  {name : 'url', content : 'www.qwant.com/maps'},
```
amatissart: Could the url be set dynamically based on the request ?

bin/middlewares/og_meta.js
```diff
@@ -0,0 +1,91 @@
...
+    res.locals.metas.push({name : 'title', content : `${res.locals. _('Qwant map')} | ${poi.name}`})
+    next()
+  }
+  function homeMeta(locale, req, res, next) {
+    commonMeta(locale, req, res)
+    res.locals.metas.push({name : 'title', content : `${res.locals. _('Qwant map')}`})
+    next()
+  }
+
+  function getLocale(req) {
```
amatissart: This function is not specific to opengraph and is maybe redundant with the existing `user_language.js` middleware. 
It should be refactored there.

bin/middlewares/og_meta.js
```diff
@@ -0,0 +1,91 @@
...
+          resolve(null)
+        } else {
+          resolve(body)
+        }
+      })
+    })
+  }
+
+  function commonMeta(locale, req, res) {
+    res.locals.metas = metas.map(meta => meta)
```
amatissart: maybe `ogMetas` instead of `metas` in locals ?

bin/middlewares/og_meta.js
```diff
@@ -0,0 +1,91 @@
...
+    res.locals.metas.push({name : 'locale', content : locale.locale})
+    res.locals.metas.push({name : 'description', content : res.locals. _('Map multiple locations. Do more with Qwant Maps.')})
+  }
+
+  function poiMeta(poi, locale, req, res, next) {
+    commonMeta(locale, req, res)
+    res.locals.poi = poi
+    res.locals.metas.push({name : 'title', content : `${res.locals. _('Qwant map')} | ${poi.name}`})
```
amatissart: The poi name is probably enough in the title, as "Qwant Maps" is already defined as the site name

bin/middlewares/og_meta.js
Julien-laville: done

bin/middlewares/og_meta.js
Julien-laville: Je veux pas le faire car non trivial 
l'url se genere avec `${req.protocol}://${req.get('host')}` mais alors il manque la query (/maps)
mais si j'ajoute queryUrl alors on aura aussi /place à la suite

bin/middlewares/og_meta.js
Julien-laville: cleaned

bin/middlewares/og_meta.js
Julien-laville: refactored

bin/middlewares/og_meta.js
Julien-laville: ok but i keep qwant map as title if there is no poi

tests/integration/tests/poi.js
```diff
@@ -32,6 +32,7 @@ test('click on a poi', async () => {
   await page.goto(APP_URL)
   await page.evaluate(() => {
     window.MAP_MOCK.evented.prepare('click', 'poi-level-1',  {originalEvent : {clientX : 1000},features : [{properties :{global_id : 1}}]})
+    window.MAP_MOCK.evented.prepare('click', 'poi-level-1',  {originalEvent : {clientX : 1000},features : [{properties :{global_id : 1}}]})
```
amatissart: Copy paste error ?

amatissart: Copy paste error ?

bin/middlewares/og_meta.js
amatissart: I am not sure. We also need to set a canonical url when the page is about a specific poi.

tests/integration/tests/poi.js
Julien-laville: fixed with rebase

PR #32, opened 2018-10-09
Fine tuning design.

Julien-laville:


nlehuby:
We don't need the red border (when the search field has been cleared with the cross):
screenshot_2018-10-16 sushi shop poincare

capture d ecran - 2018-10-16 a 13 58 07

PR #33, opened 2018-10-10
Add an endpoint to receive client events

amatissart:
A simple endpoint to receive and log events.

### DIFF
--- a/bin/app.js
+++ b/bin/app.js
@@ -57,6 +57,20 @@ function App(config) {
       res.render('index', {config : config})
   })
 
+  if(config.system.acceptPostedEvents){
+    app.post('/events',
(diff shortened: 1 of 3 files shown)


bin/metrics_builder.js
```diff
@@ -0,0 +1,31 @@
...
+
+const eventTypes = [
+  'favorite_saved'
+]
+
+module.exports = (app, config, registry) => {
+  const counters = {}
+  eventTypes.forEach(eventType => {
+    counters[eventType] = new promClient.Counter({
+      name: `erdapfel_${eventType}`,
```
antoine-de: maybe add `_count` like it's done elsewhere ?

PR #34, opened 2018-10-15
Update translations strings and map style

nlehuby:
* Fix a translation key
* Pull translations from transifex
* Get back last style improvements (boundaries, tunnels and i18n labels)


language/message/fr_FR.po
```diff
@@ -929,12 +932,12 @@ msgid "Back to favorite"
 msgstr "Retour aux favoris"
 
 #: public/build/javascript/bundle.js:2556
-msgid "FAVORIS"
-msgstr ""
+msgid "FAVORITES"
+msgstr "FAVORIS"
```
amatissart: The translation key is to change here.  
This change needs to be updated in `poi_panel.dot` and to be synchronized with the other languages.

package.json
```diff
@@ -30,7 +30,7 @@
     "@qwant/merge-po-loader": "file:local_modules/merge-po-loader",
     "@qwant/nconf-getter": "file:local_modules/nconf_getter",
     "@qwant/po-js-loader": "file:local_modules/po-js-loader",
-    "@qwant/qwant-basic-gl-style": "git+https://github.com/QwantResearch/qwant-basic-gl-style.git#ab4b6c0",
+    "@qwant/qwant-basic-gl-style": "git+https://github.com/QwantResearch/qwant-basic-gl-style.git#07a7bfca697f98d69ce7373c17b1f9ba9df47574",
```
amatissart: I think we should pin the version of `map-style-builder` too.

PR #41, opened 2018-10-22
QMAPS-535 rel noopener

patfrat:
Add noreferre and noopener to href with target="_blank"


Julien-laville:
missing : 
- phone.dot:5
- poi_panel.dot:55

patfrat:
Missing files added

PR #42, opened 2018-10-22
QMAPS-517 Filter local storage on POIs for the favorites panel

patfrat:
Add isCompliantKey method on Poi in order to take in account only POIs in Favorites panel

### DIFF
--- a/src/adapters/store.js
+++ b/src/adapters/store.js
@@ -19,30 +19,24 @@ function Store() {
 }
 
 Store.prototype.getAllPois = async function() {
-  return new Promise((resolve, reject) => {
-    abstractStore.getAllPois().then((masqData) => {
-      resolve(masqData)
-    }).catch(function (error) {
-      fire('error_h' , 'store ' + error)
-      reject(error)
-    })
-  })
+  try {
+    return await abstractStore.getAllPois()
+  } catch (error) {
+    fire('error_h' , 'store ' + error)
+    throw error
+  }
 }
 
 Store.prototype.isRegistered = async function () {
-  return new Promise((resolve) => {
-    abstractStore.getAllPois()
-      .then(() => resolve(true))
-      .catch((e) => {
-      if(e.message === 'UNREGISTERED') {
-        resolve(false)
-      }
-    })
-  })
+  try {
+    return await abstractStore.getAllPois() !== null
+  } catch (error) {
+    return false
+  }
 }
 
 Store.prototype.onConnect = async function () {
-  return abstractStore.onConnect()
+  return await abstractStore.onConnect()
 }
 
 Store.prototype.register = async function () {
@@ -56,7 +50,6 @@ Store.prototype.register = async function () {
   return abstractStore.registerApp(regParams)
 }
 
-
 Store.prototype.getPrefixes = async function (prefix) {
   const storedItems = await abstractStore.getAll()
   return storedItems.filter((storedItem) => {
@@ -66,14 +59,12 @@ Store.prototype.getPrefixes = async function (prefix) {
 }
 
 Store.prototype.has = async function(poi) {
-  return new Promise((resolve) => {
-    abstractStore.get(poi.getKey()).then((foundPoi) => {
-      resolve(foundPoi)
-    }).catch((err) => {
-      fire('error_h', 'store ' + err )
-      resolve()
-    })
-  })
+  try {
+    return await abstractStore.get(poi.getKey())
+  } catch (error) {
+    fire('error_h', 'store ' + err )
+    throw err
+  }
 }
 
 Store.prototype.add = function(poi) {
--- a/src/adapters/store.js
+++ b/src/adapters/store.js
@@ -18,9 +18,9 @@ function Store() {
   })
 }
 
-Store.prototype.getAll = async function() {
+Store.prototype.getAllPois = async function() {
   return new Promise((resolve, reject) => {
-    abstractStore.getAll().then((masqData) => {
+    abstractStore.getAllPois().then((masqData) => {
       resolve(masqData)
     }).catch(function (error) {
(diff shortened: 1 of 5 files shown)


src/panel/favorites_panel.js
```diff
@@ -97,13 +97,21 @@ Favorite.prototype.connectStore = async function () {
 
 Favorite.prototype.getAll = async function () {
   let storedData = {}
+  let poisData = []
   try {
     storedData = await this.store.getAll()
+    // check if all data are Pois
+    Object.keys(storedData).map((mapPoint) => {
```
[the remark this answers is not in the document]
patfrat: ok, i will create a method in local_store > getAllPOIs

src/mapbox/poi.js
```diff
@@ -30,6 +30,11 @@ Poi.prototype.getKey = function () {
   return `qmaps_v${version}_favorite_place_${this.id}`
 }
 
+Poi.isPoiCompliantKey = function (k) {
+  const keyPattern = /qmaps_v[0-9]_favorite_place_*/
```
Julien-laville: You should use the version number from config, it's already imported in this file (import {version} from '../../config/constants.yml')

src/libs/local_store.js
```diff
@@ -1,9 +1,11 @@
...
 
-LocalStore.prototype.getAll = async function () {
+LocalStore.prototype.getAllPois = async function () {
   return new Promise((resolve) => {
     try {
-      const items = Object.keys(localStorage).map((k) => {
+      const items = Object.keys(localStorage).filter((k) => {
+        return Poi.isPoiCompliantKey(k)
+      }).map((k) => {
         return JSON.parse(localStorage.getItem(k))
```
Julien-laville: try catch is too large, it may hide others error than json parse error.

src/libs/local_store.js
```diff
@@ -1,9 +1,11 @@
 function LocalStore() {}
 
-LocalStore.prototype.getAll = async function () {
+LocalStore.prototype.getAllPois = async function () {
```
Julien-laville: i'm not sure but when you declare a function async it already return promise. i think you can simplify by just Promise.resolve(JSON.parse...)

src/adapters/store.js
```diff
@@ -18,9 +18,9 @@ function Store() {
   })
 }
 
-Store.prototype.getAll = async function() {
+Store.prototype.getAllPois = async function() {
   return new Promise((resolve, reject) => {
-    abstractStore.getAll().then((masqData) => {
+    abstractStore.getAllPois().then((masqData) => {
```
Julien-laville: i have moved from then to async await syntax. can you update this part with the new syntax

src/mapbox/poi.js
patfrat: ok for that with RegExp

src/libs/local_store.js
patfrat: The async declaration is not from my changes but was already here ?!

src/adapters/store.js
patfrat: I prefer to refactor the code in another context rather than do it here

src/libs/local_store.js
Julien-laville: i say that i think that if you declare method with async, you dont have to return a new Promise

src/adapters/store.js
```diff
@@ -18,31 +18,29 @@ function Store() {
...
-      }
-    })
-  })
+  try {
+    let data = await abstractStore.getAllPois()
+    return data != null
+  } catch (error) {
+    if(error.message === 'UNREGISTERED') {
+      return false
+    }
```
Julien-laville: this method let an possible unsolvable promise (in the case of error.message != 'unregistred') maybe we can delete the message test and return false every time we enter in the catch.

src/libs/local_store.js
```diff
@@ -14,18 +17,18 @@ LocalStore.prototype.getAll = async function () {
   })
 }
 
-LocalStore.prototype.register = async function() {
+LocalStore.prototype.register = function() {
   console.log('local storage doesn\'t support register method')
   return new Promise((resolve)=>{resolve()})
```
Julien-laville: can be smplified ( new Promise((resolve)=>{resolve()})  = Promise.resolve() )

src/libs/local_store.js
```diff
@@ -14,18 +17,18 @@ LocalStore.prototype.getAll = async function () {
...
 }
 
-LocalStore.prototype.onConnect = async function () {
+LocalStore.prototype.onConnect = function () {
   return new Promise((resolve) => {
     resolve()
```
Julien-laville: can be simplified by return Promise.resolve()

tests/integration/tests/favorite.js
```diff
@@ -38,7 +39,7 @@ test('restore favorite from localStorage', async () => {
   expect.assertions(1)
   await page.goto(APP_URL)
   const testTitle = 'demo_fav'
-  store(page, 'demo_fav', {id : 0, name : testTitle})
+  store(page, 'qmaps_v' + version + '_favorite_place_demo_fav', {id : 0, name : testTitle})
```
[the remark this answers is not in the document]
patfrat: ok

src/adapters/store.js
```diff
@@ -18,31 +18,29 @@ function Store() {
...
-      resolve(masqData)
-    }).catch(function (error) {
-      fire('error_h' , 'store ' + error)
-      reject(error)
-    })
-  })
+Store.prototype.getAllPois = async function() {
+  try {
+    let masqData = await abstractStore.getAllPois()
+    return masqData
```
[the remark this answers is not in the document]
patfrat: right !

Julien-laville:
Thank's @patfrat !
🌈 🦉

PR #44, opened 2018-10-23
Protect Poi data against XSS

Julien-laville:


patfrat:
Does favorites_panel.dot file can be included too to this commit ?

Julien-laville:
🍾

Julien-laville:
merci @patfrat

PR #51, opened 2018-10-30
Qmaps 549 autocomplete bug fix

Julien-laville:
- fix cache
- fix submit form handling


src/vendors/autocomplete.js
```diff
@@ -126,6 +127,7 @@ var autoComplete = (function(){
       addEvent(that, 'blur', that.blurHandler);
 
       var suggest = function(data, queryTerm){
+        console.log(data, queryTerm)
```
patfrat: To remove ?

Julien-laville: haaaaaaa :'(

PR #52, opened 2018-10-30
complete French translation

nlehuby:


nlehuby:
I only left one message untranslated:
https://github.com/QwantResearch/erdapfel/blob/50ea7a4f05661158c611ed831e4beadb34a2bf59/language/message/fr_FR.po#L897

I can't find out where it is displayed :thinking:

Julien-laville:
la chaine que tu cherche provient de la doc, gettext parse un commentaire :)

PR #54, opened 2018-10-31
move autocomplete request cancellation to suggest block (this way cac…

Julien-laville:
…he response can cancel remote request)


src/vendors/autocomplete.js
```diff
@@ -216,14 +225,10 @@ var autoComplete = (function(){
...
               that.timer = setTimeout(function(){
-                if(that.sourcePending) {
-                  that.sourcePending.abort();
-                  that.sourcePending = null
-                }
                 that.sourcePending = o.source(val);
                 that.sourcePending.then((source) => {
-                  suggest(source, val);
                   that.sourcePending = null
+                  suggest(source, val);
```
amatissart: `suggest` should NOT be called here if `source == null` (= if the request has been aborted)

Julien-laville: Seems to work properly.

PR #59, opened 2018-11-06
Qmaps 507 metrics

Julien-laville:
Add common telemetry event list
Create front telemetry class
Add telemetry to
-poi
-fav
-share
-service


package-lock.json
```diff
@@ -991,7 +991,7 @@
     },
     "@sinonjs/formatio": {
       "version": "2.0.0",
-      "resolved": "http://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
```
amatissart: npm warns with low severity vulnerabilities. To fix with `npm audit fix`

src/modals/share.js
```diff
@@ -31,7 +32,12 @@ export default class Share {
     this.modal.close()
   }
 
-  openPopup () {
+  openPopup (targetName) {
+    if(targetName === 'facebook') {
+      Telemetry.add(Telemetry.SHARE_FACEBOOK)
```
amatissart: These events are triggered on each share modal opening and closing, and on app initialization :upside_down_face:

src/mapbox/extended_geolocate_control.js
```diff
@@ -21,7 +30,10 @@ class ExtendedGeolocateControl extends GeolocateControl {
 
   trigger() {
     window.navigator.permissions && window.navigator.permissions.query({ name: 'geolocation' }).then(p => {
-      if (p.state === 'prompt') {
+      if(p.state === geolocationPermissions.GRANTED) {
+        Telemetry.add(Telemetry.LOCALISE_START)
```
amatissart: This event will not be triggered when prompt is used (on the first visit for example)

src/libs/telemetry.js
```diff
@@ -0,0 +1,23 @@
+import Ajax from './ajax'
+import nconf from '@qwant/nconf-getter'
+import telemetryModule from 'telemetry'
+
+const telemetry = nconf.get().telemetry
+
+export default class Telemetry {
+  constructor() {}
+
+  static add(event) {
```
amatissart: Maybe we could find a more explicit name for this function ? `trigger` ?

src/libs/telemetry.js
```diff
@@ -0,0 +1,23 @@
...
+    return Telemetry.send(event)
+  }
+
+  static async send(event) {
+    let data = {type : event}
+    let telemetryUrl = `${telemetry.eventEndpoint}`
```
amatissart: What about `baseUrl` ? I think we need to add it as a prefix here.

config/default_config.yml
```diff
@@ -47,3 +47,7 @@ mapPlugins:
 
 statics:
   maxAge: 0  # string accepted by the ms module, or milliseconds
+
+telemetry:
+  logsEndpoint: '/logs'
+  eventEndpoint: '/events'
```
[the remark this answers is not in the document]
Julien-laville: i keep telemtry enbled inside config & move endPoints to telemetry file as const

package-lock.json
Julien-laville: fixed

src/modals/share.js
Julien-laville: i dont reproduce on app init @amatissart

src/panel/poi_panel.js
```diff
@@ -68,6 +69,7 @@ PoiPanel.prototype.close = async function() {
 }
 
 PoiPanel.prototype.restorePoi = async function (id) {
+  Telemetry.add(Telemetry.POI_RESTORE)
```
amatissart: `poi_restore` event is undefined.  
This causes a request without body to backend, that returns 400. 
I think `Telemetry.add` should validate the event, and log invalid events instead of triggering an invalid request.

src/mapbox/extended_geolocate_control.js
```diff
@@ -1,12 +1,24 @@
...
+  GRANTED : 'granted',
+  DENIED : 'denied'
+}
+
 class ExtendedGeolocateControl extends GeolocateControl {
   constructor(options, container) {
     super(options)
     this._container = container
+    this.on('trackuserlocationstart', () => {
+      Telemetry.add(Telemetry.LOCALISE_TRIGGER)
```
amatissart: I am not sure this is a very relevant metric. If location is refused or blocked for some reason, we'd still count each tap on the button.  
Why not triggering this event on the first click only ?

Julien-laville: i propose a new telemetry method to complete add() : addOnce().

src/panel/poi_panel.js
Julien-laville: good idea

config/default_config.yml
```diff
@@ -47,3 +47,5 @@ mapPlugins:
 
 statics:
   maxAge: 0  # string accepted by the ms module, or milliseconds
+
+telemetryEnabled: true
```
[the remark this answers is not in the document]
Julien-laville: i prefer keep it simple for now, and update to nested the day we have multiple configuration

src/libs/telemetry.js
```diff
@@ -0,0 +1,28 @@
...
+
+  static add(event) {
+    return Telemetry.send(event)
+  }
+
+  static async send(event) {
+    if(telemetryEnabled) {
+      let data = {type : event}
+      let telemetryUrl = `${system.baseUrl}${telemtryEventUrl}`
+      console.log(telemetryUrl)
```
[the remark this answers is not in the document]
Julien-laville: done

src/adapters/autocomplete.js
```diff
@@ -36,7 +36,7 @@ function SearchInput(tagSelector) {
       let isAbort = false
       let promise = new Promise((resolve, reject) => {
         /* 'bbox' is currently not used by the geocoder, it' will be used for the telemetry. */
-        this.suggestPromise = ajax.query(geocoderUrl, {q: term})
+        this.suggestPromise = ajax.get(geocoderUrl, {q: term})
```
amatissart: Another `ajax.query` is to change on line 95 ;)

config/default_config.yml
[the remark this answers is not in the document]
amatissart: I will insist on that one :)  
We will need other settings about logging of telemetry events very soon.  
So I'd like to keep backward compatible configuration on the different environments.

PR #61, opened 2018-11-07
QMAPS-552 Add user language code to learn osm link

patfrat:
Add language code to learn osm external link


amatissart:
I think we need to keep a whitelist of languages that are supported by learnosm.org
For example a user that uses breton language will find an error: https://learnosm.org/br/

patfrat:
Not really an error, it displays a page with available languages that the user can choose then

Julien-laville:
I agree for the simple solution for now.

PR #62, opened 2018-11-07
WIP QMAPS-510 Set local_store as default system storage in configuration

patfrat:
local_store as default system storage in configuration


Julien-laville:
i close it for now. or it will stay here forever : s

PR #63, opened 2018-11-07
QMAPS-565 Quick fix for browser not supporting navigator.permissions

patfrat:
Quick fix for browser not supporting yet navigator.permissions like Safari
See https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query


src/mapbox/geolocation_control.js
```diff
@@ -20,7 +20,7 @@ class GeolocationControl extends GeolocateControl {
   }
 
   trigger() {
-    window.navigator.permissions.query({ name: 'geolocation' }).then(p => {
+    window.navigator.permissions && window.navigator.permissions.query({ name: 'geolocation' }).then(p => {
```
Julien-laville: can you revert this file & amend to avoid future conflict, it's an old artefact that is deleted in another branch.

Julien-laville:
Perfect thank's

PR #74, opened 2018-11-15
update map style

nlehuby:
update of the map graphical chart


Julien-laville:
capture d ecran_2018-11-15_17-39-39
perfect for me (are we sure about the purple ?)

PR #75, opened 2018-11-15
poi card format for mobile

Julien-laville:
Add card for poi on mobile.
capture d ecran_2018-11-15_17-44-53

### DIFF
--- a/src/panel/poi_bloc/opening_minimal.js
+++ b/src/panel/poi_bloc/opening_minimal.js
@@ -12,6 +12,7 @@ export default class MinimalHourPanel {
 
   async set(poi) {
     let openingBlock = poi.blocks.find((block) => block.type === 'opening_hours')
+    this.opening = null
     if(openingBlock) {
       this.opening = new OsmSchedule(openingBlock, this.messages)
     }
--- a/src/panel/poi_bloc/opening_minimal.js
+++ b/src/panel/poi_bloc/opening_minimal.js
@@ -14,7 +14,7 @@ export default class MinimalHourPanel {
     let openingBlock = poi.blocks.find((block) => block.type === 'opening_hours')
     if(openingBlock) {
       this.opening = new OsmSchedule(openingBlock, this.messages)
-      return this.panel.update()
     }
+    return this.panel.update()
   }
 }
--- a/src/panel/poi_bloc/opening_minimal.js
+++ b/src/panel/poi_bloc/opening_minimal.js
@@ -1,0 +1,20 @@
+import HourPanelView from '../../views/poi_bloc/hour_next_transition_partial.dot'
+import Panel from "../../libs/panel";
+import OsmSchedule from '../../../src/adapters/osm_schedule'
+import constants from '../../../config/constants.yml'
+
+export default class MinimalHourPanel {
+  constructor() {
+    this.messages = constants.pois.find((poiConfig) => poiConfig.apiName === 'opening_hours').options.messages
+    this.panel = new Panel(this, HourPanelView)
+    this.opening = null
+  }
+
+  async set(poi) {
+    let openingBlock = poi.blocks.find((block) => block.type === 'opening_hours')
+    if(openingBlock) {
+      this.opening = new OsmSchedule(openingBlock, this.messages)
+      return this.panel.update()
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


src/panel/poi_bloc/opening_minimal.js
```diff
@@ -0,0 +1,20 @@
+import HourPanelView from '../../views/poi_bloc/hour_next_transition_partial.dot'
+import Panel from "../../libs/panel";
```
patfrat: Copy pasta ? 
import "..."; OR import '...'

src/panel/poi_panel.js
```diff
@@ -5,7 +5,8 @@ import PoiBlocContainer from './poi_bloc/poi_bloc_container'
 import Poi from "../mapbox/poi"
```
patfrat: change it to import '...' instead of import "..."

src/scss/includes/panels/poi_panel.scss
```diff
@@ -1,3 +1,5 @@
+@import "../osm_contribute";
```
patfrat: import "..."; OR import '...'

Julien-laville: i don't fix automatic import syntax for now

src/panel/poi_bloc/opening_minimal.js
```diff
@@ -0,0 +1,20 @@
...
+    this.panel = new Panel(this, HourPanelView)
+    this.opening = null
+  }
+
+  async set(poi) {
+    let openingBlock = poi.blocks.find((block) => block.type === 'opening_hours')
+    if(openingBlock) {
+      this.opening = new OsmSchedule(openingBlock, this.messages)
+      return this.panel.update()
+    }
```
amatissart: After the panel has been set, it will persist on POIs without opening_hours (with wrong values) on the next `set` call.  
I suppose `this.opening` should be reset if openingBlock does not exist.

src/scss/includes/panels/poi_panel.scss
```diff
@@ -16,6 +18,31 @@ $BLOCK_INDENT: 60px;
   display: none;
 }
 
+.poi_panel__content__card {
+  display: none;
+}
+
+.poi_panel__content__card__content {
+  float: left;
+  width: calc(100vw - (16px + 130px + 8px));
```
amatissart: ouch  :confused: , please use constants here
Where do these values come from ? And how are they related to other changes ?

src/scss/includes/map_theme.scss
```diff
@@ -111,9 +111,10 @@
     background: rgba(255, 255, 255, 0.5);
   }
 
-  .map_control_group {
+  .mapboxgl-ctrl-group.map_control_group {
```
amatissart: This change seems to break the compass button on mobile:
image

src/scss/includes/panels/poi_panel.scss
```diff
@@ -16,6 +21,31 @@ $BLOCK_INDENT: 60px;
   display: none;
 }
 
+.poi_panel__content__card {
+  display: none;
+}
+
+.poi_panel__content__card__content {
+  float: left;
+  width: calc(100vw - (2 * $CARD_PADDING + $CARD_ACTION_WIDTH + $CARD_ACTION_RIGHT_MARGIN));
```
amatissart: As you said, you may need interpolation syntax to make variables work here
See https://stackoverflow.com/a/20236515/5634396

Julien-laville: Thanks for the information : ]
falso it's fixed

src/scss/includes/panels/poi_panel.scss
```diff
@@ -400,12 +431,48 @@ $BLOCK_INDENT: 60px;
 
 @media (max-width: 640px) {
   .poi_panel {
-    max-height: calc(100vh - 200px);
+    max-height: calc(100vh - 100px);
```
amatissart: On chrome mobile, "100vh" includes the address bar, which makes the panel too high, and difficult to close.

Julien-laville: @nlehuby what do we do ? is add 30px is sufficient ?

PR #79, opened 2018-11-19
update link behaviour on share modal

Julien-laville:
peek 19-11-2018 17-39


src/scss/includes/modals/share.scss
```diff
@@ -26,20 +26,53 @@
...
+}
+
+.share_copy__input__container {
+  position: relative;
 }
+
+.share_copy__container { /* show link */
+  position: absolute;
+  top: 1px;
+  right: 2px;
```
patfrat: right: 4 px;looks better

patfrat: top: 2px; too

Julien-laville: agreed

PR #81, opened 2018-11-27
add localised style hosting for qwant IA

Julien-laville:
add pre-built style for qwant services available at /style.json
lang param can be passed as ?lang={2 letters lang code}


bin/middlewares/map_style.js
```diff
@@ -0,0 +1,40 @@
...
+  let builtStyle = styleBuilder(qwantStyle, options)
+
+  return function (req, res) {
+    /* get lang routine */
+    let langParameter = req.query.lang
+    let matchedLanguage = languages.defaultLanguage
+    if(langParameter) {
+      matchedLanguage = languages.supportedLanguages.find((supportedLanguage) => supportedLanguage.code === langParameter)
+      if(!matchedLanguage) {
+        res.send(400)
```
amatissart: `res.send` with a status code is deprecated.  
And it is actually an error to call `res.setHeader` after `res.send` so we should return here.  
Maybe we could include an explicit json error in the response too.

bin/middlewares/map_style.js
```diff
@@ -0,0 +1,40 @@
...
+  /* pre-build style on server start */
+  let options = {
+    output: 'production', // 'debug' | 'production' | 'omt'
+    outPath : __dirname + '/../public/mapstyle',
+    i18n : true,
+    icons : true,
+    pixelRatios : [1,2],
+    styleDir : path.resolve(path.join(__dirname, '..', '..', 'node_modules', '@qwant' , 'qwant-basic-gl-style')),
+    conf : {tileserver_base : '{tileserver_base}', tileserver_poi : '{tileserver_poi}', spriteserver : '{spriteserver}', fontserver : '{fontserver}'},
+    ignoreSprite : true
```
amatissart: Because of this flag, `sprite` property will be removed from the style object.  
So the spriteserver is not present in the result :(

local_modules/mapbox_style_configure/index.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function configure(style, mapStyleConfig, lang) {
+  return JSON.parse(style
+    .replace(/\{locale\}/g, lang)
+    .replace('"{tileserver_base}"', mapStyleConfig.baseMapUrl)
+    .replace('"{tileserver_poi}"', mapStyleConfig.poiMapUrl)
+    .replace('{spriteserver}', mapStyleConfig.spritesUrl)
+    .replace('{fontserver}', mapStyleConfig.fontsUrl)
```
amatissart: `spriteserver` and `fontserver` need to be absolute urls.  
We could probably build it using `req` properties (similar to  "src/adapters/scene_config.js" )

bin/app.js
```diff
@@ -59,6 +60,8 @@ function App(config) {
     maxAge: config.statics.maxAge
   }))
 
+  app.use('/style.json', new mapStyle(config, languages))
```
amatissart: I think we should use express compression on this endpoint.  
The gzipped style is 8KB, compared to 100KB without compression

bin/middlewares/map_style.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  let builtStyle = styleBuilder(qwantStyle, options)
+
+  return function (req, res) {
+    /* get lang routine */
+    let langParameter = req.query.lang
```
amatissart: At this point, it could be a good idea to add documentation about this kind of endpoints and parameters. In the readme for example.  
Especially as this one will be used by other services.

bin/middlewares/map_style.js
Julien-laville: i'll use res.status
update to a json like {error : Unsupported language code : #lang code}
and return to avoid another send :)

bin/middlewares/map_style.js
Julien-laville: i think,i'll keep the old boolean (sprite exist) for transform call & use build sprite for the spritezero call.

local_modules/mapbox_style_configure/index.js
Julien-laville: i think 
{ spritesUrl : `${req.get('host')}${config.mapStyle.spritesUrl}` ,
 fontsUrl : `${req.get('host')}${config.mapStyle.fontsUrl}`
}
is enough

bin/app.js
Julien-laville: sure, i'll look for the how to

bin/middlewares/map_style.js
Julien-laville: compression (the most used middleare) enable gzip for all request, it take a filter param but i think it's dirty way. 
or i can enable compression for all request

bin/app.js
Julien-laville: compression (the most used middleare) enable gzip for all request, it take a filter param but i think it's dirty way. 
or i can enable compression for all request

bin/middlewares/map_style.js
Julien-laville: done

bin/middlewares/map_style.js
amatissart: cool ! please create a PR on map-style-builder

bin/middlewares/map_style.js
amatissart: could we describe the default behavior (without `lang` too) ?

bin/middlewares/map_style.js
```diff
@@ -0,0 +1,51 @@
...
+        res.send({error : `Unsupported language code : ${langParameter}`})
+        return
+      }
+    } else {
+      matchedLanguage = languages.defaultLanguage
+    }
+
+    /* config rewriting */
+    let host = req.get('host')
+    let protocol = req.protocol
```
amatissart: `req.protocol` will use the `X-Forwarded-Proto` header to determine protocol.  
But it seems that in our docker swarm environments these header will be overridden with `http` by internal services. 
So it is not reliable, and we should probably always use `https` here.

bin/app.js
Julien-laville: didn't know about it 
the force is fronce in this one

bin/middlewares/map_style.js
Julien-laville: https://github.com/QwantResearch/map-style-builder/pull/23

bin/middlewares/map_style.js
Julien-laville: yep added default param to doc

bin/middlewares/map_style.js
Julien-laville: hmmm ok. i'm impressed

Julien-laville: it will be harder to test on local by replacing style object by url

amatissart: we could work around that, and use `http` when `host === 'localhost'` ?
Or define a new setting with default `https` that we could override in local environments ?

Julien-laville: :S i prefer not and say it's exclusive for external securised services

bin/middlewares/map_style.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    } else {
+      matchedLanguage = languages.defaultLanguage
+    }
+
+    /* config rewriting */
+    let host = req.get('host')
+    let urls = {
+      spritesUrl : `https://${host}${config.mapStyle.spritesUrl}` ,
+      fontsUrl : `https://${host}${config.mapStyle.fontsUrl}`
```
amatissart: Arf, I just realized that we need to handle `baseUrl` here too :(

I think we should try to refactor `toAbsoluteUrl` from "src/adapters/scene_config.js" to be use both on client and server. But that will be for another PR.

PR #82, opened 2018-11-27
Log details about telemetry events

amatissart:
Use req.logger to log telemetry events.  
The `telemetry` object contain a single `type` key. Its value is the event type.


bin/metrics_builder.js
```diff
@@ -16,8 +18,9 @@ module.exports = (app, config, registry) => {
     (req, res) => {
```
Julien-laville: i haven't seen this before, but express.json is a good practice <3

PR #86, opened 2018-11-30
fix service panel hover / map button hover & poi_panel hover

Julien-laville:
capture d ecran_2018-11-30_18-02-00

### DIFF
--- a/src/scss/includes/panels/poi_panel.scss
+++ b/src/scss/includes/panels/poi_panel.scss
@@ -299,7 +299,11 @@ $CARD_PADDING: 8px;
   padding-left: 2px;
   height: 25px;
   font-size: 12px;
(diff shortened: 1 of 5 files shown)


src/scss/includes/panels/poi_panel.scss
```diff
@@ -299,7 +299,11 @@ $CARD_PADDING: 8px;
   padding-left: 2px;
   height: 25px;
   font-size: 12px;
-  color: $secondary_text;
```
amatissart: I think we need to keep this color (by default without hover)

Julien-laville: yep, removed it and haven't seen the difference, now i see it.

PR #91, opened 2018-12-11
Display poi email under website block

Julien-laville:
mail


src/scss/includes/font.scss
```diff
@@ -28,14 +28,15 @@
 
 .icon-pin_geoloc:before {
   content: "\e90d";
+  color: #5c6f84;
```
amatissart: why this change ?

src/scss/includes/font.scss
```diff
@@ -28,14 +28,15 @@
...
 }
 .icon-check:before {
   content: "\e91c";
 }
 .icon-chevron-left:before {
   content: "\e913";
 }
-.icon-chevron-right:before {
+.icon-chevron-right2:before {
```
amatissart: why this change ?

src/scss/includes/font.scss
Julien-laville: argghhhhh ico moon corrupted

src/scss/includes/font.scss
Julien-laville: when i update icomoon i need to remove the 2 for left, i missed this time

amatissart:
Please add a test case :)

PR #94, opened 2018-12-17
fix 24/7 parsing

Julien-laville:
Fix 24/7 hour flag response parsing & add 24/7 display

### DIFF
--- a/src/panel/poi_bloc/opening_minimal.js
+++ b/src/panel/poi_bloc/opening_minimal.js
@@ -13,6 +13,7 @@ export default class MinimalHourPanel {
   async set(poi) {
     let openingBlock = poi.blocks.find((block) => block.type === 'opening_hours')
     this.opening = null
+    this.isTwentyFourSeven = openingBlock.is_24_7
     if(openingBlock) {
       this.opening = new OsmSchedule(openingBlock, this.messages)
     }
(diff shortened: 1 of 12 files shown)


src/views/poi_bloc/hour.dot
```diff
@@ -3,12 +3,14 @@
...
       <div class="icon-icon_clock poi_panel__block__symbol"></div>
       <div class="poi_panel__block__content">
-        <div class="poi_panel__info__hours__status" {{= click(this.extend, this) }}>
-          {{= this.panel.renderPartial(this.nextTransitionPartial) }}
-          <i class="icon-icon_chevron-down poi_panel__info__hours__status__toggle"></i>
-        </div>
-        <div class="poi_panel__info__hours">
-          {{? !this.opening.isTwentyForSeven }}
+        {{? this.isTwentyFourSeven }}
+          <div class="poi_panel__info__hours__status__text">{{= _('Open 24h/24', 'hour block') }}</div>
```
amatissart: the status circle is missing here.  
Maybe this 24/7 logic should be part of "nextTransitionPartial" too.

Also, in english a more common wording would be "Open 24/7".

src/panel/poi_bloc/opening_minimal.js
```diff
@@ -13,6 +13,7 @@ export default class MinimalHourPanel {
   async set(poi) {
     let openingBlock = poi.blocks.find((block) => block.type === 'opening_hours')
     this.opening = null
+    this.isTwentyFourSeven = openingBlock.is_24_7
```
amatissart: `openinBlock` may be null here, causing unhandled errors.

Julien-laville: i move it to the test and init it in the constructor

PR #95, opened 2018-12-18
Qmaps 577 popup

Julien-laville:


src/adapters/poi_popup.js
```diff
@@ -3,29 +3,37 @@ import OsmSchedule from '../../src/adapters/osm_schedule'
...
+
+    if(Device.isMobile(e)) {
+      return
+    }
+    let poi = e.features[0]
+
+    if(this.sceneState.poi === poi.properties.global_id) {
+
+
       return
```
amatissart: This check should be done in the timeout function.  
(Otherwise the popup will still be visible after a click on a poi, when the mouse doesn't move)

And remove these blank lines :)

src/adapters/poi/idunn_poi.js
```diff
@@ -16,17 +16,21 @@ export default class IdunnPoi extends Poi {
     this.address = rawPoi.address
   }
 
-  static async poiApiLoad(id) {
+  static async poiApiLoad(id, options = {}) {
     let rawPoi = null
     let url = `${serviceConfig.idunn.url}/v1/pois/${id}`
+    let requestParams = {}
+    if(options.simple) {
+      requestParams = {verbosity : 'short'}
```
amatissart: `verbosity` param is not implemented for the old `/v1/pois` route.
Use `/v1/places` instead

src/adapters/poi_popup.js
```diff
@@ -3,29 +3,37 @@ import OsmSchedule from '../../src/adapters/osm_schedule'
...
 PoiPopup.prototype.init = function(map) {
   this.map = map
   this.popupHandle = null
   this.timeOutHandler = null
+  this.sceneState = SceneState.getSceneState()
 }
 
 PoiPopup.prototype.addListener = function(layer) {
   this.map.on('mouseenter', layer, (e) => {
-    if(isTouchEvent(e)) {
```
amatissart: why removing this check ? I think it would still be useful to ignore touch events at this point. In addition to `isMobile`.

PR #96, opened 2018-12-24
Qmaps 581 autofav

Julien-laville:


src/adapters/autocomplete.js
```diff
@@ -28,30 +28,46 @@ function SearchInput(tagSelector) {
...
-          if(responses[0]) {
+        try {
+        let [bragiResponse, storeResponse] = await Promise.all([this.suggestPromise, suggestHistoryPromise])
+          if(bragiResponse) {
             this.suggestPromise = null
-            this.suggestList = responses[0].concat(responses[1])
+            this.suggestList = bragiResponse.concat(storeResponse)
             resolve(this.suggestList)
+          } else if(storeResponse) {
+            resolve(storeResponse)
```
amatissart: This breaks the abort mechanism in a subtle way  :fearful:
Currently, we rely on this promise to resolve to `null` when a suggest query is aborted. Since the `suggestPromise` is the only one to be aborted, it would now resolve to a list `storeResponse`. This list will be suggested anyway, and will cancel any pending `source` call ! Some typing in the autocomplete field will have no effect !  

I suppose we could fix this by resolving null, if any response is null:
```javascript
if (bragiResponse == null || storeResponse == null ){
    return resolve(null)
}
```

Or maybe we should clean up the `abort` mechanism to abort both the `suggestPromise` and the wrapping `Promise.all` ?

src/adapters/autocomplete.js
```diff
@@ -99,76 +115,29 @@ SearchInput.prototype.select = async function(selectedPoi) {
...
-}
-
-function HouseTemplate(poi, iconDom) {
-  let {name, postcode, city, countryName} = poi
-  let address = [postcode, city, countryName].filter((zone) => zone).join(', ')
-  return `
-<div class="autocomplete_suggestion__first_line__container">
-  ${iconDom}
-  <div class="autocomplete_suggestion__first_line">${ExtendedString.htmlEncode(name)}</div>
+${alternativeName ? `<div class="autocomplete_suggestion__second_line">${ExtendedString.htmlEncode(alternativeName)}</div>` : ''}
```
amatissart: I think it would be cleaner to always display the "second_line" div, even if it's empty. It would give a consistent height to all autocomplete results.

src/adapters/poi/map_poi.js
```diff
@@ -1,8 +1,9 @@
-import Poi from "./poi";
+import Poi,{POI_TYPE} from "./poi";
 
 export default class MapPoi extends Poi {
   constructor(feature, lngLat) {
     let id = feature.properties.global_id
-    super(id,  feature.name, 'poi', lngLat)
+    console.log(feature)
```
amatissart: to clean up

src/adapters/poi/map_poi.js
```diff
@@ -1,8 +1,9 @@
...
+    super(id,  feature.name,'to replace', POI_TYPE, lngLat)
```
amatissart: `'to_replace'` : is it temporary ? Or should it be null ?

src/adapters/poi/poi.js
```diff
@@ -32,23 +34,14 @@ export default class Poi {
...
-      latLon: this.latLon,
-      id: this.id,
-      name: this.name,
-      className: this.className,
-      subClassName: this.subClassName,
-      zoom: this.zoom,
-      type: 'poi',
-      bbox: this.bbox,
-    }
+    return this
```
amatissart: I am not fond of this solution.  
I think it would be useful to keep an explicit list of the keys that will be stored on the client. For example, to make future migrations easier, or prevent serialization inconsistencies between different browsers.

src/adapters/poi/poi.js
```diff
@@ -6,17 +6,19 @@ import ExtendedString from "../../libs/string";
...
     this.name = name
+    this.alternativeName = alternativeName
     this.type = type
     this.latLon = latLon
     this.className = className
     this.subClassName = subClassName
-    this.tags = tags
-    this.computeZoom()
+    this.zoom = this.computeZoom()
+    this.isFavorite = true
```
amatissart: :thinking: what's the use of this boolean ? and is it always `true` ?

src/adapters/poi/map_poi.js
Julien-laville: was temporary until i don't find matching data to fit alternative name (fixed to null)

src/adapters/poi/map_poi.js
Julien-laville: (fixed)

src/adapters/autocomplete.js
Julien-laville: I'll give a hint (updated screenshot incoming)

Julien-laville: acceptable

src/adapters/poi/poi.js
Julien-laville: I'll fix this in a more élégant way :)  (lodash have i very usefull pick method)

src/adapters/poi/poi.js
Julien-laville: DK
it's an artefact comming from my investigation .

src/adapters/autocomplete.js
```diff
@@ -99,76 +115,29 @@ SearchInput.prototype.select = async function(selectedPoi) {
   }
 }
 
+function remotesRender(pois) {
+  return pois.map(poi => renderItem(poi)).join('')
+}
+
+function favoritesRender(pois) {
+  return `<h3 class="autocomplete_suggestion__category_title">${_('FAVORIS', 'autocomplete')}</h3> ${pois.map(poi => renderItem(poi)).join('')}`
```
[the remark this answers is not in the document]
Julien-laville: hmmm so favorites

language/message/br_FR.po
```diff
@@ -310,9 +298,6 @@ msgstr ""
 msgid "erotic shop"
 msgstr ""
 
-msgid "escape game"
-msgstr ""
-
```
amatissart: these keys should not be deleted. They are still defined.

src/adapters/poi/poi.js
```diff
@@ -32,23 +33,20 @@ export default class Poi {
...
-      latLon: this.latLon,
-      id: this.id,
-      name: this.name,
-      className: this.className,
-      subClassName: this.subClassName,
-      zoom: this.zoom,
-      type: 'poi',
-      bbox: this.bbox,
-    }
+    const serializeKeys = ['id', 'name', 'alternativeName', 'type', 'latLon', 'className', 'subClassName']
```
amatissart: I think we want to store some other details : 
 * `zoom` is currently required to apply the correct zoom level when clicking on a favorite. With this implementation it  goes to zoom 16, instead of 18. It would be worth adding a test case.
 * are we sure that `localName` and other fields are not required ?

src/adapters/poi/poi_store.js
```diff
@@ -1,19 +1,33 @@
...
+    let storedData
+    try {
+      storedData = await store.getAllPois()
+    } catch(e) {
+      Telemetry.add(Telemetry.FAVORITE_ERROR_LOAD_ALL)
+      Error.sendOnce('favorite_panel', 'getAll', 'error getting pois', e)
+    }
+    return Object.keys(storedData).map((mapPoint) => {
+      let historySuggest = storedData[mapPoint]
+      return Object.assign(new PoiStore(), historySuggest)
```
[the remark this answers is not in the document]
enigme20182019: Yep it does (i was sure that getAllPoi return an object) Thank's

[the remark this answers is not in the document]
Julien-laville: Yep it does (i was sure that getAllPoi return an object) Thank's

src/adapters/poi/poi.js
Julien-laville: i agree for zoom, i'm not sure about localName and others fields

language/message/br_FR.po~
```diff
@@ -0,0 +1,804 @@
+# Translators:
```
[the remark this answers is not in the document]
Julien-laville: :s

PR #99, opened 2019-01-03
update style

nlehuby:
* enhance affordance
* remove unused picto
* add a few missing picto


Julien-laville:
Tested on chrome windows, seems to be ok.

PR #101, opened 2019-01-14
Update translations strings and style

nlehuby:
* add catalan language
* update br, de and fr languages
* update map style


amatissart:
I think that tests are failing because of a missing formats file (in language/date) for the new language.  
@Julien-laville What is the source for these files ?

PR #105, opened 2019-01-17
Qmaps 664 direction suggest

Julien-laville:
update autocomplete lib with set / get / destroy / prerender / onsubmit
add direction search test


amatissart:
One issue about the suggest behavior:  
Suggestions lists should be updated with their new content when clicking in a field after swapping start and end points

PR #106, opened 2019-01-18
Fix CSS (vehicles, roads, roadmap, return link)

Julien-laville:


src/scss/includes/direction.scss
```diff
@@ -159,17 +159,18 @@
   border-left: 4px solid #ff3b4a;
 }
 
-#itinerary_roadmap {
+.itinerary_roadmap {
   background: #fff;
   overflow: auto;
-  max-height: calc(100vh - 300px);
+  max-height: calc(100vh - 290px);
 }
```
amatissart: Could you also fix the background url, at the top of this file (line 6) ?

PR #107, opened 2019-01-18
Zoom on each step on mobile step-by-step view 

Julien-laville:


src/adapters/scene_direction.js
```diff
@@ -130,9 +134,9 @@ export default class SceneDirection {
     }
   }
 
-  computeBBox(route) {
```
Julien-laville: bad naming i prever a generic one.

PR #111, opened 2019-01-23
clean last_val after suggest value set

Julien-laville:
There was a little bug after invert start & end on the suggest preventing type the same word before invert


amatissart:
In this file, `last_val` is initialized with `''`, and sometimes reset to `'\n'`.
Now after this change, it may also be set to `null`.  
Could we clarify the meaning of these 3 different blank values ? :thinking:

PR #113, opened 2019-01-24
Fix mask panel display when local-storage is disabled

Julien-laville:


amatissart:
Other problems remain when `localStorage` is not available. 
For example, autocomplete results are broken, because favorite POIs cannot be retrieved.

Should we include these issues in this PR ?

Julien-laville:
@amatissart i'll evaluate this

PR #114, opened 2019-01-25
Fix keyboard native navigation (arrows + -)

Julien-laville:


src/ui_components/search_input.js
```diff
@@ -1,6 +1,14 @@
...
-import BragiPoi from "../adapters/poi/bragi_poi";
+
+const MAPBOX_RESERVED_KEYS = [
+    37 // ←
+  , 38 // ↑
+  , 39 // →
+  , 40 // ↓
+  , 107 // -
+  , 109 // +
+]
```
amatissart: we should also add `=` (https://docs.mapbox.com/mapbox-gl-js/api/#keyboardhandler)

src/ui_components/search_input.js
```diff
@@ -35,6 +43,7 @@ export default class SearchInput {
 
   handleKeyboard() {
     document.onkeydown = function(e) {
+      if(MAPBOX_RESERVED_KEYS.find((key) => key === e.keyCode)) {console.log( e.keyCode); return}
```
amatissart: console.log to remove.

amatissart: Shouldn't we use `e.key` instead of deprecated `e.keyCode` ?
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#Properties

Julien-laville: i'll never know, for now i think key is more safe, maybe i'll build an abstraction one day
(key will need a litle polyfill)

src/ui_components/search_input.js
Julien-laville: i have miss this one ;(

src/ui_components/search_input.js
Julien-laville: yep

PR #118, opened 2019-01-30
Masq integration

sujeebant:


src/adapters/poi/poi_store.js
```diff
@@ -23,7 +23,8 @@ export default class PoiStore extends Poi {
       storedData = await store.getAllPois()
     } catch(e) {
       Telemetry.add(Telemetry.FAVORITE_ERROR_LOAD_ALL)
-      Error.sendOnce('favorite_panel', 'getAll', 'error getting pois', e)
+      Error.sendOnce('poi_store', 'getAll', 'error getting pois', e)
```
Julien-laville: thx ;)

src/adapters/store.js
```diff
@@ -27,8 +35,23 @@ function Store() {
   }
 }
 
+Store.prototype.getUserInfo = async function() {
+  try {
+    if (! (await checkRegistered())) {
```
Julien-laville: the condition can be reversed to avoid that error prone syntax

src/adapters/store.js
```diff
@@ -27,8 +35,23 @@ function Store() {
...
+      throw new Error('Not registered')
+    }
+    return await abstractStore.getUserInfo()
+  } catch (e) {
+    Error.sendOnce('store', 'getAllPois', 'error getting pois', e)
```
Julien-laville: Copy pasta error 🍝

src/adapters/store.js
```diff
@@ -27,8 +35,23 @@ function Store() {
...
+    throw e
+  }
+}
+
 Store.prototype.getAllPois = async function() {
   try {
+    if (! (await checkRegistered())) {
```
Julien-laville: you need an abstraction to avoid this copy pasta sauce 🍝

src/adapters/store.js
```diff
@@ -38,6 +61,10 @@ Store.prototype.getAllPois = async function() {
 
 Store.prototype.getLastLocation = async function() {
   try {
+    if (! (await checkRegistered())) {
+      return await fallbackStore.get(`qmaps_v${version}_last_location`)
+    }
+    console.log('registered')
```
Julien-laville: linter here

src/adapters/store.js
```diff
@@ -47,18 +74,17 @@ Store.prototype.getLastLocation = async function() {
 
 Store.prototype.setLastLocation = async function(loc) {
   try {
+    if (! (await checkRegistered())) {
+      return await fallbackStore.set(`qmaps_v${version}_last_location`, loc)
```
Julien-laville: we need a better logic between fallback and masq

src/libs/local_store.js
```diff
@@ -19,15 +19,31 @@ LocalStore.prototype.getAllPois = function() {
...
+LocalStore.prototype.isRegistered = function() {
+  return Promise.resolve(true)
+}
+
 LocalStore.prototype.onConnect = function () {
   return Promise.resolve()
 }
 
+LocalStore.prototype.has = async function(k) {
+  return !!(await this.get(k))
```
Julien-laville: that's golf

src/libs/local_store.js
```diff
@@ -19,15 +19,31 @@ LocalStore.prototype.getAllPois = function() {
...
+  return Promise.resolve(null)
+}
+
+LocalStore.prototype.registerApp = function() {
   console.log('local storage doesn\'t support register method')
   return Promise.resolve()
 }
 
+LocalStore.prototype.unregisterApp = function() {
+  return Promise.resolve()
```
Julien-laville: a think declare as async and return true / false is equivalent and more clear

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+
+  const filteredKeys = Object.keys(list).reduce((filtered, k) => {
+    if (Poi.isPoiCompliantKey(k)) {
+      filtered.push(k)
+    }
+    return filtered
+  }, [])
+
+  if (filteredKeys.length === 0) {
+    return Promise.resolve([])
```
Julien-laville: i think the async keyword is enought so you can return [] directly (can you confirm ?)

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+  }
+
+  const valuePromises = filteredKeys.map(k => {
```
Julien-laville: i dont understand what this const is

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+    return this.masq.get(k)
+  })
+
+  const values = await Promise.all(valuePromises).catch(
```
Julien-laville: i think we can use the try catch syntax here instead of mixin old Promise then & await syntax

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+}
+
+MasqStore.prototype.registerApp = async function(apps) {
+  // connect to Masq
+  window.open(this.loginLink)
+  await this.masq.logIntoMasq(false)
+}
+
+MasqStore.prototype.unregisterApp = async function() {
+  await this.masq.signout()
```
Julien-laville: i'm confuse with signout logout & unregister

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+}
+
+MasqStore.prototype.onConnect = async function () {
+  // executed when maps is opened
+  this.masq = new Masq(this.config.title, this.config.desc, this.config.icon)
+  if (this.masq.isLoggedIn()) {
+    await this.masq.connectToMasq()
```
Julien-laville: you can return here

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+  } else {
+    this.loginLink = await this.masq.getLoginLink()
```
Julien-laville: & here

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+  }
+
+  return
+}
+
+MasqStore.prototype.has = async function(k) {
+  return !!(await this.get(k))
```
Julien-laville: this is golf 🏌

src/main.js
```diff
@@ -5,12 +5,16 @@ import './libs/actions'
...
 import UrlState from "./proxies/url_state";
+import Store from "./adapters/store";
 
 
 (async function main() {
   new I18n()
   await setLang()
 
+  const store = new Store()
+  await store.onConnect()
```
Julien-laville: are we sure we want to freeze app start-up to get store data ?

src/panel/register_masq.js
```diff
@@ -8,19 +8,57 @@ function RegisterMasqPanel() {
   this.store = new Store()
   this.panel = new Panel(this, RegisterMasqPanelView)
   this.isActive = false
+  this.registering = false
+  this.username = null
+  this.profileImage = null
+
+  this.isRegistered = false
+  this.store.isRegistered().then(async (b) => {
```
Julien-laville: move to an async function

src/panel/register_masq.js
```diff
@@ -8,19 +8,57 @@ function RegisterMasqPanel() {
...
+
+RegisterMasqPanel.prototype.unregister = async function () {
+  await this.store.unregister()
+  this.isRegistered = await this.store.isRegistered()
+  if (this.isRegistered) {
+    const { username, profileImage } = await this.store.getUserInfo()
+    this.username = username
+    this.profileImage = profileImage
+  }
+  window.location.reload()
```
Julien-laville: Oh waw the uge one

src/views/register_masq.dot
```diff
@@ -1,11 +1,31 @@
...
-  <p>
-    {{= _('We use MASQ to keep your preference data safe', 'mask') }}
-  </p>
-  <br>
-  <button class="register_masq_panel__action" {{= click(this.register, this) }}>
-    {{= _('Register', 'mask') }}</button>
-</div>
\ No newline at end of file
+<div class="register_masq_panel">
+  {{? (!this.isRegistered)}}
```
Julien-laville: why (

src/views/register_masq.dot
```diff
@@ -1,11 +1,31 @@
...
+    <h1 class="register_masq_panel__title">
+      {{= _('Register with masq', 'masq') }}
+    </h1>
+    <p>
+      {{= _('We use MASQ to keep your personal data safe', 'masq') }}
+    </p>
+    {{? (!this.registering)}}
```
Julien-laville: is ( necessary ?

src/adapters/store.js
```diff
@@ -3,111 +3,169 @@ import Error from '../adapters/error'
...
 import ExtendedString from "../libs/string";
 
-let moduleConfig = nconf.get().store
+export default class Store {
 
-const AbStore = require(`../libs/${moduleConfig.name}`)
-const abstractStore = new AbStore(moduleConfig.endpoint)
+  constructor() {
+    // get store from window if already initialized
+    if (window.__store) {
```
Julien-laville: ok for me, but i need to explore a little this solution ( could be better than mine ;) )

src/adapters/store.js
```diff
@@ -3,111 +3,169 @@ import Error from '../adapters/error'
...
-    })
+    // init stores
+    const LocalStore = require(`../libs/local_store`)
+    this.localStore = new LocalStore()
+    this.loggedIn = false
+    this.abstractStore = this.localStore
+    const MasqStore = require(`../libs/masq`)
+    this.masqConfig = nconf.get().masq
+    this.masqStore = new MasqStore(this.masqConfig)
+    const masqInitPromise = this.masqStore.onConnect()
```
Julien-laville: unused

src/adapters/store.js
```diff
@@ -3,111 +3,169 @@ import Error from '../adapters/error'
...
+    const masqInitialized = false
```
Julien-laville: unused

src/adapters/store.js
```diff
@@ -3,111 +3,169 @@ import Error from '../adapters/error'
...
-      this.del(poi)
-    })
-    listen('store_center', (loc) => {
-      this.setLastLocation(loc)
-    })
-    listen('store_clear', () => {
-      this.clear()
-    })
+    // init stores
+    const LocalStore = require(`../libs/local_store`)
```
Julien-laville: i'm not fond of dynamic require plus the syntax is the old one

src/adapters/store.js
sujeebant: checkRegistered is not needed anymore with the new logic of calling localStorage until logged into masq

src/adapters/store.js
sujeebant: :+1:

src/libs/local_store.js
sujeebant: :+1:

src/adapters/store.js
```diff
@@ -3,111 +3,169 @@ import Error from '../adapters/error'
...
+        url: window.location.origin + window.location.pathname,
+        title: this.masqConfig.title,
+        desc: this.masqConfig.desc,
+        icon: this.masqConfig.icon
+      }
+
+      await this.masqStore.login(loginParams)
+    } catch (e) {
+      Error.sendOnce('store', 'login', 'error logging in', e)
+      throw e
```
Julien-laville: this error is catched twice (here and masqLoginPanel)

src/libs/masq.js
sujeebant: Yes you're right !

src/libs/masq.js
sujeebant: the array of promises that resolve to the values corresponding to the keys we are getting

src/adapters/store.js
```diff
@@ -3,111 +3,169 @@ import Error from '../adapters/error'
...
-    title: moduleConfig.masq.title,
-    desc: moduleConfig.masq.desc,
-    icon: moduleConfig.masq.icon
+  async getAllPois() {
+    await this.checkInit()
+    try {
+      return await this.abstractStore.getAllPois()
+    } catch (e) {
+      Error.sendOnce('store', 'getAllPois', 'error getting pois', e)
+      throw e
```
Julien-laville: this file know wich store is used, so i prefer a more explicit error with the current store used, plus it's not necessary to re throw the excpetion, you can handle here with return []

src/libs/masq.js
sujeebant: ok !

src/libs/masq.js
sujeebant: Everything has been renamed to use login/logout

src/libs/local_store.js
```diff
@@ -27,15 +27,31 @@ LocalStore.prototype.getAllPois = function() {
...
+LocalStore.prototype.isLoggedIn = function() {
+  return Promise.resolve(false)
+}
+
 LocalStore.prototype.onConnect = function () {
   return Promise.resolve()
 }
 
+LocalStore.prototype.has = async function(k) {
+  return !!(await this.get(k))
```
Julien-laville: i prefer Boolean(

src/main.js
sujeebant: changed to start execute onConnect at instantiation and await the resolution of the promise only when needed (at method execution)

src/libs/masq.js
```diff
@@ -1 +1,112 @@
-module.exports = require('masq-client/dist/masq')
+const Masq = require('masq-lib')
+
+const Error = require('../adapters/error').default
```
Julien-laville: can you use the new import syntax when it's possible (or maybe it is not)

src/libs/masq.js
Julien-laville: await catch then async mix here

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+    return this.masq.get(k)
+  })
+
+  const values = await Promise.all(valuePromises).catch(
+    (e) => {
+      handleError('getAllPois', 'error getting pois', e)
+      throw e
+    })
+
+  return values
```
Julien-laville: value is redundant

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+  return !!(await this.get(k))
+}
+
+MasqStore.prototype.get = async function(k) {
+  const value = await this.masq.get(k).catch(
+    (e) => {
+      handleError('get', `error parsing item with key ${k}`, e)
+      throw e
+    })
+  return value
```
Julien-laville: value is redundant

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+  return Promise.resolve()
+}
+
+MasqStore.prototype.del = async function(k) {
+  await this.masq.del(k).catch(
+    (e) => {
+      handleError('del', 'error removing item', e)
+      throw e
+    })
+  return
```
Julien-laville: si return pertinent ?

src/libs/masq.js
```diff
@@ -1 +1,112 @@
...
+  return value
+}
+
+MasqStore.prototype.set = async function(k, v) {
+  await this.masq.put(k, v).catch(
+    (e) => {
+      handleError('set', 'error setting item', e)
+      throw e
+    })
+  return
```
Julien-laville: si return pertinent ?

src/libs/panel.js
```diff
@@ -111,4 +111,4 @@ Panel.prototype.animate = async function(t, selector, style = {}) {
   })
 }
 
-export default Panel
\ No newline at end of file
+export default Panel
```
Julien-laville: thank's :3

src/panel/login_masq.js
```diff
@@ -0,0 +1,66 @@
...
+export default class LoginMasqPanel {
+  constructor() {
+    this.panel = new Panel(this, LoginMasqPanelView)
+    if(moduleConfig.name === 'masq') {
+      this.store = new Store()
+      this.isActive = false
+      this.loggingIn = false
+      this.username = null
+      this.profileImage = null
+      listen('login_panel__show', () => {
```
Julien-laville: you can remove _panel

src/panel/login_masq.js
```diff
@@ -0,0 +1,66 @@
...
+        this.panel.animate(.25, '.login_masq_panel', {top : '100px'})
+      })
+
+      this.isLoggedIn = false
+      this.store.isLoggedIn().then(async (b) => {
```
Julien-laville: can you move this to an async function so you can clean async syntax

src/panel/login_masq.js
```diff
@@ -0,0 +1,66 @@
...
+    try {
+      await this.store.login()
+      this.isLoggedIn = await this.store.isLoggedIn()
+      if (this.isLoggedIn) {
+        const { username, profileImage } = await this.store.getUserInfo()
+        this.username = username
+        this.profileImage = profileImage
+      }
+    } catch(e) {
+      Error.sendOnce('login_masq', 'login', 'error loggingIn masq', e)
```
Julien-laville: i think this error is already handle in store 
i prefer avoid sending 2 messages with the same error

src/scss/includes/panels/login.scss
```diff
@@ -0,0 +1,41 @@
+.login_masq_panel {
```
Julien-laville: very nice css !

src/adapters/store.js
Julien-laville: after a little of reading, i'm completely ok with the approach, it's fully correct !
but 
the singleton can be recognized by it's signature (getInstance) i let you decide which one you prefer
by my side i have learnt something

src/adapters/store.js
sujeebant: It was supposed to be an attribute...

src/adapters/store.js
sujeebant: It's an error on my part, it was supposed to be an attribute used in checkInit

src/adapters/store.js
sujeebant: :+1:

src/libs/masq.js
sujeebant: moved the whole file to es6 syntax !

src/libs/local_store.js
sujeebant: done in local_store.js and masq.js

src/libs/masq.js
sujeebant: fixed !

src/libs/masq.js
sujeebant: :+1:

src/libs/masq.js
sujeebant: :+1:

src/libs/masq.js
sujeebant: cleaned unnecessary return statements in masq.js

src/libs/local_store.js
sujeebant: cleaned !

src/libs/masq.js
sujeebant: cleaned !

src/libs/masq.js
sujeebant: cleaned unnecessary return statements in masq.js

src/panel/login_masq.js
sujeebant: moved to an async function called in the constructor without await

src/scss/includes/panels/login.scss
sujeebant: just renamed some class names

src/views/register_masq.dot
sujeebant: You are right '(' is not necessary

src/views/register_masq.dot
sujeebant: You are right '(' is not necessary

src/adapters/store.js
sujeebant: As you have pointed out this implementation hides the fact that Store is a singleton, but the user of the class is able to just call 'new Store()' to get the Store instance.
I guess the two approaches are valid, I will leave it as it is, we can change it later if needed.

src/panel/register_masq.js
sujeebant: :man_facepalming: I wanted to refresh the view and favorite list...
Changed to fire an event on login and logout caught by Favorite panel to refresh the list

package.json
```diff
@@ -89,6 +88,7 @@
     "ejs": "^2.6.1",
     "express": "^4.16.3",
     "express-static-gzip": "^1.1.3",
+    "masq-lib": "git+https://github.com/QwantResearch/masq-lib.git#maps-integration",
```
amatissart: I think this dependency is only required on build, and should be moved to "devDependencies".  

This currently blocks the docker image build, as it also depends on other devDependencies.

sujeebant: You're righ ! I moved masq-lib to devDependencies

src/adapters/scene.js
```diff
@@ -123,7 +123,7 @@ Scene.prototype.initMapBox = function () {
       let lng = this.mb.getCenter().lng
```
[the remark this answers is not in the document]
sujeebant: done

src/panel/login_masq.js
```diff
@@ -0,0 +1,64 @@
+import LoginMasqPanelView from '../views/login_masq.dot'
+import Panel from "../libs/panel";
+import Store from "../adapters/store"
+import Error from '../adapters/error'
+import nconf from "../../local_modules/nconf_getter";
+let moduleConfig = nconf.get().store
```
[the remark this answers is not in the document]
sujeebant: removed nconf import and moduleConfig declaration

src/panel/login_masq.js
sujeebant: removed the listener

src/adapters/store.js
sujeebant: removed error logging in panel/login_masq

src/panel/login_masq.js
sujeebant: okay ! removed the second error logging in panel/login_masq as the first one can be more specific with the name of the store which is used

src/adapters/store.js
```diff
@@ -2,112 +2,170 @@ import nconf from '@qwant/nconf-getter'
...
+    this.loggedIn = false
+    this.abstractStoreStr = 'local_store'
+    this.abstractStore = this.localStore
+    this.masqConfig = nconf.get().masq
+    if (this.masqConfig.enabled) {
+      this.masqStore = new MasqStore(this.masqConfig)
+      if (this.masqStore.isLoggedIn()) {
+        this.abstractStore = this.masqStore
+      }
+    }
```
Julien-laville: else new localStore 
should be more acceptable, here you instanciate local store, and you instanciate masqStore

src/adapters/store.js
```diff
@@ -2,112 +2,170 @@ import nconf from '@qwant/nconf-getter'
...
-    listen('store_center', (loc) => {
-      this.setLastLocation(loc)
-    })
-    listen('store_clear', () => {
-      this.clear()
-    })
+    // init stores
+    this.localStore = new LocalStore()
+    this.loggedIn = false
+    this.abstractStoreStr = 'local_store'
```
Julien-laville: why not displace the logic to currentStore.getName or something

src/adapters/store.js
```diff
@@ -2,112 +2,170 @@ import nconf from '@qwant/nconf-getter'
...
+    this.abstractStore = this.localStore
+    this.masqConfig = nconf.get().masq
+    if (this.masqConfig.enabled) {
+      this.masqStore = new MasqStore(this.masqConfig)
```
Julien-laville: you should let masqStore handle is own config internaly

src/adapters/store.js
```diff
@@ -2,112 +2,170 @@ import nconf from '@qwant/nconf-getter'
...
+import MasqStore from "../libs/masq"
 
-let moduleConfig = nconf.get().store
+export default class Store {
 
-const AbStore = require(`../libs/${moduleConfig.name}`)
-const abstractStore = new AbStore(moduleConfig.endpoint)
+  constructor() {
+    // get store from window if already initialized
+    if (window.__store) {
```
Julien-laville: I have gather more data about our "problem" 
https://stackoverflow.com/questions/519520/difference-between-static-class-and-singleton-pattern
the fact that this class is never used as instance made me think that we should declare it as a "static class"
now it's a little late

PR #131, opened 2019-02-19
Update poi panel on login/logout

sujeebant:
On login or logout refresh the boolean isPoi and update the panel


sujeebant:
> Maybe it's too late but as store is a singleton can we use method instead of spaguetto event ?
> whe should evaluate the size of the change before make a choice
> Tell me

We could have store extend EventTarget  #134

PR #132, opened 2019-02-19
Reformat getAllPois function and change description

sujeebant:


src/libs/masq.js
```diff
@@ -84,25 +84,18 @@ export default class MasqStore {
...
-    if (filteredKeys.length === 0) {
-      return []
-    }
+    try {
+      const list = await this.masq.list()
 
-    const valuePromises = filteredKeys.map(k => {
-      return this.masq.get(k)
-    })
+      const filteredValues = Object.entries(list).reduce((filtered, kv) => {
```
Julien-laville: I feel that its more readable to replace the generic reduce by a more specific [map] ◯ [filter]
you can avoid duplicated variable

src/libs/masq.js
```diff
@@ -84,25 +84,18 @@ export default class MasqStore {
 
   async getAllPois() {
     await this.checkInit()
-    const list = await this.masq.list()
```
Julien-laville: list need prefix param, you take a risk passing undefined

sujeebant: prefix is optional

src/libs/masq.js
sujeebant: The reduce should be more efficient than applying filter and map, but I can change it if you think it would be more readable ?
It would be something like
`Object.entries(list).filter(kv => Poi.isPoiCompliantKey(kv[0])).map(kv => kv[1])`

Julien-laville: Yep more readable maybe add some new line :)

src/libs/masq.js
Julien-laville: it's error prone : you should use as you want another method listPrefix or an optionnal defaulted to {} field  which could contain prefix

PR #134, opened 2019-02-21
Use EventTarget instead of fire and listen

sujeebant:
https://caniuse.com/#search=eventTarget

### DIFF
--- a/src/adapters/store.js
+++ b/src/adapters/store.js
@@ -5,10 +5,10 @@ import ExtendedString from "../libs/string";
 import LocalStore from "../libs/local_store"
 import MasqStore from "../libs/masq"
 
-export default class Store extends EventTarget {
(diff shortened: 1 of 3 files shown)


src/adapters/store.js
```diff
@@ -5,9 +5,11 @@ import ExtendedString from "../libs/string";
 import LocalStore from "../libs/local_store"
 import MasqStore from "../libs/masq"
 
-export default class Store {
+export default class Store extends EventTarget {
```
Julien-laville: Very nice !
It's an open subject, but i think composition is better here than heritance. Here store have nothing to do with Event. it's Like you build a House and herit from firetruck beacause you need red paint.

sujeebant: done !

PR #141, opened 2019-02-26
Qmaps 720

xem:


public/images/direction_icons/corner-up-left-grey.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="#353c52" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="feather feather-corner-up-left"><polyline points="9 14 4 9 9 4"></polyline><path d="M20 20v-7a4 4 0 0 0-4-4H4"></path></svg>
```
amatissart: color is not consistent with the geolocation button (`#5c6f84` vs `#353c52`)
Ideally we would add this to the icomoon font (that already contains "corner-up-right")

src/panel/direction/direction_panel.js
```diff
@@ -134,6 +134,9 @@ export default class DirectionPanel {
     this.panel.update()
     this.cleanDirection()
     UrlState.pushUrl()
+    document.querySelectorAll('.menu__button')[0].style.opacity = 1
+    document.querySelectorAll('.search_form')[0].style.opacity = 1
+    document.querySelectorAll('.direction_shortcut')[0].style.opacity = 1
```
amatissart: this looks a bit hackish :/
At least, we could put the elements to collapse in a single constant, and use `visibility` instead of `opacity` to prevent accessibility issues.

src/views/direction/direction.dot
```diff
@@ -27,3 +27,5 @@
   </div>
   {{= this.roadMapPanel.render() }}
 </div>
+
+<button class="direction_shortcut" {{= click(this.open, this) }}></div>
```
amatissart: Maybe this button should be part of `ExtendedControl`, just like the geolocation button ?

The position is currently fixed, and is broken after the itinerary preview has been enabled:
image

src/panel/direction/direction_panel.js
xem: +1 for visibility,
Is it ok if I toggle a class on the topmost container to hide/show these 3 items at once?

public/images/direction_icons/corner-up-left-grey.svg
xem: I removed the colored svgs,
I fixed the colors,
I used the corner-up-right icon from the font (we can change it when the up-left icon will be added in the font)

src/panel/direction/direction_panel.js
amatissart: and set `visibility` in css ? yes, I would prefer this solution.

src/views/direction/direction.dot
xem: The bug shown in your screenshot is fixed: when opening the itinerary panel (or a route preview), the direction button is hidden.

but I'll need help for moving the button in ExtendedControl (why is there no .dot template for them?)

PR #143, opened 2019-02-28
Propagate login/logout across tabs

sujeebant:


src/adapters/store.js
```diff
@@ -26,6 +26,18 @@ export default class Store {
       if (this.masqStore.isLoggedIn()) {
         this.abstractStore = this.masqStore
       }
+
+      this.masqStore.masq.eventTarget.addEventListener('logged_in', async () => {
```
amatissart: Since masqStore init is async, `this.masqStore.masq` may not be defined here: 
```
TypeError: this.masqStore.masq is null
```

amatissart: Or maybe that will be solved with #164 ?

amatissart:
Included into #164

PR #150, opened 2019-03-18
Prerender favorite in autocompletes

Julien-laville:
Capture d’écran_2019-03-18_10-53-03


Julien-laville:
@jbgriesner not related to this feature but still fixed

src/ui_components/search_input.js
```diff
@@ -40,6 +41,9 @@ export default class SearchInput {
     this.searchInputHandle = document.querySelector(tagSelector)
     this.handleKeyboard()
     this.suggest = new Suggest(tagSelector, (selectedPoi) => this.selectItem(selectedPoi))
+    PoiStore.getAll().then((storeItems) => {
```
sujeebant: Initializing suggested favorites in the constructor of the search input component means the suggestions won't be updated when a new favorite is added.

Steps to reproduce the bug:
1. focus on search focus to see the list of current favorites as suggestion
2. add a new favorite
3. focus again on the search input, the new favorite is not displayed

Julien-laville: We will fix that in the future ticket favorite synch, thanks

sujeebant: Okay, there is also an issue when clearing the field of direction input, we don't have the suggestion anymore, but I guess it is not directly related to this PR and will be fixed in a future ticket.

PR #153, opened 2019-03-19
Fix back button

Julien-laville:


src/panel/direction/direction_panel.js
```diff
@@ -22,9 +20,18 @@ export default class DirectionPanel {
     this.origin = null
     this.destination = null
     this.vehicle = this.vehicles.DRIVING
-    this.roadMapPanel = new RoadMapPanel()
+    this.roadMapPanel = new RoadMapPanel(this)
```
G-Ray: Would it be possible to register a callback instead of passing `this` ? I find a bit dangerous to let roadMapPanel handle the instance

G-Ray: I mean, a callback to handle a close event

src/panel/direction/direction_panel.js
```diff
@@ -130,6 +137,10 @@ export default class DirectionPanel {
     PanelManager.resetLayout()
   }
 
+  back() {
+    this.activePanel.closeAction()
```
G-Ray: Registering a close event/cb would simplyfy the logic and would avoid the need to update this.activePanel

src/panel/direction/road_map_panel.js
```diff
@@ -25,16 +26,22 @@ export default class RoadMapPanel {
   }
 
   preview() {
+    this.directionPanel.activePanel = this
```
G-Ray: same thing, maybe we can call a callback to signal to the DirectionPanel instance that it is active. This way we don't need to directly handle the instance

src/panel/direction/direction_panel.js
Julien-laville: good question, i find it more elegant than register callback and let the panel handle his own close method.

PR #154, opened 2019-03-19
Fix url restore vehicle

Julien-laville:


src/panel/direction/direction_panel.js
```diff
@@ -236,6 +237,13 @@ export default class DirectionPanel {
         Error.sendOnce('direction_panel', 'restoreUrl', `Error restoring Poi from Url ${getParams.get('destination')}`, err)
       }
     }
+    if(getParams.get('vehicle')) {
```
amatissart: actually, this parameter should be called `mode` and is already restored on line 217

Julien-laville: x_x

PR #165, opened 2019-04-04
Load Masq lib in separate chunk (fixes IE11)

amatissart:
Masq lib bundle is not compatible with IE11.
With these changes, the Masq lib is loaded dynamically, so that webpack builds a separate chunk that will be used only if Masq integration is enabled.

This also reduces the default bundle size (from 160KB to 60KB gzipped).

In the future, it would probably be useful to ensure Masq keeps disabled on incompatible browsers, such as IE11.

### DIFF
--- a/build/babel.config.js
+++ b/build/babel.config.js
@@ -1,4 +1,8 @@
 module.exports = function (mode) {
+  const plugins = [
+    "@babel/plugin-syntax-dynamic-import"
+  ]
+
   const production = {
     presets :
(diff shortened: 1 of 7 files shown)


amatissart:
This way is much cleaner indeed  !

sujeebant:
The function isLoggedIn has been written to be synchronous, the dynamic loading of Masq makes every function of masq asynchronous, some changes are needed to make this work.

PR #177, opened 2019-04-16
Display error when signalling server is unreachable

sujeebant:


src/modals/masq_favorite_modal.js
```diff
@@ -40,10 +40,11 @@ export default class MasqFavoriteModal {
   }
 
   async openMasq () {
-    await store.login()
-    if (await store.isLoggedIn()) {
-      this.close()
+    try {
+      await store.login()
+    } catch(e) {
```
amatissart: could we at least log a message in these cases, to prevent errors from being unnoticed ?

sujeebant: yes you're right ! :+1:

PR #180, opened 2019-04-18
QMAPS-741 - Dotted walking directions

mlnor27:


src/adapters/scene.js
```diff
@@ -123,6 +123,10 @@ Scene.prototype.initMapBox = function () {
       store.setLastLocation({ lng, lat, zoom })
     })
 
+    this.mb.loadImage(`${baseUrl}statics/images/direction_icons/walking_bullets.png`, (error, image) => {
```
amatissart: * this function `loadImage` breaks the tests as it is not present in "mapbox-gl-js-mock" that we rely on to mock the map. Could you add a mocked implementation in https://github.com/QwantResearch/mapbox-gl-js-mock/blob/master/classes/map.js ?

* Also note that there is a typo in the file name: `walking_bullets` :arrow_right: `walking_bullet`. That causes an error that should be checked before calling `addImage` ;)

src/adapters/scene_direction.js
```diff
@@ -139,24 +139,36 @@ export default class SceneDirection {
...
-          MAIN_ROUTE_COLOR,
-          ALTERNATE_ROUTE_COLOR
-        ],
-        "line-width": 7
+    const geojson = (vehicle === "walking") ? {
+        "id": `route_${route.id}`,
+        "type": "symbol",
+        "source": `source_${route.id}`,
+        "layout": {
+          'icon-image': 'walking_bullet',
```
amatissart: here the same image is used whether the route is active or not. I think we would like to use grey dots on inactive routes, similarly to how a different "line-color" is used below for non-walking vehicle.

src/adapters/scene.js
mlnor27: Here is the PR to make our tests pass https://github.com/QwantResearch/mapbox-gl-js-mock/pull/6 . After the merge, we also need to point to the lastest commit for the dependency `mapbox-gl-js-mock` if i'm not mistaking 👍

src/adapters/scene_direction.js
mlnor27: I added a case expression here 👍 
And whenever a `toggle_route` is triggered, i implemented a `setLayoutProperty` to directly update the `icon-image`

src/adapters/scene_direction.js
```diff
@@ -65,10 +65,12 @@ export default class SceneDirection {
...
     let mainRoute = null;
     this.routes.forEach((route) => {
-      if (route.id === routeId) {
+      const isActive = route.id === routeId
+      if (isActive) {
         mainRoute = route
       }
-      this.map.setFeatureState({source: `source_${route.id}`, id: 1}, {isActive: route.id === routeId})
+      this.map.setFeatureState({source: `source_${route.id}`, id: 1}, {isActive})
+      this.map.setLayoutProperty(`route_${route.id}`, 'icon-image', isActive ? 'walking_bullet_active' : 'walking_bullet_inactive')
```
amatissart: On a car or bicycle route (without `icon-image` layout) this will break the alternative route selection, with an error:
```
TypeError: this._values[t] is undefined
    getValue mapbox-gl.js:29
    getLayoutProperty mapbox-gl.js:29
    setLayoutProperty mapbox-gl.js:33
    setLayoutProperty mapbox-gl.js:33
    setMainRoute scene_direction.js:73
...
```

src/adapters/scene_direction.js
```diff
@@ -151,24 +153,41 @@ export default class SceneDirection {
...
+        "id": `route_${route.id}`,
+        "type": "symbol",
+        "source": `source_${route.id}`,
+        "layout": {
+          'icon-image': [
+            "case",
+            route.isActive,
+            'walking_bullet_active',
+            'walking_bullet_inactive',
+          ],
```
amatissart: this `case` is not necessary as the value is static.  I think it would be clearer to simply use:

```js
"icon-image": route.isActive ? "walking_bullet_active" :  "walking_bullet_inactive"
```

src/adapters/scene.js
```diff
@@ -123,6 +123,14 @@ Scene.prototype.initMapBox = function () {
       store.setLastLocation({ lng, lat, zoom })
     })
 
+    this.mb.loadImage(`${baseUrl}statics/images/direction_icons/walking_bullet_active.png`, (error, image) => {
+      this.mb.addImage("walking_bullet_active", image)
+    })
+
+    this.mb.loadImage(`${baseUrl}statics/images/direction_icons/walking_bullet_inactive.png`, (error, image) => {
+      this.mb.addImage("walking_bullet_inactive", image)
```
amatissart: Please check for the error when loading the images. `image` may be null here.

src/adapters/scene_direction.js
```diff
@@ -65,10 +65,15 @@ export default class SceneDirection {
...
+      const isActive = route.id === routeId
+      if (isActive) {
         mainRoute = route
       }
-      this.map.setFeatureState({source: `source_${route.id}`, id: 1}, {isActive: route.id === routeId})
+      this.map.setFeatureState({source: `source_${route.id}`, id: 1}, {isActive})
+
+      try {
+        this.map.setLayoutProperty(`route_${route.id}`, 'icon-image', isActive ? 'walking_bullet_active' : 'walking_bullet_inactive')
+      } catch (e) {}
```
amatissart: I am not fond of this solution. It may cause errors to stay unnoticed in the future.  
Wouldn't it work to simply check `this.vehicle` value before calling `setLayoutProperty` ?

mlnor27: I replaced the `try { ... } catch(e) {}` with a simple condition and it works just fine

PR #181, opened 2019-04-19
Move layer between water level & important POI labels

Sense:
This PR is linked to the new basic style update in [this other PR](https://github.com/QwantResearch/qwant-basic-gl-style/pull/69)

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -112,8 +112,7 @@ export default class SceneDirection {
         this.showPolygon(route, this.vehicle)
       })
       let mainRoute = this.routes.find((route) => route.isActive)
-      this.map.moveLayer(`route_${mainRoute.id}`)
-
+      this.map.moveLayer(`route_${mainRoute.id}`, "poi-level-street-furniture")
       this.updateMarkers(mainRoute)
 
       const markerOrigin = document.createElement('div')
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -94,14 +94,6 @@ export default class SceneDirection {
     })
     this.markersSteps = []
 
-    if(this.markerOrigin){
-      this.markerOrigin.remove()
-    }
-
-    if(this.markerDestination){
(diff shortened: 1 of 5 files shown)


amatissart:
It looks good, but works only for the route that is initially active. Other calls to `moveLayer` should be modified too : 

routes_road_names

---

SIde note: It could be useful to define the layerId (`"poi-level-street-furniture"` or another layer all routes should be drawn before) as a constant somewhere, for maintainability.

Sense:
> It looks good, but works only for the route that is initially active. Other calls to `moveLayer` should be modified too :
> 
> routes_road_names
> 
> SIde note: It could be useful to define the layerId (`"poi-level-street-furniture"` or another layer all routes should be drawn before) as a constant somewhere, for maintainability.

I defined the `routes_layer` constants in the `config/constants.yml` file under the `map` section.

I also specified to move the direction layer below the `routes_layer` in `setMainRoute` func which is triggered when switching to an alternate possible route.

src/adapters/scene_direction.js
```diff
@@ -65,15 +67,10 @@ export default class SceneDirection {
...
 
   setMainRoute(routeId) {
-    let mainRoute = null;
     this.routes.forEach((route) => {
-      if (route.id === routeId) {
-        mainRoute = route
-      }
       this.map.setFeatureState({source: `source_${route.id}`, id: 1}, {isActive: route.id === routeId})
     })
-    this.updateMarkers(mainRoute)
```
amatissart: The merge conflict has not been correctly fixed: these lines are part from a previous PR (#174) and should not be removed.

Sense: Thank you, just the restored the function as it previously was.

Sense:
> It looks great.
> Could you also update the style version to apply your changes from [QwantResearch/qwant-basic-gl-style#69](https://github.com/QwantResearch/qwant-basic-gl-style/pull/69) ?

I updated the `package.json` and `package-lock.json` files in order to use the updated map style.
Thanks.

PR #182, opened 2019-04-22
Improvements on development workflow

amatissart:
This should remove some source of annoyance during the development process.

* Map-style fonts will not be re-built on `npm install` if the output files already exist in the same version
* Configuration values set by environment variables are now ignored during tests run (no more conflict with test configuration)
* Remove unused dependencies (that included a second install of puppeteer and chromium...)


Sense:
Is this all we can do about the dev workflow improvement or is there anything else to discuss furthermore about ?

amatissart:
Thanks for merging.

I think there is nothing more ongoing about this issue. If you noticed other potential improvements though, your contribution is welcome.

PR #184, opened 2019-04-23
WIP: QMAPS-772 Add category filter at autocomplete

Spasfonx:
The point is to display a "category" item in autocomplete list when the searched terms are matching a simple regex while waiting a more complex NLP system.


Spasfonx:
We cancel that MR, this will be resolved in #188

PR #186, opened 2019-04-24
[QMAPS-776] OSM pois layer toggling according to selected category

Sense:


Spasfonx:
Attendre le merge de la #188, il y a des parties de cette MR qui ont été merge dans cette branche

src/adapters/scene_category.js
```diff
@@ -27,4 +30,13 @@ export default class SceneCategory {
       )}
     )
   }
+
+  removeCategoryMarkers() {
+    this.markers.map(mark => mark.remove())
+    this.setOsmPoisVisibility('visible')
```
Spasfonx: For the visiblity, there is only two possible values (`visible` and `none`) ? If this is the case, there is better to use a boolean for param. With a method like this : `displayOsmPois(displayed)` where `displayed` is a boolean. And this is handled in the method with `const visiblity = displayed ? 'visible' : 'none'`.

If someday the string `visible` change, we only have to change this inside the `displayOsmPois` method 😌

src/panel/category_panel.js
```diff
@@ -50,24 +56,53 @@ export class CategoryPanel {
 
   }
 
+  async open (options = {}) {
+    SearchInput.minify()
+    document.querySelector('#panels').classList.add('panels--hide-services')
```
Spasfonx: This class is added in `open()` method but not removed in `close()` method, is this normal ? Idem for the `.map_bottom_button_group` in the `close()` method

src/panel/category_panel.js
```diff
@@ -50,24 +56,53 @@ export class CategoryPanel {
...
+    PanelManager.resetLayout()
+  }
+
   addCategoryMarkers(){
     fire("add_category_markers", this.pois);
   }
 
   removeCategoryMarkers(){
+    fire("remove_category_markers", this.pois);
     //todo
```
Spasfonx: This is still in todo ?

src/proxies/panel_manager.js
```diff
@@ -76,6 +77,16 @@ PanelManager.openFavorite = async function () {
   })
 }
 
+PanelManager.openCategory = async function (options) {
+  __panel_manager.panels.forEach((panel) => {
```
Spasfonx: Is the `__panel_manager` comme from `window` ? If so, explicit it to avoid misunderstanding of the code

tests/integration/tests/direction.js
```diff
@@ -101,9 +102,10 @@ test('destination', async () => {
 test('origin & destination', async () => {
   expect.assertions(3)
   await page.goto(`${APP_URL}/${ROUTES_PATH}/?origin=latlon:47.4:7.5@Monoprix Nice&destination=latlon:47.4:7.5@Franprix Cannes`)
-
+  await wait(1000)
```
Spasfonx: Usefull ?

src/proxies/panel_manager.js
Sense: Used as is everywhere

src/panel/category_panel.js
Sense: not in todo anymore, thanks.

src/panel/category_panel.js
Sense: indeed we don't need this.

src/adapters/scene_category.js
Sense: yup, ternary operator is coming

PR #188, opened 2019-04-25
QMAPS-772 Add category filter at autocomplete

Spasfonx:
The point is to display a "category" item in autocomplete list when the searched terms are matching a simple regex while waiting a more complex NLP system.

### DIFF
--- a/config/categories.yml
+++ b/config/categories.yml
@@ -1,25 +1,49 @@
 ---
-- label : _('Hotel')
+- label : hotel
+  name: hotel
   color: "#495063"
+  matcher:
+    regex: hotels?|campanile|premi[eè]re classe|b&b hotels|ibis budget|ibis|kyriad|mercure|fasthotel|ace hotel|best western|novotel|balladins|brit hotel|hotel f1|p'tit dej hotel|qualys hotel|holiday inn|ibis style|inter-hotel|akena hotel|citotel|comfort|logis|pullman
   icon: lodging
-- label : _('Restaurant')
+- label : _('restaurant')
+  name: restaurant
   color: "#ef8000"
+  matcher:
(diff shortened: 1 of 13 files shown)


config/categories.yml
```diff
@@ -1,25 +1,49 @@
 ---
-- label : _('Hotel')
```
mlnor27: Isn't it a mistake here to replace `_('Hotel')` by `hotel` ?

src/adapters/suggest.js
```diff
@@ -37,17 +38,29 @@ export default class Suggest {
...
+
+            /* 'bbox' is currently not used by the geocoder, it' will be used for the telemetry. */
             this.historyPromise = PoiStore.get(term)
             this.bragiPromise = BragiPoi.get(term)
+            this.categoryPromise = withCategories ? CategoryService.getMatchingCategories(term) : null
+
             try {
-              let [bragiResponse, storeResponse] = await Promise.all([this.bragiPromise, this.historyPromise])
-              if (bragiResponse !== null) {
+              let [bragiResponse, storeResponse, categoryResponse] = await Promise.all([
```
mlnor27: you could use `const` as there is no reassignement for those values

src/adapters/suggest.js
```diff
@@ -169,14 +192,32 @@ export default class Suggest {
...
-</div>
-`
+      <div class="autocomplete_suggestion${fromHistory ? ' autocomplete_suggestion--history' : ''}" data-id="${id}" data-val="${ExtendedString.htmlEncode(poi.getInputValue())}">
+        ${this.renderLines(iconDom, name, alternativeName)}
+      </div>
+    `
+  }
+
+  renderCategory(category) {
+    let { label, alternativeName, color, backgroundColor } = category
```
mlnor27: 3 `const`s could be used instead of 3 `let`s

config/categories.yml
Spasfonx: No, there is already an existing key named `hotel` in the po, so that's why trads won't work in the branch 709 for the categories

PR #190, opened 2019-04-29
QMAPS-773 Refresh current active category POIS on map moveend

Sense:


src/proxies/panel_manager.js
```diff
@@ -89,6 +89,14 @@ PanelManager.openCategory = async function (options) {
   })
 }
 
+PanelManager.refreshCategoryPois = async function () {
```
Spasfonx: Instead of looping over all the panels, it will be better to fire an event when the move is finished (in `scene.js`) and listen that event on `CategoryPanel`. With that we can plug anything on that event later if we need and this avoid to loop over every panels on every moves.

Sense: Done

PR #197, opened 2019-05-02
QMAPS-775 - little refactor of markers

mlnor27:
The last PR was creating an icon using html and css but utilizing small pixels sizes wasn't rendering the icon with a look as good as an `svg` image. And the implementation took a little bit more code.
So in this PR, i replaced the `img` element with a `background-image` and trimmed a good portion of the css 👍

### DIFF
--- a/src/scss/includes/app.scss
+++ b/src/scss/includes/app.scss
@@ -108,95 +108,52 @@ BODY {
   }
 }
 
-.map_container .poi-marker {
-  width: 36px;
-  height: 42px;
-  background: transparent;
-  position: relative;
+.map_container .marker-container {
   display: flex;
+  align-items: center;
   justify-content: center;
-  pointer-events: none;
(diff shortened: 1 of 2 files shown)


src/scss/includes/app.scss
```diff
@@ -108,95 +108,52 @@ BODY {
...
-  transform: translateY(4px);
-  cursor: pointer;
-}
-
-.map_container .poi-hoverable:hover,
-.map_container .poi-hoverable.active {
-  z-index: 1;
+  width: 35px;
+  height: 42px;
+  background-image: url('/statics/images/map/pin_map.svg');
```
amatissart: we need to use relative URLs in css files, to handle potential prefix used by the deployed app (/maps/...)

mlnor27: whoops, i forgot that.., i'll fix it

PR #198, opened 2019-05-03
Telemetry log events

Sense:


src/panel/category_panel.js
```diff
@@ -111,10 +113,9 @@ export default class CategoryPanel {
   }
 
   selectPoi(poi){
-    // TODO telemetry
     fire('fit_map', poi, layouts.LIST)
     this.close(false)
-    PanelManager.loadPoiById(poi.id, {isFromList : true, list: this})
+    PanelManager.loadPoiById(poi.id, {isFromList : true, list: this, source: 'pagesjaunes'})
```
mlnor27: Wouldn't it be better if we create a constant object for the poi sources ?

src/panel/poi_panel.js
```diff
@@ -139,12 +149,20 @@ PoiPanel.prototype.setPoi = async function (poi, options = {}) {
 }
 
 PoiPanel.prototype.center = function() {
-  Telemetry.add(Telemetry.POI_GO)
+  if (this.poiSource && this.poiSource == 'pagesjaunes') {
```
mlnor27: Following my last comment, now that i saw more code, a constant should be used to reference the poi sources 👍

src/panel/category_panel.js
```diff
@@ -111,10 +114,9 @@ export default class CategoryPanel {
   }
 
   selectPoi(poi){
-    // TODO telemetry
     fire('fit_map', poi, layouts.LIST)
     this.close(false)
-    PanelManager.loadPoiById(poi.id, {isFromList : true, list: this})
+    PanelManager.loadPoiById(poi.id, {isFromList : true, list: this, source: constants.telemetry.source.PJ})
```
amatissart: `source` should be read from the API response. The category panel may be filled with POIs from different sources (depending on the location, the category, etc.)

Sense: Ok @amatissart, I don't see any `source` attribute coming from the API response in the IdunnPOI object. I don't think extracting it from the `id` would be a sustainable solution.

src/panel/category_panel.js
Sense: Indeed, done

src/panel/poi_panel.js
Sense: We're currently talking about using the API response to set the poiSource.

PR #201, opened 2019-05-03
Masq: add modal while activating Masq

sujeebant:


src/views/masq_activating_modal.dot
```diff
@@ -0,0 +1,40 @@
...
+    <div class="icon-masq-by-qwant modal__masq_activating__logo"></div>
+    <i class="icon-x modal__masq_activating__close" {{= click(this.close, this) }}></i>
+  </div>
+
+  <div class="modal__masq_activating__body">
+    {{? this.activating}}
+      <div class="modal__masq_activating__animation">
+        <video autoplay loop src="{{= window.location.origin }}/statics/images/masq/gifs/masq_loading.mp4"></video>
+      </div>
+      <h2 class="modal__masq_activating__message">{{= _('Patience, Masq s\'active...') }}</h2>
```
amatissart: this translation key needs to be defined in english :)

sujeebant: :+1:

src/modals/masq_activating_modal.js
```diff
@@ -0,0 +1,44 @@
...
+    // get masqErrorModal from window if already initialized
+    if (window.__masqActivatingModal) {
+      return window.__masqActivatingModal
+    }
+    // if masqErrorModal not initialized, use this
+    window.__masqActivatingModal = this
+
+    this.modal = new Modal(this, MasqActivatingModalView, false)
+
+    this.opened = false
```
amatissart: this flag is not used. And as `this.activating` is true by default, the modal and the video are present in DOM at app start. (Chromium at least tries to load the .mp4 before it is visible)

src/views/masq_activating_modal.dot
```diff
@@ -0,0 +1,40 @@
+<div class="modal__masq_activating">
+  <div class="modal__masq_activating__header">
+    <div class="icon-masq-by-qwant modal__masq_activating__logo"></div>
+    <i class="icon-x modal__masq_activating__close" {{= click(this.close, this) }}></i>
+  </div>
+
+  <div class="modal__masq_activating__body">
+    {{? this.activating}}
+      <div class="modal__masq_activating__animation">
+        <video autoplay loop src="{{= window.location.origin }}/statics/images/masq/gifs/masq_loading.mp4"></video>
```
amatissart: Using "origin" is incorrect here: the "baseUrl" ( /maps/...) is missing.  
Actually an absolute path can directly be used in templates, since a `<base>` tag is defined in the html head: 
```suggestion
        <video autoplay loop src="/statics/images/masq/gifs/masq_loading.mp4"></video>
```

sujeebant: :+1:

src/modals/masq_activating_modal.js
sujeebant: Fixed !
removed opened flag and made activating false by default

PR #202, opened 2019-05-06
scene_category: Cleaning old markers when new markers arrive

mlnor27:


src/adapters/scene_category.js
```diff
@@ -8,7 +8,11 @@ export default class SceneCategory {
     this.markers = []
 
     listen('add_category_markers', (pois) => {
-      this.addCategoryMarkers(pois);
+      this.markers = this.markers.reduce((accumulator, marker) => {             
+        marker.remove()
+        return accumulator
+      }, [])
+      this.addCategoryMarkers(pois)
```
xem: Well, it works but seems a bit overkill... why not just a for loop or a forEach to remove the markers?

mlnor27: i finally changed it to deduplicate with another piece of code

PR #203, opened 2019-05-06
add: label under category marker

Spasfonx:


src/scss/includes/app.scss
```diff
@@ -116,15 +116,37 @@ BODY {
   height: 42px;
   background-image: url('../images/map/pin_map.svg');
   background-size: cover;
+  margin: auto;
 }
 
+.map_container .marker-label {
+  line-height: 20px;
+  margin-bottom: -20px;
```
Spasfonx: Is this usefull anymore the `margin-bottom` rule ?

src/adapters/icon_manager.js
```diff
@@ -51,13 +51,17 @@ export default class IconManager {
...
+export function createIcon(iconOptions, name, hoverEffect = false) {
   const icon = IconManager.get(iconOptions)
 
+  const markerLabel = name ? `<div class="marker-label">${name}</div>` : ''
   const element = document.createElement('div')
   element.innerHTML = `
-    <div class="marker-container${hoverEffect ? ' poi-hoverable' : ''}">
-      <i class="icon icon-${icon.iconClass}"></i>
+    <div class="marker">
+      <div class="marker-container${hoverEffect ? ' poi-hoverable' : ''}">
```
Sense: The string template can be simplified by something like : ``${hoverEffect && ' poi-hoverable'}``

Spasfonx: Unfortunaly we're not in jsx, so it doesn't work 😬 if we do this, when `hoverEffect` is `true` it work as expected, but if `false`, we have `false` instead of `''` in the value.
```
`marker-container${true && ' poi-hoverable'}` // "marker-container poi-hoverable"
`marker-container${false && ' poi-hoverable'}` // "marker-containerfalse"
```

PR #205, opened 2019-05-06
scene - Keeping place marker after page reload

mlnor27:


src/adapters/scene.js
```diff
@@ -154,6 +154,15 @@ Scene.prototype.initMapBox = function () {
 
     window.execOnMapLoaded = (f) => f()
     fire('map_loaded')
+
+    const place = UrlShards.parseUrl().filter(shard => shard.prefix === 'place')
+    if (place.length) {
+      const { value } = place[0]
+      const poi = await PanelManager.loadPoiById(value.split('@')[0])
```
Sense: Careful @mlnor27 it may throws something like `TypeError: Cannot read property '0' of undefined`

PR #207, opened 2019-05-07
fix: every requests in suggest are now correctly handled

Spasfonx:
The bug was when we type in search bar a category name quickly (like "restaurant", or "education" or "supermarket" something long enough), sometimes items are not loaded.

This fix seems to fix the bug (to be honest i don't know why, but it work).

Feel free to test on your device ! (think to clear the "cache" used in window object : `window = null` in developer console and reload)

### DIFF
--- a/src/adapters/suggest.js
+++ b/src/adapters/suggest.js
@@ -79,7 +79,7 @@ export default class Suggest {
         pois.forEach((poi) => {
           if (poi instanceof PoiStore) {
             favorites.push(poi)
-          } if (poi instanceof  Category) {
+          } else if (poi instanceof  Category) {
             categories.push(poi)
           } else {
             remotes.push(poi)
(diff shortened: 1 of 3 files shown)


src/adapters/suggest.js
```diff
@@ -50,14 +48,16 @@ export default class Suggest {
...
                 this.suggestList = this.suggestList.concat(categoryResponse)
 
-              if (bragiResponse) {
-                this.bragiPromise = null
-                this.suggestList = this.suggestList.concat(bragiResponse)
-              }
-
+              this.bragiPromise = null
+              this.suggestList = this.suggestList.concat(bragiResponse)
               this.suggestList = this.suggestList.concat(storeResponse)
```
amatissart: it looks good to me.

Not directly related to this change: "favorite" POIs from store are rendered twice in the suggestions, because a `else if` is missing in the rendering function (line 82)

Spasfonx: oh yes, a missmerge i think, thanks !

PR #208, opened 2019-05-15
QMAPS-806 - Implemented the 'no results' panel

mlnor27:


src/scss/includes/panels/categories.scss
```diff
@@ -177,3 +173,56 @@
...
+
+.category__panel__notfound {
+  position: relative;
+  width: 400px;
+  height: 182px;
+  border-radius: 3px;
+  box-shadow: 0 0 4px 0 rgba(0, 0, 0, 0.16);
+  background-color: #ffffff;
+  display: flex;
+  flex-direction: column;
```
amatissart: we need to be cautious when using "flex", as we want to support IE11. Could you check it works as expected ?

mlnor27: @amatissart  Sorry i couldn't successfully test it on ie11. I tried browserstack but it seems to give a false overview of the reality due to whatever reasons. I also tried adding certificates to ensure every network request doesn't fail but it doesn't seem to work.
I also would like to mention that some of my previous PRs included flexboxes and they've been merged without any discussions regarding ie11 support.

PR #211, opened 2019-05-16
QMAPS-805 - Keep searchbar when the category panel is up

mlnor27:


src/panel/category_panel.js
```diff
@@ -69,10 +69,9 @@ export default class CategoryPanel {
   }
 
   async open (options = {}) {
-    SearchInput.minify()
-    document.querySelector('.top_bar').classList.add('top_bar--category-open')
     if(options.category) {
       this.categoryName = options.category.name
+      SearchInput.setInputValue(options.category.name)
```
amatissart: category `label` should be used here instead of `name` which is not translated.

src/ui_components/search_input.js
```diff
@@ -35,6 +35,10 @@ export default class SearchInput {
     window.__searchInput.searchInputHandle.blur()
   }
 
+  static setInputValue(value) {
+    document.getElementById('search').value = value
```
amatissart: for maintainability, we could use the existing suggest `setValue` method
```suggestion
    window.__searchInput.suggest.setValue(value)
```

PR #216, opened 2019-05-21
Fixed regression to highlight markers when hovering poi listing panel

mlnor27:


amatissart:
these changes are now included in #209

PR #217, opened 2019-05-22
QMAPS-810 - Pictures section inside poi_panel

mlnor27:


src/views/poi_panel.dot
```diff
@@ -80,6 +80,48 @@
...
+        let pictures;
+        const id = this.poi.id.replace('pj:', '');
+        const imagesBlock = this.poi.blocks.find((block) => block.type === 'images');
+        if (imagesBlock) {
+          pictures = imagesBlock.images.filter(img => img.url !== this.poi.topImageUrl);
+        }
+      }}
+
+      {{? pictures && pictures.length }}
+        <div class="poi_panel__pictures">
```
amatissart: This block should use the mechanism that already exists to render all block from separate templates.
See existing files: `images_panel.js` and `images.dot` (currently empty)

src/views/poi_panel.dot
```diff
@@ -80,6 +80,48 @@
...
+            <h5 class="poi_panel__pictures_title">
+              <i class="icon-icon_camera"></i>
+              {{= _('Photos')}}
+            </h5>
+            <a href="https://www.pagesjaunes.fr/pros/{{= id}}" target="_blank" rel="noopener noreferrer" style="text-decoration: none">
```
amatissart: The same here, this URL format should not be hardcoded.  
The API provides a `source_url` for each image, that could be used as a link instead.

src/views/poi_panel.dot
```diff
@@ -80,6 +80,48 @@
...
+              <div class="poi_panel__pictures_block" style="background-image: url('{{= pictures[0].url}}')">
```
amatissart: This breaks if the URL contains a single quote `'`.  
`&quot;` should be used instead for all images, see `header.dot` for an example.
  
Maybe we could define a function somewhere to build this style property and avoid this bug in the future ?

src/views/poi_panel.dot
```diff
@@ -80,6 +80,48 @@
 
       {{= this.PoiBlocContainer.render(this.poi) }}
 
+      {{
+        let pictures;
+        const id = this.poi.id.replace('pj:', '');
```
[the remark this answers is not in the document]
mlnor27: It turns out that this line isn't useful anymore

src/views/poi_panel.dot
mlnor27: Ahh, i thought we had to link the whole picture block to the poi's page jaune profil

src/views/poi_panel.dot
mlnor27: I thought it would be escaped..

src/panel/poi_bloc/images_panel.js
```diff
@@ -2,9 +2,20 @@ import ImagesPanelView from '../../views/poi_bloc/images.dot'
 import Panel from "../../libs/panel";
 
 export default class ImagesPanel {
-    constructor(block){
-      this.panel = new Panel(this, ImagesPanelView)
-      this.name = block.name
-      this.block = block
+  constructor() {
```
amatissart: this constructor receives all necessary details to render the block : 
```suggestion
  constructor(block, poi, options) {
```

src/panel/poi_bloc/images_panel.js
```diff
@@ -2,9 +2,20 @@ import ImagesPanelView from '../../views/poi_bloc/images.dot'
...
+    this.panel = new Panel(this, ImagesPanelView)
+    this.images = []
+  }
+
+  set({ blocks, topImageUrl }) {
```
amatissart: this method is not necessary, everything could be handled in the constructor.

src/panel/poi_panel.js
```diff
@@ -33,6 +34,7 @@ function PoiPanel(sharePanel) {
   this.card = true
   this.headerPartial = headerPartial
   this.minimalHourPanel = new MinimalHourPanel()
+  this.imagesPanel = new ImagesPanel()
```
amatissart: It should not be necessary to define this panel here, the images block could be handled automatically by the `PoiBlocContainer`

src/views/poi_panel.dot
```diff
@@ -80,6 +80,8 @@
 
       {{= this.PoiBlocContainer.render(this.poi) }}
 
+      {{= this.imagesPanel.set(this.poi).render() }}
```
amatissart: The same here, this should be handled by the `PoiBlocContainer` above.

src/scss/includes/panels/poi_panel.scss
```diff
@@ -380,6 +380,73 @@ $HEADER_SIZE: 30px;
...
 }
 
+.poi_panel__pictures {
+  margin-bottom: 20px;
+}
+
+.poi_panel__pictures_title {
+  border-top: 1px solid #e0e1e6;
+  padding: 15px 0;
+  font-family: Asap;
```
G-Ray: Isn't the font already defined somewhere for all the app ?

PR #220, opened 2019-05-23
Fix category selection via enter key

amatissart:
`Suggest.onSubmit` calls `onSelect` directly. 

As a consequence, all suggestions items (both Poi and Category) need to be handled by `SearchInput.selectItem`

### DIFF
--- a/src/adapters/osm_schedule.js
+++ b/src/adapters/osm_schedule.js
@@ -31,7 +31,7 @@ function scheduleStatus(scheduleResponse, timeMessages) {
  * Both params may be null (eg for 24/7 places)
  */
 function nextTransitionTime(seconds, nextTransitionDate) {
-  if(seconds < 12 * 60 * 60) {
+  if (nextTransitionDate && seconds < 12 * 60 * 60) {
     /*
        extract local time from nextTransitionDate
        "2019-05-12T18:00:00+02:00" => "18:00:00"
--- a/src/adapters/osm_schedule.js
+++ b/src/adapters/osm_schedule.js
@@ -24,6 +24,12 @@ function scheduleStatus(scheduleResponse, timeMessages) {
   return {msg : '', color : '#fff'}
 }
 
+/**
+ * Format next transition time
+ * @param seconds The nb of seconds before next transition
+ * @param nextTransitionDate Datetime of next transition
+ * Both params may be null (eg for 24/7 places)
+ */
 function nextTransitionTime(seconds, nextTransitionDate) {
   if(seconds < 12 * 60 * 60) {
     /*
--- a/src/adapters/suggest.js
+++ b/src/adapters/suggest.js
@@ -93,25 +93,16 @@ export default class Suggest {
         return suggestDom
       },
 
-      onSelect: (e, term, item, items) => {
+      onSelect: (e, term, item, items = []) => {
         e.preventDefault()
-        const type = item.getAttribute('data-type')
-        const val = item.getAttribute('data-val')
         const itemId = item.getAttribute('data-id')
 
-        if ('category' === type) {
-          PanelManager.openCategory({
-            category: CategoryService.getCategoryByName(val)
-          })
-        }
-        else {
-          let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
-          if(prefixPoint) {
-            this.onSelect(prefixPoint)
-          } else {
-            let poi = items.find(poi => poi.id === itemId)
-            this.onSelect(poi)
-          }
+        let prefixPoint = this.prefixes.find(prefix => prefix.id === itemId)
+        if (prefixPoint !== undefined) {
+          this.onSelect(prefixPoint)
+        } else {
+          let selectedItem = items.find(item => item.id === itemId)
+          this.onSelect(selectedItem)
         }
         this.searchInputDomHandler.blur()
       }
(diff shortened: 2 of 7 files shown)


src/adapters/suggest.js
```diff
@@ -95,23 +95,14 @@ export default class Suggest {
...
-        else {
-          let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
-          if(prefixPoint) {
-            this.onSelect(prefixPoint)
-          } else {
-            let poi = items.find(poi => poi.id === itemId)
-            this.onSelect(poi)
-          }
+        let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
+        if(prefixPoint) {
```
GuillaumeGomez: nit: missing whitespace before '('.

src/adapters/suggest.js
```diff
@@ -95,23 +95,14 @@ export default class Suggest {
...
-        }
-        else {
-          let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
-          if(prefixPoint) {
-            this.onSelect(prefixPoint)
-          } else {
-            let poi = items.find(poi => poi.id === itemId)
-            this.onSelect(poi)
-          }
+        let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
```
GuillaumeGomez: Just an idea: comparisons in JS are faster when they are strict. Why not using `findIndex` and then compare like this: `if (prefixPoint !== -1) {`?

src/adapters/suggest.js
```diff
@@ -95,23 +95,14 @@ export default class Suggest {
...
+        if(prefixPoint) {
+          this.onSelect(prefixPoint)
+        } else {
+          let selectedItem = items.find(item => item.id === itemId)
```
GuillaumeGomez: Don't you check if you found something in here?

GuillaumeGomez: Or if `items` actually exists?

src/proxies/panel_manager.js
```diff
@@ -93,7 +93,7 @@ PanelManager.openCategory = async function (options) {
   __panel_manager.panels.forEach((panel) => {
     if(panel instanceof CategoryPanel) {
       panel.open(options)
-    } else if(panel.active && !(panel instanceof CategoryPanel)) {
+    } else if(panel.active) {
```
GuillaumeGomez: Missing whitespace after `if`.

src/ui_components/search_input.js
```diff
@@ -80,11 +82,14 @@ export default class SearchInput {
     }
   }
 
-  async selectItem (selectedPoi) {
-    if(selectedPoi) {
-      fire('fit_map', selectedPoi, selectedPoi.type === 'poi' ? layouts.POI : layouts.FULL)
-      fire('map_mark_poi', selectedPoi)
-      PanelManager.loadPoiById(selectedPoi.id)
+  async selectItem (selectedItem) {
+    if(selectedItem instanceof Poi) {
```
GuillaumeGomez: Missing whitespace after `if`.

src/ui_components/search_input.js
```diff
@@ -80,11 +82,14 @@ export default class SearchInput {
...
+      fire('fit_map', selectedItem, selectedItem.type === 'poi' ? layouts.POI : layouts.FULL)
+      fire('map_mark_poi', selectedItem)
+      PanelManager.loadPoiById(selectedItem.id)
+    }
+    else if(selectedItem instanceof Category){
```
GuillaumeGomez: Missing whitespace after `if`.

src/adapters/suggest.js
amatissart: I made a change to use `if (prefixPoint !== undefined)`  
This function being called a few times per session, performance is probably not significant here.

src/adapters/suggest.js
amatissart: a default value for `items` has been added as a precaution, thanks

src/adapters/suggest.js
```diff
@@ -93,25 +93,16 @@ export default class Suggest {
...
-        }
-        else {
-          let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
-          if(prefixPoint) {
-            this.onSelect(prefixPoint)
-          } else {
-            let poi = items.find(poi => poi.id === itemId)
-            this.onSelect(poi)
-          }
+        let prefixPoint = this.prefixes.find((prefix) => prefix.id === itemId)
```
GuillaumeGomez: `(prefix)` shouldn't be surrounded with parens.

src/proxies/panel_manager.js
```diff
@@ -91,9 +91,9 @@ PanelManager.getCategoryPanel = function () {
 
 PanelManager.openCategory = async function (options) {
   __panel_manager.panels.forEach((panel) => {
-    if(panel instanceof CategoryPanel) {
+    if (panel instanceof CategoryPanel) {
       panel.open(options)
-    } else if(panel.active && !(panel instanceof CategoryPanel)) {
+    } else if (panel.active) {
```
GuillaumeGomez: Please `panel.active === true`.

src/ui_components/search_input.js
```diff
@@ -82,15 +83,14 @@ export default class SearchInput {
...
-        fire('map_mark_poi', selectedPoi)
-        PanelManager.loadPoiById(selectedPoi.id)
-      }
+  async selectItem (selectedItem) {
+    if (selectedItem instanceof Poi) {
+      fire('fit_map', selectedItem, selectedItem.type === 'poi' ? layouts.POI : layouts.FULL)
+      fire('map_mark_poi', selectedItem)
+      PanelManager.loadPoiById(selectedItem.id)
+    }
+    else if (selectedItem instanceof Category){
```
GuillaumeGomez: Please put the `else if` just after the `}` (on the same line).

src/adapters/osm_schedule.js
```diff
@@ -25,7 +25,7 @@ function scheduleStatus(scheduleResponse, timeMessages) {
 }
 
 function nextTransitionTime(seconds, nextTransitionDate) {
-  if(seconds < 12 * 60 * 60) {
+  if (nextTransitionDate && seconds < 12 * 60 * 60) {
```
GuillaumeGomez: Please make explicit comparison for `nextTransitionDate` (or a comment since it seems difficult to check what's going on here).

PR #223, opened 2019-05-27
QMAPS-814 - Correct PJ Styling (first half)

Sense:


src/views/poi_panel.dot
```diff
@@ -55,7 +55,7 @@
         <div class="poi_panel__actions__store_container button_container" {{= click(this.toggleStorePoi, this) }}>
           <button class="poi_panel__actions__icon__store poi_panel__actions__icon {{= this.poi.stored ? 'icon-icon_star-filled' : 'icon-icon_star'}}">
           </button>
-          <div class="poi_panel__actions__text">{{= _('FAVORITES', 'poi') }}</div>
+          <div class="poi_panel__actions__text">{{? this.poi.stored }} {{= _('ENREGISTRÉ', 'poi') }} {{??}} {{= _('FAVORITES', 'poi') }} {{?}}</div>
```
mlnor27: you could write it like this `{{= this.poi.stored ? _('ENREGISTRÉ', 'poi') : _('FAVORITES', 'poi') }}`, it's a bit shorter

src/views/category_panel.dot
```diff
@@ -49,7 +49,7 @@
               {{?}}
 
               {{? grades }}
-                <div class="panel__note">
+                <div class="panel__note" {{= click(this.goToFeedbackUrl, this, poi) }}>
```
mlnor27: Wouldn't it be semantically more correct to use an `a` tag here and build the `href` dynamically ?

src/views/poi_panel.dot
Sense: Indeed you right

src/views/category_panel.dot
Sense: There are some logic instructions behind the `feedbackUrl` that needs to be scoped into a function in my opinion, instead of splitting strings directly in the template 😬

PR #224, opened 2019-05-28
QMAPS-814 - Correct PJ style

Sense:


src/panel/category_panel.js
```diff
@@ -136,4 +136,12 @@ export default class CategoryPanel {
       marker.classList.remove("active")
     }
   }
+
+  goToFeedbackUrl(poi) {
+    let pj_id;
```
GuillaumeGomez: What do you think about:

```js
goToFeedbackUrl(poi) {
  if (poi.id.indexOf('pj:') === -1) {
    return
  }
  let pj_id = poi.id.split('pj:')[1]
  window.open(`https://www.pagesjaunes.fr/pros/${pj_id}#ancreBlocAvis`)
```

?

src/panel/poi_panel.js
```diff
@@ -203,6 +204,14 @@ PoiPanel.prototype.emptyClickOnMap = function() {
   }
 }
 
+PoiPanel.prototype.goToFeedbackUrl = function() {
```
GuillaumeGomez: This function is exactly the same as the one in `category_panel.js`. Couldn't you factorize it instead of redeclaring it?

src/panel/category_panel.js
Sense: That's indeed a neat way to do it.

src/panel/poi_panel.js
Sense: Moved this func in the Poi class 👍

PR #225, opened 2019-05-28
Qmaps-824 zoom fixes

xem:


src/adapters/poi/idunn_poi.js
```diff
@@ -58,6 +58,9 @@ export default class IdunnPoi extends Poi {
       if(err === 404) {
         return
       }
+      else if(err === 400){
```
GuillaumeGomez: Why not combining this condition with the previous one considering there is no specific treatment?

src/panel/category_panel.js
```diff
@@ -52,15 +52,15 @@ export default class CategoryPanel {
   }
 
   async search() {
-    this.loading = true
+    this.loading = true;
```
GuillaumeGomez: Just curiosity: why adding the semicolon when there are none in the rest of the code?

src/panel/category_panel.js
```diff
@@ -52,15 +52,15 @@ export default class CategoryPanel {
...
     let bbox = window.map.bbox()
     let urlBBox = [bbox.getWest(),bbox.getSouth(),bbox.getEast(),bbox.getNorth()]
       .map((cardinal) => cardinal.toFixed(7))
       .join(',')
 
-    this.pois = await IdunnPoi.poiCategoryLoad(urlBBox, 50, this.categoryName)
-    
-    this.loading = false
+    this.pois = await IdunnPoi.poiCategoryLoad(urlBBox, 50, this.categoryName, this)
+    this.loading = false;
```
GuillaumeGomez: Same question.

src/panel/category_panel.js
```diff
@@ -82,6 +82,23 @@ export default class CategoryPanel {
     this.search()
     await this.panel.update()
     UrlState.pushUrl()
+
+    // Apply minimal zoom when opening a category
+
+    // Zoom < 5: focus on Paris
+    if(window.map.mb.getZoom() < 5){
```
GuillaumeGomez: Wouldn't a `switch` be more appropriate in here? It'd allow to call `window.map.mb.getZoom()` only once.

src/panel/category_panel.js
xem: sorry, reflex

src/panel/category_panel.js
xem: I agree to calling getZoom only once, but not for the switch. Switches are not ideal to test < / > conditions...  right?

PR #229, opened 2019-05-29
QMAPS-815 - Use a limit for autocomplete suggestions

Sense:


Sense:
> Tests should be changed I guess

@jbgriesner I solved the failing test 😎

Sense:
Indeed you right @sujeebant Favorites has to be rendered as last items of the suggest list.

> Screenshot from 2019-06-04 11-16-41
> Favorites are now placed on the top of the list, we do not see where favorites end and where suggestions begin.

src/adapters/suggest.js
```diff
@@ -76,17 +76,20 @@ export default class Suggest {
         let remotes = []
         let categories = []
         pois.forEach((poi) => {
-          if (poi instanceof PoiStore) {
+          if (poi instanceof PoiStore && favorites.length < 2) {
+            // 2 favorites pois max
             favorites.push(poi)
```
amatissart: With this implementation, extra favorites (if length >2) will be pushed to `remotes`.  
For example on a empty search bar, all favorites are returned, but only 2 are described as "favorites":
image

Actually we may want to display more than 2 favorites in that case (think about choosing a favorite address when selecting an origin or a destination on the directions panel)

src/adapters/suggest.js
```diff
@@ -76,17 +76,20 @@ export default class Suggest {
...
-          } else if (poi instanceof  Category) {
+          } else if (poi instanceof Category && categories.length === 0) {
+            // 1 category pois max
             categories.push(poi)
```
amatissart: the same here

src/adapters/suggest.js
Sense: I made some updates in order to avoid getting favorites into remotes.
I got some new guidelines like :
- 5 favorites max will be displayed on empty query
- when query is not empty :
display 7 items maximum
2 favorites maximum
1 category maximum
remaining items has to be filled with remotes
- currently working on displaying addresses on second-line's item

src/adapters/suggest.js
Sense: outdated now

PR #231, opened 2019-06-04
QMAPS-822 - Save & Restore location on Poi toggling

Sense:


src/panel/poi_panel.js
```diff
@@ -89,6 +89,7 @@ PoiPanel.prototype.closeAction = function() {
   if (this.poi.meta && this.poi.meta.source) Telemetry.add("close", "poi", this.poi.meta.source)
   fire('clean_marker')
   PanelManager.resetLayout()
+  fire("restore_location")
```
amatissart: I am not sure we should try to restore a location on close. Especially when the poi panel is not related to any list.

src/adapters/scene.js
```diff
@@ -178,6 +179,25 @@ Scene.prototype.initMapBox = function () {
...
+}
+
+Scene.prototype.saveLocation = function() {
+  let locationShard = UrlShards.parseUrl().find(shard => shard.prefix === 'map')
+  this.savedLocation = locationShard.value
+}
+
+Scene.prototype.restoreLocation = function() {
+  let flyOptions = {center : this.urlCenter, zoom : this.urlZoom, animate: true, screenSpeed: 1.5}
+  this.restore(this.savedLocation)
```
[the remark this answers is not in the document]
mlnor27: yes you're right. I made the changes

src/panel/poi_panel.js
mlnor27: Sounds fare. I removed it.

PR #232, opened 2019-06-05
Qmaps 811

xem:
show tooltips on PJ markers hover

### DIFF
--- a/src/adapters/poi_popup.js
+++ b/src/adapters/poi_popup.js
@@ -13,16 +13,14 @@ const WAIT_BEFORE_DISPLAY = 350
 
 function PoiPopup() {}
 
-
-
 PoiPopup.prototype.init = function(map) {
   this.map = map
   this.popupHandle = null
   this.timeOutHandler = null
   this.sceneState = SceneState.getSceneState()
 
-  listen("open_popup", (poi, e)=>this.createPJPopup(poi, e));
-  listen("close_popup", ()=>this.close());
+  listen("open_popup", (poi, e) => this.createPJPopup(poi, e));
+  listen("close_popup", () => this.close());
 
 }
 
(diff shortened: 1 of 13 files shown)


src/scss/includes/popup.scss
```diff
@@ -1,4 +1,8 @@
 .poi_popup__container {
+
+  position: absolute;
+  z-index: 6;
```
mlnor27: I think we have a dedicated file for `z-index`es, you could eventually put it there to be in adequation with the purpose of this file 👍

src/adapters/poi_popup.js
```diff
@@ -42,39 +46,49 @@ PoiPopup.prototype.addListener = function(layer) {
...
   let poi = await ApiPoi.poiApiLoad(layerPoi.properties.global_id, {simple : true})
   if(poi) {
-    if(this.popupHandle) {
-      this.popupHandle.remove()
-      this.popupHandle = null
-    }
-    let {color} = IconManager.get(poi)
-    let category = poiSubClass(poi.subClassName)
-    let hours = poi.blocks.find(block =>
+    this.showPopup(poi, event);
```
mlnor27: Sorry, It looks like we do not use semicolons in this project. It is pretty unpleasing to make the devs go back to their PRs just for those kind of things. Maybe we could introduce a code formatter as a precommit hook in the future or eslint

src/adapters/poi_popup.js
```diff
@@ -42,39 +46,49 @@ PoiPopup.prototype.addListener = function(layer) {
...
+  }
+}
+
+PoiPopup.prototype.createPJPopup =  function (poi, event) {
+  if(poi) {
+    this.showPopup(poi, event);
```
mlnor27: Again, for the semi-colon 👍

src/adapters/poi_popup.js
xem: ;)

PR #238, opened 2019-06-10
Beta is coming...

amatissart:
* Fix attribution/scale positioning 
* Upgrade to mapbox-gl 1.0.0 (with several optimizations, bugfix on Safari, etc)
* Implement a version flag in configuration (its values is now "Beta")

### DIFF
--- a/src/adapters/poi_popup.js
+++ b/src/adapters/poi_popup.js
@@ -129,7 +129,9 @@ PoiPopup.prototype.setPopupPosition = function(event, popupOptions) {
 };
 
 PoiPopup.prototype.close = function() {
-  this.popupHandle.remove();
+  if (this.popupHandle) {
+    this.popupHandle.remove();
+  }
 };
 
 /* private */
--- a/src/panel/direction/direction_panel.js
+++ b/src/panel/direction/direction_panel.js
@@ -198,7 +198,7 @@ export default class DirectionPanel {
     this.initDirection();
     UrlState.pushUrl();
     window.execOnMapLoaded(() => {
-      this.searchDirection()
+      this.searchDirection();
     });
   }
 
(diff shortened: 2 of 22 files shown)


src/scss/includes/direction.scss
```diff
@@ -692,9 +692,10 @@
     color: $background;
     line-height: 17px;
     position: absolute;
-    left: 5px;
-    top: 10px;
+    left: 0px;
```
GuillaumeGomez: In case of `0px`, it's better to juste write `0`.

src/scss/includes/direction.scss
```diff
@@ -692,9 +692,10 @@
...
+    top: 0px;
```
GuillaumeGomez: Same.

src/adapters/poi_popup.js
```diff
@@ -129,7 +129,9 @@ PoiPopup.prototype.setPopupPosition = function(event, popupOptions) {
 };
 
 PoiPopup.prototype.close = function() {
-  this.popupHandle.remove();
+  if (this.popupHandle) {
+    this.popupHandle.remove();
```
GuillaumeGomez: Shouldn't you change `this.popupHandle` value to avoid it being still "available"?

src/panel/direction/direction_panel.js
```diff
@@ -197,7 +197,9 @@ export default class DirectionPanel {
     await this.panel.update();
     this.initDirection();
     UrlState.pushUrl();
-    this.searchDirection();
+    window.execOnMapLoaded(() => {
+      this.searchDirection()
```
GuillaumeGomez: I'm surprised that the linter is okay with his.

src/adapters/poi_popup.js
GuillaumeGomez: Badly understood the condition...

PR #247, opened 2019-06-13
QMAPS-851 - Back to list or close POI panel either it's PJ or OSM.

Sense:


src/views/poi_panel.dot
```diff
@@ -29,7 +29,7 @@
 <div class="poi_panel {{= this.active ? '' : 'poi_panel--hidden ' }}{{= this.card ? 'poi_panel--card ' : '' }}">
   {{? this.poi }}
     <div class="poi_panel__content__card">
-      <div class="poi_panel__close" {{= click(this.closeAction, this) }}>
+      <div class="poi_panel__close" {{= this.card && this.poi.isFromPagesjaunes() ? click(this.backToList, this) : click(this.closeAction, this) }}>
```
amatissart: The data source should not be used here : the list may contain POIs of all sources. Use `this.isFromList` instead.  

Also, this will impact both desktop and mobile. Don't we want to keep a button to simply close the panel ?

PR #256, opened 2019-06-18
Masq design changes

sujeebant:
- add link to Masq in burger menu
- add cursor for buttons inside onboarding
- move username and avatar to the rectangle where activate is displayed before logging in
- display avatar instead of burger when logged in
- use default profile image with first letter of username


src/scss/includes/modals/masq_onboarding.scss
```diff
@@ -105,19 +88,27 @@
...
 }
 
-.modal__masq_onboarding__steps_image_1 {
+.modal__masq_onboarding__image_1 {
+  background-image: url('../images/masq/gifs/OnBoarding01_732x420.gif');
+}
+
+.modal__masq_onboarding__image_1 {
+  background-image: url('../images/masq/gifs/OnBoarding01_732x420.gif');
+}
```
amatissart: duplicated section

sujeebant: :+1:

PR #257, opened 2019-06-19
QMAPS-870 - update search input query according to the url shard

Sense:
Fixing search input query not updating when accessing map through shared url (e.g `/maps/places/?q=Decathlon#map=12.40/43.7098068/7.2545280`)

### DIFF
--- a/src/panel/category_panel.js
+++ b/src/panel/category_panel.js
@@ -145,7 +145,6 @@ export default class CategoryPanel {
     SearchInput.unMinify();
     document.querySelector('.top_bar').classList.remove('top_bar--category-open');
     this.active = false;
-    this.query = '';
     this.panel.update();
     UrlState.pushUrl();
     if (toggleMarkers) {
--- a/src/ui_components/search_input.js
+++ b/src/ui_components/search_input.js
@@ -84,7 +84,7 @@ export default class SearchInput {
 
   async restore(fragment) {
     let shards = UrlShards.parseUrl();
-    if (shards.length === 1) {
+    if (shards.length > 0) {
       return await this.suggest.preselect(fragment);
     }
   }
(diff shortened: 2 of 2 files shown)


amatissart:
I think this change will interfere unexpectedly with the search_input.
When loading `/maps/places/?q=<query>`, the query should only be used by the category_panel. 

Wouldn't be sufficient to keep the query in the category_panel when selecting a place : 
```diff
diff --git a/src/panel/category_panel.js b/src/panel/category_panel.js
index 2659e6f..c210dc1 100644
--- a/src/panel/category_panel.js
+++ b/src/panel/category_panel.js
@@ -145,7 +145,6 @@ export default class CategoryPanel {
     SearchInput.unMinify();
     document.querySelector('.top_bar').classList.remove('top_bar--category-open');
     this.active = false;
-    this.query = '';
     this.panel.update();
     UrlState.pushUrl();
     if (toggleMarkers) {
```

Sense:
> I think this change will interfere unexpectedly with the search_input.
> When loading `/maps/places/?q=<query>`, the query should only be used by the category_panel.
> 
> Wouldn't be sufficient to keep the query in the category_panel when selecting a place :
> 
> ```diff
> diff --git a/src/panel/category_panel.js b/src/panel/category_panel.js
> index 2659e6f..c210dc1 100644
> --- a/src/panel/category_panel.js
> +++ b/src/panel/category_panel.js
> @@ -145,7 +145,6 @@ export default class CategoryPanel {
>      SearchInput.unMinify();
>      document.querySelector('.top_bar').classList.remove('top_bar--category-open');
>      this.active = false;
> -    this.query = '';
>      this.panel.update();
>      UrlState.pushUrl();
>      if (toggleMarkers) {
> ```

This can does the trick better than I expected.
Thanks.

PR #259, opened 2019-06-19
Fix menu search not getting correctly triggered

GuillaumeGomez:
When clicking on "Rechercher" in the hamburger menu, if there is a direction panel open (for example), nothing will happen. This fixes it.

### DIFF
--- a/src/panel/menu.js
+++ b/src/panel/menu.js
@@ -60,6 +60,7 @@ export default class Menu {
   }
 
   search() {
+    PanelManager.resetLayout();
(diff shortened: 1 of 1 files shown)


src/panel/menu.js
```diff
@@ -60,6 +60,7 @@ export default class Menu {
   }
 
   search() {
+    window.app.closeAllPanels();
```
amatissart: Couldn't we reuse `PanelManager.resetLayout()` here ?

PR #262, opened 2019-06-20
CSS fixes for Safari

amatissart:
* Disable default style on all input fields (including directions origin and destination)
* Fix icons position in burger menu

### DIFF
--- a/src/scss/includes/menu.scss
+++ b/src/scss/includes/menu.scss
@@ -131,7 +131,8 @@
   }
 
   .menu__panel__action__icon {
-    margin: -5px 20px 0 0;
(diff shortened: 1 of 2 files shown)


src/scss/includes/menu.scss
```diff
@@ -131,7 +131,8 @@
   }
 
   .menu__panel__action__icon {
-    margin: -5px 20px 0 0;
+    margin: -4px 20px 0 0;
+    padding: 4px 0px;
```
GuillaumeGomez: `0`, not `0px`.

amatissart: done

PR #276, opened 2019-06-24
QMAPS-864 CSS fixes on PJ POI panel

xem:


src/scss/includes/panels/poi_panel.scss
```diff
@@ -337,25 +337,14 @@ $HEADER_SIZE: 30px;
...
-  }
+  min-width: 24%;
+  max-width: 24%;
+  width: 24%;
 
   &:hover {
     text-decoration: none;
   }
-
+l
```
GuillaumeGomez: Type. :)

src/scss/includes/panels/poi_panel.scss
```diff
@@ -630,6 +620,17 @@ $HEADER_SIZE: 30px;
   }
 }
 
+@media (max-width: 490px) {
+  .poi_panel__container {
+    padding: 5px 15px 0;
+  }
+
+  .poi_panel__actions {
+    width: 110%;
```
GuillaumeGomez: That doesn't sounds like a good idea at all.

src/scss/includes/panels/poi_panel.scss
```diff
@@ -630,6 +620,17 @@ $HEADER_SIZE: 30px;
...
+    margin-left: -5%;
```
GuillaumeGomez: Same.

src/scss/includes/panels/poi_panel.scss
```diff
@@ -616,10 +606,10 @@ $HEADER_SIZE: 30px;
   }
 
   .poi_panel__header {
-    font-size: 13px;
+    font-size: 12px;
     color: $secondary_text;
     text-transform: uppercase;
-    height: $HEADER_SIZE;
+    height: 40px;
```
GuillaumeGomez: The constant is here for a reason. :-/

src/scss/includes/panels/poi_panel.scss
```diff
@@ -337,25 +337,14 @@ $HEADER_SIZE: 30px;
...
-  }
-
-  &.poi_panel__actions__phone_container {
-    max-width: 28%;
-  }
-
-  &:not(:first-child){
-    min-width: 20%
-  }
+  min-width: 24%;
```
GuillaumeGomez: I'm a bit skeptical about those changes: why setting this value three times? Also: it'll need to be checked very carefully.

PR #278, opened 2019-06-25
QMAPS-868: show PJ logo on POI panel after refresh

xem:


GuillaumeGomez:
On mobile when reloading from the poi:

Screenshot from 2019-06-25 10-47-55

Also, it doesn't show up when you click on "show more", only when reloading.

PR #284, opened 2019-06-26
QMAPS-865 fix PJ phone numbers reveals

xem:


GuillaumeGomez:
Why did the package-lock changed?

src/panel/poi_panel.js
```diff
@@ -216,6 +216,11 @@ PoiPanel.prototype.emptyClickOnMap = function() {
   }
 };
 
+PoiPanel.prototype.showPhone = function() {
+  document.querySelector(".poi_phone_container_hidden").style.display = "none"
```
GuillaumeGomez: Why not using the template system?

src/views/poi_panel.dot
```diff
@@ -81,11 +81,12 @@
...
-          <a class="poi_panel__actions__phone_container button_container"
-            data-rel="external" rel="noopener noreferrer"
-            href="{{= this.poi.blocksByType.phone.url }}">
-            <button class="poi_panel__actions__icon icon-icon_phone">
-            </button>
+          <a class="poi_panel__actions__phone_container button_container poi_phone_container_hidden" {{= click(this.showPhone, this) }}>
+            <button class="poi_panel__actions__icon icon-icon_phone"></button>
+            <div class="poi_panel__actions__text">{{= _('SEE NUMBER', 'poi') }}</div>
+          </a>
+          <a style="display:none" class="poi_panel__actions__phone_container button_container poi_phone_container_revealed" data-rel="external" rel="noopener noreferrer" href="{{= this.poi.blocksByType.phone.url }}">
```
GuillaumeGomez: Why not using the template system?

src/panel/poi_panel.js
xem: You mean re-render all the template, just to show the number? It seemed overkill.
Also, in the category list, the phone number is revealed in the same way (CSS display none => block, and it seemed fine so I did the same)

src/views/poi_panel.dot
xem: (same answer)

GuillaumeGomez:
> No idea, I had to do npm install before creating this branch and it got pushed with my code, but I'm not able to revert it ... is it bad?

Yep, quite bad considering none of those changes have anything to do with your PR.

src/panel/poi_panel.js
GuillaumeGomez: No, just part of the template which is much more maintainable than having to handle the DOM ourselves.

GuillaumeGomez:
You still have the lock file changes to remove.

PR #288, opened 2019-07-08
QMAPS-882 - Resizable panels (categories/favorites)

mlnor27:


src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+    window.deployedListHold = this.holdResizer;
+    window.deployedListStop = this.stopResize;
+  }
+
+  /**
+   * Looks for a resizable element
+   */
+  checkPanel() {
+    if (!this.panel) {
+      const element = document.querySelector(`.${resizableClassName}`);
```
GuillaumeGomez: What happens if we have multiple resizable elements? Wouldn't it be better to set an id instead?

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * Triggered on mouse move inside the body element while holding the panel resizer
+   * @param {MouseEvent|TouchEvent} e event
+   * @param {boolean} force Apply the height outside a panel resize
+   */
+  move(e, force = false) {
+    e.preventDefault();
+
+    const resizer = document.querySelector(`.${resizerClassName}`);
```
GuillaumeGomez: Same remark.

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+
+    if (!resizer) {
+      return;
+    }
+
+    if ((this.holding || force) && this.checkPanel() && !this.isTransitioning) {
+      const clientY = e.changedTouches ? e.touches[0].clientY : e.clientY;
```
GuillaumeGomez: You should check if `e.touches` exists before trying to access one of its values.

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+      return;
+    }
+
+    const clientY = e.changedTouches ? e.changedTouches[0].clientY : e.clientY;
+    const index = getClosest(Object.keys(sizes), clientY);
+    const adequateHeight = sizes[index];
+    let height = adequateHeight;
+
+    if (adequateHeight === 50) {
+
```
GuillaumeGomez: Why this empty line?

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+      if (this.holding) {
+        this.panel.classList.add(REDUCED);
+        fire('panel_view_state', { reduced: true });
+      } else {
+        height = window.innerHeight / 2;
+      }
+
```
GuillaumeGomez: Why this empty line?

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+    } else if (!this.holding && adequateHeight === window.innerHeight / 2) {
+
```
GuillaumeGomez: Why this empty line?

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+      this.panel.classList.add(REDUCED);
+      fire('panel_view_state', { reduced: true });
+      height = 50;
+
```
GuillaumeGomez: Why this empty line?

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+    } else if (adequateHeight === window.innerHeight) {
+
```
GuillaumeGomez: Why this empty line?

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
+const sizes = {
```
GuillaumeGomez: Please capitalize this global.

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+  async stopResize(e) {
+    e.preventDefault();
+
+    clearTimeout(this.timer);
+
+    if (!this.checkPanel()) {
+      return;
+    }
+
+    const clientY = e.changedTouches ? e.changedTouches[0].clientY : e.clientY;
```
GuillaumeGomez: Please check if `e.changedTouches` exists before indexing on it.

src/libs/resizable_panel.js
```diff
@@ -0,0 +1,170 @@
...
+      height = 50;
+
+    } else if (adequateHeight === window.innerHeight) {
+
+      if (!this.holding) {
+        this.panel.style.height = null;
+      }
+      this.panel.classList.add(FULL);
+      fire('panel_view_state', { maximized: true });
+    }
```
GuillaumeGomez: What happens in case `adequateHeight === undefined`?

src/libs/resizable_panel.js
mlnor27: deleted 👍

src/libs/resizable_panel.js
mlnor27: deleted 👍

src/libs/resizable_panel.js
mlnor27: deleted 👍

src/libs/resizable_panel.js
mlnor27: deleted 👍

src/libs/resizable_panel.js
mlnor27: 👍

src/libs/resizable_panel.js
mlnor27: deleted 👍

src/libs/resizable_panel.js
mlnor27: It can't be. We either display the favorite panel or the category panel, never both. This feature was intended solely for the main panels and not for other elements, so i don't see the need to handle this use case.
I can see the benefit of using an id. I'll do the change 👍

src/libs/resizable_panel.js
mlnor27: the use of an id is fair as we display one resizer button per main panel. I'll do the change

src/libs/resizable_panel.js
mlnor27: this line already checks if `e.changedTouches` exists with the ternary. Did you mean i should check if `e.changedTouches` isn't empty as i'm accessing `[0]` ?

src/libs/resizable_panel.js
mlnor27: It isn't possible due to how its value is calculated.
We have 3 possible states for the panel (reduced at the bottom of the page, at the middle, and fully deployed on the screen). And based on where the user released the panel (`clientY`), `adequateHeight` takes the closest value out of the 3 states. `undefined` can't be assigned in this context 👍

GuillaumeGomez: Then please add a `else` branch where you spawn an error and some log about an impossible situation. ;)

src/libs/panel_resizer.js
```diff
@@ -0,0 +1,155 @@
...
+
+  get handleElement() {
+    return document.querySelector(`[data-cid="${this.panel.cid}"] .${this.HANDLE_CLASS}`);
+  }
+
+  /**
+   * Triggered on mouse down of the panel resizer
+   * @param {MouseEvent|TouchEvent} e event
+   */
+  async holdResizer({event}) {
```
bbecquet: It doesn't seem to be an `async` method.

amatissart: `async` is now removed

src/libs/panel_resizer.js
```diff
@@ -0,0 +1,155 @@
...
+
+    const clientY = event.changedTouches ? event.changedTouches[0].clientY : event.clientY;
+
+    const SIZES = {
+      0: window.innerHeight,
+      [window.innerHeight / 2]: window.innerHeight / 2,
+      [window.innerHeight]: 50,
+    };
+    const index = getClosest(Object.keys(SIZES), clientY);
+    const adequateHeight = SIZES[index];
```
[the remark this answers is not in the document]
amatissart: an array is now defined, and `getClosest` becomes `getClosestIndex` to find the closest state among reduced, default and maximized.

PR #289, opened 2019-07-10
Qmaps 844 + QMAPS-845 add points anywhere on the map (+ panel with favorite / share / directions / categories buttons)

xem:
Capture d’écran de 2019-08-12 14-58-05


bin/middlewares/og_meta.js
```diff
@@ -1,94 +1,92 @@
-const request = require('request')
+const request = require('request');
```
GuillaumeGomez: Why 3 blank lines btw?

xem: I just added the ";", the blank lines were already there... but fixed!

bin/middlewares/og_meta.js
```diff
@@ -80,8 +77,11 @@ module.exports = function(config) {
   return function(req, res, next) {
     const placeUrlMatch = req.originalUrl.match(/place\/(.*)/);
     const locale = res.locals.language;
+    let poiId;
     if (placeUrlMatch && placeUrlMatch.length > 0) {
-      const poiId = placeUrlMatch[1];
+      poiId = placeUrlMatch[1];
+    }
+    if (poiId && !poiId.startsWith('latlon:')) {
```
amatissart: this change is no longer necessary, now that Idunn API can handle `latlon:` places

src/views/poi_panel.dot
```diff
@@ -110,9 +116,32 @@
...
+                </button>
+            {{?}}
+          {{~}}
+          </div>
+        {{?}}
+
       {{= this.PoiBlocContainer.render(this.poi) }}
 
-      {{? this.poi.isFromOSM() }}
+      {{? isAnywherePoi || (this.poi.isFromOSM && this.poi.isFromOSM()) }}
```
amatissart: this block should not be visible with anywhere POIs: OSM view and edit URLs are not defined for them

src/views/poi_partial/header.dot
```diff
@@ -1,6 +1,6 @@
 {{
   const ico = IconManager.get(this.poi);
-  const grades = this.poi.blocksByType.grades;
+  const grades = this.poi.blocksByType ? this.poi.blocksByType.grades : null;
```
amatissart: this check should not be necessary as `blocksByType` is now defined in base `Poi` class

bin/middlewares/og_meta.js
bbecquet: Fixed https://github.com/QwantResearch/erdapfel/pull/289/commits/6d7bb0eb8694653143ca059f006ec7d8aa5c62dd

src/views/poi_panel.dot
bbecquet: Fixed https://github.com/QwantResearch/erdapfel/pull/289/commits/a00d7b79c4c72c7cdc06a892332609492fb6a030

src/scss/includes/panels/poi_panel.scss
```diff
@@ -25,8 +25,8 @@ $HEADER_SIZE: 40px;
 }
 
 .poi_panel__image {
-  width: 100px;
-  height: 100px;
+  width: 90px;
+  height: 90px;
```
amatissart: do we really have to reduce the image size, that is already quite small ?

bbecquet: :man_shrugging: I don't even see the cases where it's used, so I don't have any opinion. Maybe we can simply revert this change for now to merge, and discuss every UI tweaks in a following PR.

bbecquet: Thanks for the IRL explanation, so this is the illustration picture (most often photo) that exists for some POIs. I agree this is important, so I reverted it at 100px for now. https://github.com/QwantResearch/erdapfel/pull/289/commits/90dc60f06d4fb05814d8fdec7afe623130384145
I'll probably change the layout of this component later so it doesn't depend on fixed height, so we don't have a big empty space when this photo doesn't exist.

PR #296, opened 2019-07-16
Add ESLint rule to use "const" instead of "let" where possible

bbecquet:
Not sure it's a priority, but here is a linting proposition related to [this comment](https://github.com/QwantResearch/erdapfel/pull/295#discussion_r303542215)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -73,6 +73,7 @@
         "no-unused-vars": [
             "error",
             { "argsIgnorePattern": "^_" }
-        ]
+        ],
+        "prefer-const": ["error"]
     }
 }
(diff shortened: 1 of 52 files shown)


GuillaumeGomez:
There is an error on CI.

bbecquet:
@GuillaumeGomez should be ok now, something I had to change manually. Two variables declared on the same line, one const and one let. Maybe a case for https://eslint.org/docs/rules/one-var later :D

GuillaumeGomez:
Hum. I think it needs a debate for the multiple variable declaration (I don't like it personally). But the current code is good so let's merge!

PR #298, opened 2019-07-16
Don't rotate geoloc icon background when active

bbecquet:
Fix proposal for the rotating black square background when hovering active geoloc icon.

Cause: the black bg is specific to Qwant, but the rotation effect comes from Mapbox GL CSS, which applies it to the whole element. We override it to apply it to the icon only.

| Before | After |
|---|---|
|Capture d’écran du 2019-07-16 15-22-28|Capture d’écran du 2019-07-16 13-58-37|


bbecquet:
Apparently it's been fixed upstream by @GuillaumeGomez

PR #303, opened 2019-07-19
Add ESLint rule to enforce '===' for equality tests

bbecquet:


src/panel/direction/direction_panel.js
```diff
@@ -50,7 +50,7 @@ export default class DirectionPanel {
   }
 
   setInputValue(type = 'origin', value) {
-    type == 'origin' ? this.searchInputStart.value = value : this.searchInputEnd.value = value;
+    type === 'origin' ? this.searchInputStart.value = value : this.searchInputEnd.value = value;
```
GuillaumeGomez: Could you actually rewrite this using an `if else` block structure please?

PR #307, opened 2019-07-23
Declare resolve aliases for 'config' and 'src' directories

bbecquet:
## Description
First example of how webpack resolve aliases can simplify imports.
If you like it, I'll do the same with `src` imports.

## Why
- Shorter and simpler to read imports
- No more hard coupling between files based on their relative place in the tree


GuillaumeGomez:
Looks promising!

PR #308, opened 2019-07-23
[Urls] Shards system refacto: don't expose low-level impl

bbecquet:
## Description
First PR to clean and simplify the url "shards" system.
Here we remove low-level calls to the shard implementation by components. Now the only needed import is `UrlState`, and everything passes through it only.

## Why
Debug the browser history management.

### DIFF
--- a/src/proxies/url_state.js
+++ b/src/proxies/url_state.js
@@ -21,7 +21,7 @@ UrlState.registerResource = function(component, prefix) {
 
 UrlState.registerUrlShard = function(component, prefix, paramType) {
   if (!component.store || !component.restore) {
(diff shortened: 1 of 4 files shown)


src/proxies/url_state.js
```diff
@@ -57,4 +57,13 @@ UrlState.load = function() {
   });
 };
 
+UrlState.getShardValue = function(shardPrefix) {
```
GuillaumeGomez: This is weird: from the name of this function, we'd expect it to return a value or it returns a boolean. Is it the expected behavior? If so, please rename the function.

GuillaumeGomez: Indeed, thanks for info. TIL.

PR #310, opened 2019-07-24
Remove custom Ajax file loader

bbecquet:
Now that the map bundle async loading is done through webpack, the custom file loader is used only to load the language file, and we don't need two implementations. Let's simplify!


dev/performances.js
```diff
@@ -106,8 +106,7 @@ async function serverStart() {
 
   configBuilder.set('mapStyle:poiMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`)
   configBuilder.set('mapStyle:baseMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`)
-  configBuilder.set('system:evalFiles', false)
-  configBuilder.set('performance:enabled', true)
+    configBuilder.set('performance:enabled', true)
```
amatissart: extra spaces

src/libs/async_file_loader.js
```diff
@@ -1,14 +1,10 @@
...
-    fileLoader.onLoad = () => {
-      resolve();
-    };
-    if (options.onProgress) {
-      fileLoader.onProgress = this.onProgress;
-    }
+    const sc = document.createElement('script');
+    sc.src = uri;
+    document.body.appendChild(sc);
+    sc.onload = resolve;
```
amatissart: Wouldn't it be safer to set `onload` before calling `appendChild` ?

dev/performances.js
bbecquet: Thanks. Fixed!

src/libs/async_file_loader.js
bbecquet: Indeed. I changed it.

PR #311, opened 2019-07-24
Define stricter match for categories labels in suggestions

amatissart:
A category is returned in autocomplete suggestions if the user query matches the regex associated to the category, or its translated label.  

This PR suggests to make this second check less permissive by requiring the query to start with the category label as a separate word (if there are enough letters in the query string).


src/adapters/category.js
```diff
@@ -37,7 +37,12 @@ export default class Category {
     const prefixLength = Math.min(4, this.label.length);
     // Match label prefix (eg: "supe" for "Supermarché", but not "supel")
     if (!matched && term.length >= prefixLength) {
-      matched = ExtentedString.compareIgnoreCase(term, this.label.slice(0, term.length)) >= 0;
+      if (this.label.length < term.length) {
+        // check if first word in term is label
+        matched = ExtentedString.compareIgnoreCase(term, `${this.label} `) === 0;
```
bbecquet: Not related to the feature, but there is a typo in the class name, should be `ExtendedString`.

src/adapters/category.js
```diff
@@ -37,7 +37,12 @@ export default class Category {
...
+      } else {
+        matched = ExtentedString.compareIgnoreCase(term, this.label.slice(0, term.length)) === 0;
```
bbecquet: Just a matter of taste maybe, but I would find it clearer to use `substring` instead of `slice`.

tests/integration/tests/autocomplete.js
```diff
@@ -306,8 +305,27 @@ test('Retrieve restaurant category when we search "restau"', async() => {
...
+  await page.keyboard.type(searchQuery);
+  await page.waitForSelector('.autocomplete_suggestion');
+
+  const firstLine = await page.evaluate(() => {
+    return document.querySelector(
+      '.autocomplete_suggestion:first-child .autocomplete_suggestion__first_line'
+    ).innerText;
+  });
+
+  expect(firstLine).toEqual('Test Result 1');
```
bbecquet: Strange that it's equal, as the mock data isn't capitalized :thinking:

src/adapters/category.js
amatissart: fixed

src/adapters/category.js
amatissart: done

tests/integration/tests/autocomplete.js
amatissart: Good point. A css rule had been defined to capitalize suggestions results, but this should be applied on category suggestions only. It's now fixed.

PR #313, opened 2019-07-25
Add custom HTTP headers to give picked suggestion ranking

GuillaumeGomez:


src/adapters/poi/idunn_poi.js
```diff
@@ -81,28 +81,35 @@ export default class IdunnPoi extends Poi {
...
     let requestParams = {};
     if (options.simple) {
       requestParams = {verbosity: 'short'};
     }
     try {
-      rawPoi = await Ajax.getLang(url, requestParams);
+      let headers = {};
+      if (obj.suggestionRank !== undefined && obj.query !== undefined) {
+        if (obj.position !== undefined) {
+          const pos = obj.position;
```
bbecquet: Why do you need to copy it? If it's just to have a shorter line below, a better solution would be to deconstruct the object:
```js
const { lat, lon, zoom } = obj.position;
```

src/adapters/poi/idunn_poi.js
```diff
@@ -81,28 +81,35 @@ export default class IdunnPoi extends Poi {
...
     let rawPoi = null;
-    const url = `${serviceConfig.idunn.url}/v1/places/${id}`;
+    const url = `${serviceConfig.idunn.url}/v1/places/${obj.id}`;
     let requestParams = {};
     if (options.simple) {
       requestParams = {verbosity: 'short'};
     }
     try {
-      rawPoi = await Ajax.getLang(url, requestParams);
+      let headers = {};
```
bbecquet: You should have a linter error because it should be a `const`.

src/adapters/poi/idunn_poi.js
```diff
@@ -81,28 +81,35 @@ export default class IdunnPoi extends Poi {
...
+      if (err !== 404) {
         const s_requestParams = JSON.stringify(requestParams);
         Error.sendOnce(
           'idunn_poi', 'poiApiLoad',
           `unknown error getting idunn poi reaching ${url} with options ${s_requestParams}`,
           err
         );
-        return;
       }
+      return;
```
bbecquet: This `return` is useless here.
Another way would be to do a real early return by keeping the `err === 404` test but ditching the `else`.
IMO it's a better pattern to manage the easy cases first.

src/adapters/suggest.js
```diff
@@ -113,7 +113,13 @@ export default class Suggest {
         if (prefixPoint !== undefined) {
           this.onSelect(prefixPoint);
         } else {
-          const selectedItem = items.find(item => item.id === itemId);
+          const selectedItemIndex = items.findIndex(item => item.id === itemId);
+          let selectedItem = items[selectedItemIndex];
```
bbecquet: Another `const` :)
BTW, it may be safer to copy the items props instead of modifying the object on place. No risk of side effects.

src/libs/ajax.js
```diff
@@ -5,21 +5,37 @@ const systemConfigs = nconf.get().system;
...
 
-Ajax.getLang = async(url, data = {}, options) => {
+Ajax.getLang = async(url, data = {}, options = {}, headers = {}) => {
   data.lang = window.getLang().code;
-  return Ajax.get(url, data, options);
+  return Ajax.get(url, data, options, headers);
 };
 
+function setHeaders(xhr, headers) {
+  if (typeof headers !== 'object') {
```
bbecquet: Not sure this test is useful.

src/proxies/panel_manager.js
```diff
@@ -22,9 +22,9 @@ PanelManager.setPoi = async function(poi, options = {}) {
   window.app.unminify();
 };
 
-PanelManager.loadPoiById = async function(id, options) {
-  if (id) {
-    const poi = await ApiPoi.poiApiLoad(id);
+PanelManager.loadPoiById = async function(obj, options) {
```
bbecquet: Something gets wrong or unclear with the function name now that you pass full objects…

src/adapters/poi/idunn_poi.js
GuillaumeGomez: :+1:

src/libs/ajax.js
GuillaumeGomez: It allows to remove `undefined`.

src/adapters/poi/idunn_poi.js
```diff
@@ -86,28 +86,35 @@ export default class IdunnPoi extends Poi {
...
+    const url = `${serviceConfig.idunn.url}/v1/places/${obj.id}`;
     let requestParams = {};
     if (options.simple) {
       requestParams = {verbosity: 'short'};
     }
     try {
-      rawPoi = await Ajax.getLang(url, requestParams);
+      const headers = {};
+      if (obj.suggestionRank !== undefined && obj.query !== undefined) {
+        if (obj.position !== undefined) {
```
amatissart: actually, we will need more context, such as the language used in the query.  

It might be cleaner to define a `queryContext` object that could be passed in the `options` object, instead of this cumbersome validation ?

src/adapters/scene.js
```diff
@@ -113,7 +113,11 @@ Scene.prototype.initMapBox = function() {
               offset: [(layout.sizes.panelWidth + layout.sizes.sideBarWidth) / 2, 0],
             });
           }
-          window.app.loadPoi(mapPoi, { disableMapPan: true });
+
+          const poi = await window.app.loadPoi(mapPoi, { disableMapPan: true });
+          if (poi) {
+            this.addMarker(poi);
+          }
```
amatissart: this change is probably obsolete and should have been absent after rebase, no ?

src/libs/ajax.js
```diff
@@ -58,10 +74,14 @@ const query = (url, data, method = 'GET', options = {}) => {
...
     if (method === 'GET') {
       xhr.open(method, `${url}?${dataToUrl(data)}`);
+      setHeaders(xhr, headers);
       xhr.send();
     } else {
       xhr.open(method, url);
-      xhr.setRequestHeader('Content-Type', 'application/json;charset=UTF-8');
+      setHeaders(xhr, headers);
+      if (!headersContain(headers, 'Content-Type')) {
+        xhr.setRequestHeader('Content-Type', 'application/json;charset=UTF-8');
```
amatissart: shouldn't this header be set on POST requests only ?

src/adapters/scene.js
GuillaumeGomez: Absolutely no idea.

src/libs/ajax.js
GuillaumeGomez: I didn't change the method check. Do you want me to check for `POST` only?

amatissart: Yes. Before this change we used only `GET` and `POST`, and this header should be defined only when JSON data is present in body.

src/adapters/scene.js
bbecquet: @amatissart is right, it's done through events in the `loadPoi` method.

src/adapters/poi/idunn_poi.js
GuillaumeGomez: Indeed.

src/adapters/poi/idunn_poi.js
```diff
@@ -86,28 +86,30 @@ export default class IdunnPoi extends Poi {
...
+    const url = `${serviceConfig.idunn.url}/v1/places/${obj.id}`;
     let requestParams = {};
     if (options.simple) {
       requestParams = {verbosity: 'short'};
     }
     try {
-      rawPoi = await Ajax.getLang(url, requestParams);
+      const headers = {};
+      if (obj.queryContext) {
+        obj.queryContext.fillHeaders(headers);
```
amatissart: This feature should be defined behind a feature flag, and enabled by default.  
Just in case it should be disabled because of some CORS issues.

src/adapters/query_context.js
```diff
@@ -0,0 +1,26 @@
...
+    // `position` field is supposed to contain `lat`, `lon` and `zoom`.
+    this.position = position;
+  }
+
+  fillHeaders(headers) {
+    if (this.position.lat !== undefined &&
+        this.position.lon !== undefined &&
+        this.position.zoom !== undefined) {
+      const { lat, lon, zoom } = this.position;
+      headers['X-QwantMaps-FocusPosition'] = `${lat};${lon};${zoom}`;
```
amatissart: `y;x;z` is a weird convention,
why not `x;y;z` ?
or `z;y;x` to be consistent with the `#map` hash value ?

And the values should be formatted with a fixed number of decimals.

src/libs/ajax.js
```diff
@@ -58,10 +74,15 @@ const query = (url, data, method = 'GET', options = {}) => {
...
     if (method === 'GET') {
       xhr.open(method, `${url}?${dataToUrl(data)}`);
+      setHeaders(xhr, headers);
       xhr.send();
     } else {
       xhr.open(method, url);
-      xhr.setRequestHeader('Content-Type', 'application/json;charset=UTF-8');
+      setHeaders(xhr, headers);
+      if (method === 'POST' && !headersContain(headers, 'Content-Type') &&
+          data && data.length > 0) {
```
amatissart: `data` is an object (that will be stringified below), so using `data.length` is incorrect here.

src/adapters/query_context.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+    headers['X-QwantMaps-Query'] = encodeURI(this.term);
```
amatissart: `encodeURIComponent` would be more appropriate

src/adapters/poi/idunn_poi.js
GuillaumeGomez: Yes but I won't make this check here but where we create the headers instead.

GuillaumeGomez: Actually, I'll make it into the `fillHeaders` method instead. Like that no need to make the check every time we create a new `QueryContext`.

config/default_config.yml
```diff
@@ -72,3 +72,6 @@ direction:
 category:
   enabled: false
   maxPlaces: 40
+
+cors:
+  enabled: true
```
amatissart: This name does not describe the exact feature. It's actually not directly related to CORS.  
What about `telemetry.sendQueryContextHeaders` ?

GuillaumeGomez:
Updated.

PR #315, opened 2019-07-25
Centralize POI state

bbecquet:
## Description
Still trying to simplify the app state management…
There is no reason to keep the main App and the panel manager separated (Panel manager does other stuff too, like fetching POIs), so let's merge them => first commit.
In the second commit, remove the separated `scene_state` which held a copy of the active POI for the map, but was used explicitely elsewhere.

May be split in 2 separate PRs if you find it safer to review.

## Why
Reducing the number of code paths which change the app state.
The end goal is to pass through single centralized functions to do global app actions, like setting the active POI, instead of having all parts calling themselves and sending events in every direction.


src/panel/app_panel.js
```diff
@@ -93,4 +97,81 @@ export default class AppPanel {
       this.minify();
     }
   }
+
+  async setPoi(poi, options = {}) {
+    this.activePoiId = poi.id;
+    this.panels.forEach(panel => {
+      if (panel.isPoiCompliant) {
+        panel.setPoi(poi, options);
+      } else if (!options.isFromList && !options.isFromFavorite) {
```
amatissart: `isFromList` has been renamed :)
```suggestion
      } else if (!options.isFromCategory && !options.isFromFavorite) {
```

src/panel/app_panel.js
```diff
@@ -93,4 +97,81 @@ export default class AppPanel {
...
+
+  openCategory(options) {
+    this._openPanel(this.categoryPanel, options);
+  }
+
+  resetLayout() {
+    this._openPanel(this.servicePanel);
+  }
+
+  async keepOnlyPoi() {
```
amatissart: this function is unused now and can be removed

src/panel/app_panel.js
```diff
@@ -18,7 +18,7 @@ import DirectionPanel from './direction/direction_panel';
 import Menu from './menu';
 import Telemetry from '../libs/telemetry';
 import CategoryPanel from './category_panel';
-import PanelManager from 'src/proxies/panel_manager';
```
amatissart: so "panel_manager.js" file can be safely deleted, no ?

bbecquet: Ah yes, just forgot. Thanks!

src/panel/app_panel.js
bbecquet: Done.

src/panel/poi_panel.js
```diff
@@ -8,19 +8,17 @@ import SearchInput from '../ui_components/search_input';
...
 import Device from '../libs/device';
-import PanelManager from 'src/proxies/panel_manager';
 import poiSubClass from '../mapbox/poi_subclass';
 
 const store = new Store();
 const masqFavoriteModal = new MasqFavoriteModal();
 
 function PoiPanel(sharePanel) {
-  this.isPoiComplient = true; /* Poi Compliant */
+  this.isPoiCompliant = true;
```
GuillaumeGomez: :+1:

PR #316, opened 2019-07-26
Factorize set/unset POI

bbecquet:
## Description
Move some stuff related to setting/unsetting the active POI in the top level. Individual panels shouldn't have to micro-manage that individually.

## Why
Improving bits by bits the consistency of the state management across the app.
Prepare the road for other simplification and fixes

### DIFF
--- a/src/adapters/poi/map_poi.js
+++ b/src/adapters/poi/map_poi.js
@@ -1,8 +1,10 @@
+import { LngLat } from 'mapbox-gl--ENV';
 import Poi, {POI_TYPE} from './poi';
 
 export default class MapPoi extends Poi {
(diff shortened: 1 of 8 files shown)


bbecquet:
I put this PR on WIP until the tests are fixed.
And there is actually a UX regression, the map move events are thrown *after* the POI data is fetched from the API.

bbecquet:
Here we go, the tests now pass and there is no more UX/perf regression: the map still updates as soon as possible without waiting for the full POI info to be feched from the API.

The `setPoi`/`loadPoi` options are still a bit complicated. I'll try to simplify that once more in another PR.

PR #321, opened 2019-07-30
Change links to instead edit POI

GuillaumeGomez:
Screenshot from 2019-07-30 16-18-19


src/views/poi_panel.dot
```diff
@@ -119,9 +119,13 @@
           <p class="osm_contribute__title">
             {{= _('Qwant Maps uses OpenStreetMap') }}
```
amatissart:
```suggestion
            {{= _('Qwant Maps uses OpenStreetMap data.') }}
```

PR #322, opened 2019-07-31
Introduce a basic URL router system

bbecquet:
## Description
After trying to simplify the "URL shards" system, I'm introducing a real router as a better way to manage navigation between the app states and interaction with the browser history.

This is a really simple, ad-hoc implementation, very easy to remove or replace if we decide to use something more elaborate in the future (`src/proxies/app_router.js`).

In this new architecture, the root component (`src/panel/app_panel.js`) defines URL templates with optional parameters, corresponding to valid app states, and for each template a rendering function. The panels are now only switched on or off through this function. Re-renders are the result of new URL states and not the contrary.
Transition between valid states is made explicit through a single dedicated function (`navigateTo`), so it happens reliably at a single place.

## Why
- the "URL shards" system was far too complicated and required logic parts scattered everywhere in the code.
- proper browser history management was near impossible with the shards, as history pushes happened in the panel functions doing the actual renderings. Restoring past state through browser `back`/`forward` buttons resulted in new pushes. With the new system this is seamless.
- valid app states and corresponding URLs are now explicit and grouped in a single place, making them easier to manage.
- first step towards a single rendering flow system, which would be a better app architecture IMO.


src/adapters/scene.js
```diff
@@ -306,6 +289,30 @@ Scene.prototype.fitMap = function(item, padding) {
...
+    }
+  }
+  const offset = Device.isMobile()
+    ? [0, 0]
+    : [(layout.sizes.panelWidth + layout.sizes.sideBarWidth) / 2, 0];
+  this.mb.flyTo({
+    center: poi.getLngLat(),
+    zoom: poi.zoom,
+    offset,
+    maxDuration: 1200,
```
GuillaumeGomez: A magical value. :)

GuillaumeGomez:
Code looks good to me. Still need to check it locally though.

src/panel/app_panel.js
```diff
@@ -72,6 +75,74 @@ export default class AppPanel {
...
+
+    this.router.addRoute('Favorites', '/favs', () => {
+      this.openFavorite();
+    });
+
+    this.router.addRoute('Routes', '/routes(?:/?)(.*)', routeParams => {
+      this.openDirection(parseQueryString(routeParams));
+    });
+
+    this.router.addRoute('Direct search query', '/?q=(.*)', query => {
```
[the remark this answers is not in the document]
bbecquet: You're right, this regexp is too specific. I'll change the system to allow arbitrary query params and check the presence of `q`.

src/panel/app_panel.js
```diff
@@ -62,8 +67,6 @@ export default class AppPanel {
 
     this.menu = new Menu();
 
-    this.activePoiId = null;
```
amatissart: this is still used in `poi_popup.js` (it is supposed to avoid showing the popup on click)

bbecquet: Indeed, I broke this behavior. I'll repair it.

src/adapters/poi/poi.js
```diff
@@ -78,6 +79,18 @@ export default class Poi {
...
+
+  static deserialize(raw) {
+    const { id, name, alternativeName, type, latLon, className, subClassName, bbox } = raw;
+    return new Poi(id, name, alternativeName, type, latLon, className, subClassName, bbox);
+  }
+
+  serialize() {
+    // In some cases the object has an `event` prop which is a low-level browser object
+    // that can't be serialized in the history state => just ignore it
+    const { event: _event, ...otherFields } = this;
```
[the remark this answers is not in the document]
bbecquet: For the record, after looking for it: it happens when clicking on a poi displayed as a div in a list, when the onclick event is attached to it with the lib defined in https://github.com/QwantResearch/erdapfel/blob/master/src/libs/actions.js
I will try to simplify this system in another PR, later.

src/adapters/scene.js
```diff
@@ -363,14 +352,23 @@ Scene.prototype.isWindowedPoi = function(poi) {
...
+};
+
 Scene.prototype.onHashChange = function() {
   window.onhashchange = () => {
-    const mapShardValue = UrlState.getShardValue('map');
-    if (mapShardValue) {
-      this.restore(mapShardValue);
-      this.mb.jumpTo({center: this.urlCenter, zoom: this.urlZoom});
-    }
+    this.restoreFromHash(window.location.hash);
```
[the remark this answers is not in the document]
bbecquet: Thanks, I changed it.

src/panel/app_panel.js
bbecquet: This is fixed with this commit https://github.com/QwantResearch/erdapfel/pull/322/commits/471d935cab3985f41d6bb833a0f775a221848f6e

PR #323, opened 2019-07-31
ESLint: object litterals standardization

bbecquet:
Let's standardize two other things:
- inner spaces for braces in object literals (https://eslint.org/docs/rules/object-curly-spacing)
- use of shorthand syntax when prop name and var name are the same (https://eslint.org/docs/rules/object-shorthand).

### DIFF
--- a/bin/app.js
+++ b/bin/app.js
@@ -38,7 +38,7 @@ function App(config) {
 
   /* Define child logger in req */
   app.use((req, res, next) => {
-    req.logger = logger.child({req: req});
+    req.logger = logger.child({ req });
     next();
   });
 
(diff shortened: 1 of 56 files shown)


bin/app.js
```diff
@@ -38,7 +38,7 @@ function App(config) {
 
   /* Define child logger in req */
   app.use((req, res, next) => {
-    req.logger = logger.child({req: req});
+    req.logger = logger.child({ req: req });
```
GuillaumeGomez: Please make a rule for it. :)

GuillaumeGomez:
Waiting for the second eslint add before merging.

bin/app.js
bbecquet: Done.

PR #326, opened 2019-08-05
Capitalize first letter in favorites too

GuillaumeGomez:
Screenshot from 2019-08-05 17-09-24

The first letter was not capitalized before in the favorites menu. It is now.


bbecquet:
It looks like tests have to be adapted.

PR #333, opened 2019-08-09
POC React

bbecquet:
## Description
Just a quick PoC to see what it would take to migrate the non-map part of the app to React.

Starting with the easiest part, the ServicePanel. For now it requires a thin glue layer to mount the React component in a DOM node rendered by dot, but the final goal is to have all panels being rendered as a single React tree.

## Why
Because I don't think doT.js is a viable long-term solution to build the app UX on. It is too simple, too static, not maintained anymore, and not suited for the needs of a full scale dynamic app when states and interactions are complex.


bbecquet:
I'll split this PR in two separate ones so it's easier to focus on what's done.
Here is the first one, which just installs React and configure Babel/ESLint https://github.com/QwantResearch/erdapfel/pull/346

bbecquet:
Seconde part, the panel migration itself: https://github.com/QwantResearch/erdapfel/pull/348
We can now close this one.

PR #335, opened 2019-08-13
Simplify POI panel CSS

bbecquet:
## Description
Simplify some markup and CSS in the POI panel, especially start using `display: flexbox` instead of `float` to improve some layouts (tested successfully on IE11) and simplify the complicated action buttons.

## Why
Most layout DOM and CSS are far too complicated and use obsolete layout techniques. Ease future migration to another view framework.

### DIFF
--- a/src/scss/includes/panels/poi_panel.scss
+++ b/src/scss/includes/panels/poi_panel.scss
@@ -2,7 +2,6 @@
 
 $BLOCK_INDENT: 28px;
 $CARD_ACTION_WIDTH: 112px;
-$CARD_ACTION_RIGHT_MARGIN: 8px;
 $CARD_PADDING: 8px;
 $BLOCK_PADDING: 24px;
 $BLOCK_ICON_FONT_SIZE: 16px;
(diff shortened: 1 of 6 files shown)


src/scss/includes/panels/poi_panel.scss
```diff
@@ -313,75 +289,36 @@ $HEADER_SIZE: 40px;
 }
 
 .poi_panel__actions {
-  width: 100%;
-  position: relative;
-  margin-bottom: 10px;
-  overflow: auto;
+  display: flex;
```
amatissart: Without this `margin-bottom`, the "osm_contribute" div is packed below the buttons (when no other block is present)
image

Maybe add a margin-top on `.osm_contribute` ?

bbecquet: Thanks, I was a bit too quick to remove rules. I re-introduced a `margin-bottom`, as it's easier to manage only bottom margins when stacking components. It looks better indeed :)
Capture d’écran de 2019-08-14 17-52-58

PR #336, opened 2019-08-14
If custom headers are rejected, send request without them

GuillaumeGomez:


src/adapters/poi/idunn_poi.js
```diff
@@ -103,6 +103,10 @@ export default class IdunnPoi extends Poi {
     } catch (err) {
       if (err === 404) {
         return;
+      } else if (err === 0) { // When the OPTIONS request is rejected, the error is 0 and not 405
+        console.warn('Headers aren\'t allowed, sending query without them...');
```
GuillaumeGomez: Just wondering about this: is there a logger I should use or just this warn is fine?

src/adapters/poi/idunn_poi.js
```diff
@@ -103,6 +103,10 @@ export default class IdunnPoi extends Poi {
     } catch (err) {
       if (err === 404) {
         return;
+      } else if (err === 0) { // When the OPTIONS request is rejected, the error is 0 and not 405
```
bbecquet: `0` isn't a valid HTTP status code, is it a custom one from Idunn API? And is it guaranteed this will be used only for this error case? Otherwise it's a bit dangerous to resend the request in this `catch`, you may enter an infinite loop of requests…

src/adapters/poi/idunn_poi.js
bbecquet: `console.warn` is fine with me, especially because this case is supposed to happen only in dev envs.

src/adapters/poi/idunn_poi.js
bbecquet: Just as small syntax preference, but in cases of early returns like here I find it simpler to not use `else` constructs at all.

GuillaumeGomez: Oh good point, I need to add a check to confirm that we should resend it.

Also, not sure why 0 is returned (my guess is that it comes from the fact that it is the OPTIONS request which is failing) but since we're in the catch block, I assume it's fine?

GuillaumeGomez: Takes more place if not. :p

bbecquet: So?
Ok, I'll push another ESLint rule later for that :P

GuillaumeGomez: You make me sad sometimes haha.

PR #342, opened 2019-08-20
Start using browser-ui-test framework

GuillaumeGomez:
This is just a POC for the moment with not much done. It currently just checks that clicking on the "my favorites" button, it goes to the expected panel. It also doesn't use screenshot generation/comparison.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3444,13 +3444,13 @@
       }
     },
     "browser-ui-test": {
-      "version": "0.1.6",
-      "resolved": "https://registry.npmjs.org/browser-ui-test/-/browser-ui-test-0.1.6.tgz",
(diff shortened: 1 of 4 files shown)


bbecquet:
Before diving too deep into this, can you explain what the rationale is?
Is it to simplify test writing and maintenance, or to enable screenshot comparison capabilities, or smth else?
In the first two cases, I think a lot could be done with the current tools to improve the experience instead of introducing a very new tool.

GuillaumeGomez:
I had in mind the screenshot mainly. However I find the syntax much simpler but not sure it would allow to cover all what we currently have. To be debated.

bbecquet:
Let's focus on improving upon our current test tooling instead of switching to a very new stack.

PR #344, opened 2019-08-21
[PWA] Manifest.json and other easy requirements

bbecquet:
## Description
First step to make a PWA around the existing webapp. Fills in the easy declarative requirements, as listed by Chrome PWA devtools:
- have a `manifest.json` file with some required properties
- additional meta data and logos
- a mandatory `<noscript>` tag.

## Why
Allows the installation of the web site as a standalone app on the start screen (at least in Chrome Android), and opens it with a basic splashscreen, in fullscreen, and in locked portrait position.

This does not provide any other advanced PWA functionality like offline mode. This will be addressed in further developments.

|"Add to home screen"|Splashcreen|Fullscreen mode|
|---|---|---|
|Screenshot_20190821-103649_Chrome|Screenshot_20190821-102339_Chrome|Screenshot_20190821-102318_Chrome|


views/index.ejs
```diff
@@ -50,5 +53,6 @@
       <script>window.times = {init : Date.now()}</script>
     <% } %>
     <script src="./statics/build/javascript/bundle.js"></script>
+    <noscript>JavaScript is required to view this application.</noscript>
```
amatissart: should this message be translated ?

views/index.ejs
```diff
@@ -50,5 +53,6 @@
       <script>window.times = {init : Date.now()}</script>
     <% } %>
     <script src="./statics/build/javascript/bundle.js"></script>
+    <noscript><%= _('JavaScript is required to view this application.') %></noscript>
```
amatissart: this element is rendered behind the "top_bar" :crying_cat_face:

bbecquet: I fixed it in the last commit localhost_3000_(Galaxy S5) (1)

PR #346, opened 2019-08-21
Install React

bbecquet:
## Description
Install the minimal React setup for our toolset.
As React is not imported yet in any code, this doesn't change what's built.

## Why
Prepare things to start migrating stuff.


.eslintrc.json
```diff
@@ -3,7 +3,15 @@
         "browser": true,
         "es6": true
     },
-    "extends": "eslint:recommended",
+    "extends": [
+      "eslint:recommended",
+      "plugin:react/recommended"
```
bbecquet: This is to make ESLint understand React specifics. We can configure it further later (see [doc](https://github.com/yannickcr/eslint-plugin-react)).

.eslintrc.json
```diff
@@ -85,7 +93,8 @@
         "no-var": "error",
         "eqeqeq": "error",
         "object-curly-spacing": ["error", "always"],
-        "object-shorthand": "error"
+        "object-shorthand": "error",
+        "react/prop-types": 0
```
bbecquet: We just disable warning for props used without prop-types declaration, because we don't use it yet. https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

build/babel.config.js
```diff
@@ -3,6 +3,7 @@
 module.exports = function(mode) {
   const plugins = [
     '@babel/plugin-syntax-dynamic-import',
+    '@babel/plugin-proposal-class-properties',
```
bbecquet: This is not strictly related to React, but this allows to use simplified ES class syntax, which is handy for state, static properties and autobound methods. See [this article](https://michalzalecki.com/react-components-and-class-properties/) for examples.

PR #348, opened 2019-08-22
React: migrate service panel

bbecquet:
## Description
Second part of splitting the original [React POC](https://github.com/QwantResearch/erdapfel/pull/333) in two. Actually migrate the service panel.

## Why
Start the migration with an easy component, and explain things in the process (see this PR comments).


src/views/app_panel.dot
```diff
@@ -9,7 +9,7 @@
       {{= this.categoryPanel.render() }}
     {{?}}
   </div>
-  {{= this.servicePanel.render() }}
```
bbecquet: Instead of calling the rendering the panel even if it's hidden, just prepare a container in the DOM so that we can render a React component tree when needed.

tests/integration/tests/menu.js
```diff
@@ -62,21 +62,14 @@ test('menu open favorite', async () => {
 });
 
 test('one panel open at a time', async () => {
-  expect.assertions(2);
+  expect.assertions(1);
   await page.goto(APP_URL);
-  const servicePanelOpen = await page.waitForSelector('.service_panel--active');
```
bbecquet: The `service_panel` isn't "active" (`display:block`) or "inactive" (`display:none`) anymore. It either is in the DOM or not, so we have to adapt the tests.

src/panel/category_panel.js
```diff
@@ -87,8 +87,6 @@ export default class CategoryPanel {
 
     this.addCategoryMarkers();
     fire('save_location');
-
-    document.querySelector('.service_panel').classList.remove('service_panel--active');
```
bbecquet: When the service panel isn't shown, it isn't in the DOM at all, so we don't need to control it explicitely like this anymore => one less ugly dependency.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
-import ServicePanelView from '../views/service_panel.dot';
-import Panel from '../libs/panel';
-import nconf from '@qwant/nconf-getter';
+/* globals _ */
+import React from 'react';
```
bbecquet: Files that use the [JSX syntax](https://reactjs.org/docs/introducing-jsx.html) always need to import the `react` module, event it isn't explicitely called.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+import ReactDOM from 'react-dom';
```
bbecquet: For now we need `react-dom` too because we render this component in the DOM in this file.
Later, when the migration is finished, we'll have only one ReactDOM call at the app top level, every other component will be purely declarative.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
-  constructor() {
-    this.panel = new Panel(this, ServicePanelView);
-    this.categories = CategoryService.getCategories();
-    this.mustDeployCategories = this.categories.length > 8;
-    this.isDeployed = false;
-    this.isDirectionActive = nconf.get().direction.enabled;
-    this.active = true;
-  }
+// Returns a function, which when called will navigate to the url
+const navTo = url => () => {
```
bbecquet: This has nothing to do with React per-se, I just wanted clean calls to that without multiplying references to the `app` global value.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+  window.app.navigateTo(url);
+};
 
-  toggleCategories() {
-    this.isDeployed = !this.isDeployed;
-    this.panel.update();
-  }
+const CategoryButton = ({ category }) => {
```
bbecquet: There are two ways of defining React components (see [React doc](https://reactjs.org/docs/components-and-props.html)): 
- the simpler way is like here, as a function which takes properties (called *props* in React) as fields in its first argument, and returns HTML.
- another way is below, as an ES6 class. It must have at least a `render` function which returns HTML. The props are accessible as keys of the `this.props` object.

The main difference is the class components can have inner state and life cycle control methods, whereas the functional components are stateless, just rendering stuff. They can be compared to a dotT.js partial.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+  return <button
+    className="service_panel__category"
```
bbecquet: JSX is mostly XHTML, which some particularities. One is that some attribute names which are also JS reserved terms cannot be used. The most common one is `class` which is replaced by `className`.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+    type="button"
+    onClick={navTo(`/places/?type=${category.name}`) }
```
bbecquet: This is how you attach events in JSX. Here `navTo` returns a function itself, maybe it's a bit confusing then… Would be clearer like that:
```jsx
<div onClick={() => { console.log('plop'); }} />
```

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+  >
+    <div className="service_panel__category__icon" style={{
+      background: category.backgroundColor,
+    }}>
+      <span className={`icon icon-${category.iconName}`} />
```
bbecquet: In JSX, everything between `{ }` is interpreted.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+      {/* TODO: use real links instead of all these buttons */}
 
-  openFavorite() {
-    window.app.navigateTo('/favs');
+      <div className="service_panel__actions">
+        <button
+          className="service_panel__action service_panel__item__fav"
+          onClick={navTo('/favs') }
+        >
+          <i className="icon-icon_star" />{' '}
```
bbecquet: In JSX, empty tags can use the XML shortcut closing syntax `<foo />`. React knows which HTML elements it has to really render as opening/closing tags.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+          onClick={navTo('/routes') }
+        >
+          <i className="icon-corner-up-right" />{' '}
+          <span className="service_panel__action__text">
+            {_('DIRECTIONS', 'service panel')}
+          </span>
+        </button>
+      </div>
+
+      {categories.length > 0 && <div className="service_panel__categories">
```
bbecquet: This is the easiest way to do a conditional rendering. What follows `&&` won't be rendered if `categories.length > 0` is false.

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+        {categories.map(category => <CategoryButton key={category.label} category={category} />) }
```
bbecquet: This it how you call another component in JSX. Notice the CamelCase for `CategoryButton`.
Props are passed with the HTML attribute syntax.

`key` is a required prop to pass to children when rendering a list dynamically. You get a warning if you don't add it. This is linked to the React rendering model ([explanation](https://reactjs.org/docs/lists-and-keys.html#keys))

src/panel/service_panel.js
```diff
@@ -1,47 +1,81 @@
...
+// Won't be needed when all panels are migrated
+export default class _ServicePanel {
+  constructor(containerSelector) {
+    this.containerSelector = containerSelector;
   }
 
   open() {
-    this.active = true;
-    this.panel.update();
+    ReactDOM.render(
```
bbecquet: This renders the given component with the given props to the DOM. After that, React will manage the life cycle of components without having to do that again.

bbecquet:
I'm closing this for now because it changed a lot and to make room for the others React PR (please review :pray:).

PR #352, opened 2019-08-23
Handle direction button differently

GuillaumeGomez:
Instead of creating the direction button on its own outside of the mapbox owned DOM, we wait for mapbox to be loaded and then create it alongside other buttons.

### DIFF
--- a/src/mapbox/extended_nav_control.js
+++ b/src/mapbox/extended_nav_control.js
@@ -21,6 +21,12 @@ export default class ExtendedControl {
     this._compass = this._createButton(compassClass, 'Reset North', () => {
       this._resetNorthAndTilt();
     });
+    this._direction = this._createButton('direction_shortcut icon-corner-up-right', 'direction',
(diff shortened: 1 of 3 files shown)


bbecquet:
Can you explain simply in the PR description the changes you introduce or/and why? The code alone isn't so obvious.

src/mapbox/extended_nav_control.js
```diff
@@ -21,6 +21,16 @@ export default class ExtendedControl {
     this._compass = this._createButton(compassClass, 'Reset North', () => {
       this._resetNorthAndTilt();
     });
+    this._direction = this._createButton('direction_shortcut icon-corner-up-right', 'direction',
+      () => {
+        Array.prototype.slice.call(
+          document.getElementsByClassName('direction_shortcut')
```
bbecquet: I think you don't need an array.
`document.querySelector('direction_shortcut')` will return the one and only component.

PR #357, opened 2019-08-27
[Public transport] Display basic route results on the map

bbecquet:
*Probably easier to review commit by commit*

## Description
Displays public transport route results on the map, for now with the basic style, i.e. without any distinction of step type, metro line color, etc.
As public transport result geoms are GeoJSON FeatureCollection (and not unique geometries like car/walking/bike), we clean and refactor some stuff to be able to manage all results in a generic way, independently of the routing mode:
- always convert result data to a FeatureCollection MapBox-GL data source, even where there is a single feature
- use [turf/bbox](http://turfjs.org/docs/#bbox) to compute bbox for all kinds of data

## Why
- Display a first iteration of public transport results
- Increase the genericity of how we manage geoms

localhost_3000_routes__origin=addr_2 361615;48 822339_12@12%20Rue%20Philibert%20Lucot destination=latlon_48 87149_2 37028 mode=publicTransport

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -1,4 +1,5 @@
-import { Marker, LngLat, LngLatBounds } from 'mapbox-gl--ENV';
+import { Marker, LngLatBounds } from 'mapbox-gl--ENV';
+import bbox from '@turf/bbox';
 import { map } from '../../config/constants.yml';
 import Device from '../libs/device';
 import layouts from '../panel/layouts.js';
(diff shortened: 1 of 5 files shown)


src/adapters/scene_direction.js
```diff
@@ -73,8 +74,11 @@ export default class SceneDirection {
       if (isActive) {
         mainRoute = route;
       }
-      this.map.setFeatureState({ source: `source_${route.id}`, id: 1 }, { isActive });
+      const geoJson = this.getRouteGeoJson(route.geometry, isActive);
+      this.map.getSource(`source_${route.id}`).setData(geoJson);
```
amatissart: I understand that using `setFeatureState` is not practical to update multiple features. 
But reloading the whole geoJson data seems to be significantly slower. 

Maybe another solution could be to define a routeId property in the features, and update the layer style (to set the `line-color` rule dynamically) ?

bbecquet: Thanks for the suggestion! :ok_hand: 
I did it with the second alternative, as it didn't require manipulating each feature independently, requires a lot less code, and we already did something similar for the pedestrian case.
Now it's generic and snappy :)

PR #358, opened 2019-08-28
Qmaps 852 categories panel on mobile

xem:


src/panel/service_panel_mobile.js
```diff
@@ -0,0 +1,50 @@
+import ServicePanelView from '../views/service_panel_mobile.dot';
+import Panel from '../libs/panel';
+import nconf from '@qwant/nconf-getter';
+import CategoryService from '../adapters/category_service';
+import PanelResizer from '../libs/panel_resizer';
+
+export default class ServicePanelMobile {
```
bbecquet: That's a lot of duplication from the equivalent desktop panel. I guess it's a bit late to ask but wasn't there a way to reuse parts? At least separate view from controller, to have a single panel module managing the logic of both desktop and mobile (which is exactly the same in both cases).

xem: Well, I started like that, but realized that the only common part between desktop and mobile service panels is the 8 first categories and the openFavorites and openDirections functions...

Categories 9-12, good plans, actions (favorites/direction buttons) and resizeability are specific to mobile.

bbecquet: > Categories 9-12, good plans, actions (favorites/direction buttons) [...] are specific to mobile.

I don't think so. It's not because there is no explicit design yet that the feature isn't meaningful to desktop also. Only resizeability is specific to mobile.

Anyway, let's merge like that as it works great.

PR #359, opened 2019-08-28
Improve formatting of route distance/duration

bbecquet:
## Description
- Extract itinerary distance and duration format utilities to a shared file
- Add unit tests to cover these functions
- Add some non-breaking spaces before units and pad minutes

## Why
Improve legibility with respect to common typographic rules.
We should actually do that depending on the locale, but let's improve it later, this fixes most cases.

|Before|After|
|---|---|
|localhost_3000_routes__origin=latlon_48 89360_2 45293 destination=latlon_48 83291_2 31665 mode=cycling (1)|localhost_3000_routes__origin=latlon_48 89360_2 45293 destination=latlon_48 83291_2 31665 mode=cycling|

## Test suite
```
 PASS  tests/units/route_utils.js
  route_utils
    formatDuration
      ✓ Formats 0 seconds as '1&nbsp;min' (4ms)
      ✓ Formats 37 seconds as '1&nbsp;min'
      ✓ Formats 125 seconds as '2&nbsp;min'
      ✓ Formats 3600 seconds as '1&nbsp;h' (1ms)
      ✓ Formats 5100 seconds as '1&nbsp;h&nbsp;25&nbsp;min'
      ✓ Formats 36000 seconds as '10&nbsp;h'
    formatDistance
      ✓ Formats 0 meters as '' (1ms)
      ✓ Formats 15 meters as '15&nbsp;m'
      ✓ Formats 500 meters as '500&nbsp;m'
      ✓ Formats 1234 meters as '1,2&nbsp;km' (1ms)
      ✓ Formats 9999 meters as '10,0&nbsp;km'
      ✓ Formats 123456 meters as '123&nbsp;km'
```

### DIFF
--- a/src/libs/route_utils.js
+++ b/src/libs/route_utils.js
@@ -1,0 +1,30 @@
+
+export function formatDuration(sec) {
+  sec = Math.max(60, sec); // For duration < 60s, return '1 min'
+  let min = Math.round(sec / 60);
+
+  if (min < 60) {
+    return `${min}&nbsp;min`;
+  }
+
+  const hour = Math.floor(min / 60);
+  min = min - 60 * hour;
+  let ret = `${hour}&nbsp;h`;
+  if (min > 0 && hour < 10) {
+    ret += `&nbsp;${min < 10 ? '0' : ''}${min}&nbsp;min`;
+  }
+  return ret;
+}
+
+export function formatDistance(m) {
+  if (m > 99000) {
+    return `${Math.round(m / 1000)}&nbsp;km`;
+  }
+  if (m > 1000) {
+    return `${(m / 1000).toFixed(1).replace('.', ',')}&nbsp;km`;
+  }
+  if (m > 5) {
+    return `${m.toFixed(0)}&nbsp;m`;
+  }
+  return '';
+}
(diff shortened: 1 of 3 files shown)


src/libs/route_utils.js
```diff
@@ -0,0 +1,33 @@
+
+export function formatDuration(sec) {
+  sec = Math.max(60, sec); // For duration < 60s, return '1 min'
+  let min = Math.round(sec / 60);
```
GuillaumeGomez: I'm surprised eslint didn't say that it should be const.

GuillaumeGomez: Oh nevermind. Didn't see below. :)

src/libs/route_utils.js
```diff
@@ -0,0 +1,30 @@
...
+    ret += `&nbsp;${min < 10 ? '0' : ''}${min}&nbsp;min`;
+  }
+  return ret;
+}
+
+export function formatDistance(m) {
+  if (m > 99000) {
+    return `${Math.round(m / 1000)}&nbsp;km`;
+  }
+  if (m > 1000) {
```
GuillaumeGomez: We had a debate about this but I still prefer using `else if` in such cases. But you can completely ignore this comment.

bbecquet: Seriously, this pyramid was ridiculous…
```
if (m > 5) {
  if (m > 1000) {
    if (m > 99000) {
```

GuillaumeGomez: That wasn't my point. I didn't like that pyramid either.

PR #362, opened 2019-09-03
QMAPS-852: hover style

xem:
Fix: restores the hover style of category icons that got disabled in this diff https://github.com/QwantResearch/erdapfel/pull/358/files#diff-1765d48a01be94f5af1b3475b372c0f2L65


bbecquet:
It would be great to always provide at least a minimum PR description, even for the simple ones like this. And when it's a fix of something that was broken previously, a link to the PR that introduced the bug.

PR #364, opened 2019-09-03
React: migrate the road map part of direction panel

bbecquet:
## Description
Migrate the route result map panel to React.
Migrating the full direction panel was too much in one round, so let's start with only the part that display the route results as road map summaries and steps. The rest of the panel (the form) will come next.

Instead of one or two big doT.js templates, I've tried to split everything into a hierarchy of clearly named components:
```
RouteResult
  Route*
    RouteSummary
    RoadMap
      RoadMapStep*
```

As it's better to reduce the number of code branches depending on `Device.isMoObile`, the differences between the two UI are now managed mostly in CSS (thanks grid layout). That means the `<RouteSummary>` and `<RoadMapStep>` components have the same DOM in desktop and mobile, even it they look different on the two device types.

## Why
- React migration progress
- Improve code re-use and responsive UX
- Facilitate evolutions of the direction feature (for example for the public transport integration)

## Screenshots
One route result next to the React devtools component view, to showcase the component hierarchy:
Capture d’écran de 2019-09-03 18-17-44


src/panel/direction/road_map_panel.js
```diff
@@ -1,136 +1,35 @@
...
-      return roadStep;
-    });
-    this.vehicle = vehicle;
-    this.error = false;
-    this.panel.update();
+    ReactDOM.render(
+      <RouteResult
+        routes={routes}
+        vehicle={vehicle}
+        origin={origin.getInputValue()}
```
amatissart: `origin` may be null here. 
For example when called on blur with empty input field: https://github.com/QwantResearch/erdapfel/pull/364/files#diff-f91eb1ec5c604bd61eb88e3aa50ed7f6R79

src/panel/direction/road_map_panel.js
```diff
@@ -1,136 +1,35 @@
...
-    this.panel.update();
-  }
-
-  preview() {
-    this.showRoute = false;
-    this.previewRoadMap.setRoad(this.routes.find(route => route.isActive));
-    this.onOpen();
-    this.panel.update();
-    fire('show_marker_steps');
-    document.querySelector('.map_bottom_button_group').classList.add('itinerary_preview--active');
```
amatissart: this class `itinerary_preview--active` was used (in a ugly way...) to move the geolocation button above the previous/next button during preview.

It seems this behavior is lost now: the geolocation button is not visible
image

We should also check how it behaves once changes from #365 are applied

amatissart: this class `itinerary_preview--active` was used (in a ugly way...) to move the geolocation button above the previous/next button during preview.

It seems this behavior is lost now: the geolocation button is not visible
image

We should also check how it behaves once changes from #365 are applied

src/panel/direction/MobileRoadMapPreview.js
```diff
@@ -0,0 +1,59 @@
+/* globals fire */
```
xem: fire?

src/panel/direction/RoadMap.js
```diff
@@ -0,0 +1,23 @@
...
+    <div className="itinerary_roadmap_step">
+      <div className="itinerary_roadmap_icon itinerary_roadmap_icon_origin">
+        <div className="itinerary_icon_origin_inner" />
+      </div>
+      <div className="itinerary_roadmap_instruction">{`${_('Start')} ${origin}`}</div>
+      <div className="itinerary_roadmap_distance" />
+    </div>
+    {steps.map((step, index) => <RoadMapStep
+      key={index}
+      step={step}
```
xem: Par curiosité:
Lors de la formation react tu nous as dit que key était nécessaire en JSX pour bien gérer les boucles.
Par contre step, c'est un attribut custom? Et sa valeur c'est du JSON qui vient de l'api ? Le parsing de ce json est fait de façon transparente ?

src/panel/direction/RoadMapStep.js
```diff
@@ -0,0 +1,13 @@
...
+const RoadMapStep = ({ step, ...rest }) =>
+  <div className="itinerary_roadmap_step" {...rest}>
+    <div className={`itinerary_roadmap_icon itinerary_roadmap_icon_${getStepIcon(step)}`} />
+    <div className="itinerary_roadmap_instruction">{step.maneuver.instruction}</div>
+    <div className="itinerary_roadmap_distance">
+      {step.distance ? formatDistance(step.distance) : null}
+    </div>
+  </div>;
+
+export default RoadMapStep;
```
xem: Curiosité (suite):
Là par exemple je suis perdu.
Je vois que ce JSX est utilisé dans MobileRoadMapPreview.js et RoadMap.js
je vois d'où vient le paramètre "step" dans ces deux cas, mais pas "rest", ni pourquoi "{...rest}" est inscrit ainsi juste après le className à la ligne 5.

src/panel/direction/RouteResult.js
```diff
@@ -0,0 +1,108 @@
...
+import { getVehicleIcon } from 'src/libs/route_utils';
+import MobileRoadMapPreview from './MobileRoadMapPreview';
+
+export default class RouteResult extends React.Component {
+  static propTypes = {
+    routes: PropTypes.array,
+    origin: PropTypes.string,
+    vehicle: PropTypes.string,
+    isLoading: PropTypes.bool,
+    error: PropTypes.bool,
```
xem: que fait prop-types à part définir un array/objet/string/bool vide ?

src/panel/direction/RouteResult.js
```diff
@@ -0,0 +1,108 @@
...
+      </div>;
+    }
+
+    if (this.props.isLoading) {
+      return <div className="itinerary_leg itinerary_leg--placeholder">
+        <div className="itinerary_leg_summary">
+          <div className={`itinerary_leg_icon ${getVehicleIcon(this.props.vehicle)}`} />
+          <div className="itinerary_leg_via">
+            <div className="itinerary_placeholder-box" style={{ width: '133px' }} />
+            <div className="itinerary_placeholder-box" style={{ width: '165px' }} />
```
xem: Obligés de mettre un objet? est-ce que `style="width:Npx" ` aurait marché aussi?

src/panel/direction/RouteResult.js
bbecquet: Comme expliqué pendant la peer-review, il ne s'agit pas de valeur par défaut, mais de fonctions de typechecking "light". Elles se contentent de vérifier à l'utilisation de ce composant que les props passées vérifient ce format décrit, sinon elles affichent des warnings.

src/panel/direction/road_map_panel.js
bbecquet: Fixed it, thanks.

src/panel/direction/road_map_panel.js
amatissart: This issue persists. Should be it solved here or in a separate PR ?

bbecquet: I'll try to fix it here so it can be merged, this has been stuck too long :)

bbecquet: It should be ok now. I also fixed the direction shortcut button (the one above) who was present in UI states where it's useless. It's still a bit ugly, but at least the fix is done in the old school code which is already full of such imperative DOM manips, so it doesn't change much. And I have good hopes we'll clean that when refactoring the whole panel :)

PR #365, opened 2019-09-04
Qmaps 916 move mobile bottom ui for each panel

xem:
## Description
*Make the bottom UI mobile moveable using an event. Make it generic so any future UI item is also moveable. Use CSS3D (GPU) for moving the UI items.*

## Why
*On mobile, most of the panels (resizeable or not) are on the bottom of the screen, and hide the map's UI. This branch places the UI above each panel, and makes it follow the resize in real time *

## Screenshots
mmm

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -394,24 +394,24 @@ Scene.prototype.onHashChange = function() {
   };
 };
 
+Scene.prototype.translateUIControl = function(selector, bottom) {
+  const item = document.querySelector(selector);
+  if (item) {
+    item.style.transform = `translateY(${-bottom}px)` ;
+  }
+};
+
 Scene.prototype.moveMobileBottomUI = function(bottom = 0) {
   if (Device.isMobile()) {
-    const attrib = document.querySelector('.mapboxgl-ctrl-attrib'); // default bottom: 2px
-    const scale = document.querySelector('.map_control__scale'); // default bottom: 8px
-    const geoloc = document.querySelector('.mapboxgl-ctrl-geolocate'); // default bottom: 10px
-    const direction = document.querySelector('.direction_shortcut'); // default bottom: 65px
-    if (attrib) {
-      attrib.style.bottom = bottom + 2 + 'px';
-    }
-    if (scale) {
-      scale.style.bottom = bottom + 8 + 'px';
-    }
-    if (geoloc) {
-      geoloc.style.bottom = bottom + 10 + 'px';
-    }
-    if (direction) {
-      direction.style.bottom = bottom + 65 + 'px';
-    }
+    const uiControls = [
+      '.mapboxgl-ctrl-attrib',
+      '.map_control__scale',
+      '.mapboxgl-ctrl-geolocate',
+      '.direction_shortcut',
+    ];
+    uiControls.forEach(uiControl => {
+      this.translateUIControl(uiControl, bottom);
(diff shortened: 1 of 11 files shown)


src/adapters/scene.js
```diff
@@ -390,6 +394,27 @@ Scene.prototype.onHashChange = function() {
   };
 };
 
+Scene.prototype.moveMobileBottomUI = function(bottom = 0){
+  if(Device.isMobile()){
```
GuillaumeGomez: Linter won't be happy with your PR. :)

xem: ARG

src/panel/poi_panel.js
```diff
@@ -144,12 +149,14 @@ PoiPanel.prototype.backToSmall = function() {
...
   Telemetry.add(Telemetry.POI_BACKTOFAVORITE);
   window.app.navigateTo('/favs');
+  //fire("move_mobile_bottom_ui", 0);
 };
 
 PoiPanel.prototype.backToList = function() {
   Telemetry.add(Telemetry.POI_BACKTOLIST);
   fire('restore_location');
   window.app.navigateTo(`/places/?type=${this.sourceCategory}`);
+  //fire("move_mobile_bottom_ui", 0);
```
GuillaumeGomez: Is it expected to keep this?

src/panel/poi_panel.js
```diff
@@ -144,12 +149,14 @@ PoiPanel.prototype.backToSmall = function() {
 PoiPanel.prototype.backToFavorite = function() {
   Telemetry.add(Telemetry.POI_BACKTOFAVORITE);
   window.app.navigateTo('/favs');
+  //fire("move_mobile_bottom_ui", 0);
```
GuillaumeGomez: https://t.co/oVtCGk39AC?amp=1

src/panel/poi_panel.js
xem: no, sorry, incomplete cleanup

src/adapters/scene.js
```diff
@@ -390,6 +394,27 @@ Scene.prototype.onHashChange = function() {
   };
 };
 
+Scene.prototype.moveMobileBottomUI = function(bottom = 0){
+  if (Device.isMobile()) {
+    let attrib = document.querySelector(".mapboxgl-ctrl-attrib"); // default bottom: 2px
```
bbecquet: I understand with all these absolute positionings it's not easy to come with a generic solution, but here there's a lot of code duplication and a hard coupling between CSS and JS (the default bottom), which is always a bad idea. There are at least 2 ways to make this better:
1. see if it's possible to use another CSS property than `bottom`, to be able to apply the translation without having to know the initial value. It's common to use `transform: translateY([value]px)` for that. And as every CSS 3D transform, it has the added benefit of being hardware-accelerated, so it's probably better for overall perf.
2. creating a function taking the selector and new bottom as arguments, so the logic is written once and you have only 4 calls… that you can put in a loop so you just have to declare an array of selectors somewhere.
```js
function moveMobileBottomUI(bottom = 0) {
  const uiControls = ['.mapboxgl-ctrl-attrib', '.map_control__scale', '.mapboxgl-ctrl-geolocate', '.direction_shortcut'];
  uiControls.forEach(uiControl => { translateUIControl(uiControl, bottom); });
}
```

PR #366, opened 2019-09-06
React: migrate Masq Status in menu

bbecquet:
## Description
Migrate the Masq account manager in the menu to a React component.
Also simplify some things, like DOM/CSS.
 Seriously, we have a problem with the number of divs and classes we use to display even really simple layouts like that. Let's keep it simple!

## Why
Needed step to react-ify the whole app menu. The menu in itself is simple, except for this masq-related part which was a bit tricky, so I wanted to put it in a separate PR.

## Screenshots
I changed some paddings/margins/etc. to try to use uniform values everywhere, but there is no real difference and we don't care about pixel perfect. Anyway, here are comparison screens:

|Before|After|
|---|---|
|localhost_3000_place_latlon_48 91955_2 38585 (1)|localhost_3000_place_latlon_48 91963_2 38678|
|localhost_3000_place_latlon_48 91955_2 38585|localhost_3000_place_latlon_48 91963_2 38678 (1)|


src/panel/MasqStatus.js
```diff
@@ -0,0 +1,89 @@
+/* globals _ */
```
G-Ray: What about naming new React components with `.jsx` extension ? This way it can be easier to track which components still need to be migrated, and editors like vscode better handle these files by default

src/panel/MasqStatus.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  render() {
+    const user = this.props.user;
+    const isLoggedIn = !!user;
+
+    return <div className="masqStatus">
+      <div className="icon-masq masqStatus__masq_icon" onClick={this.openMasq} />
+      {isLoggedIn
+        ? <Fragment>
```
G-Ray: we can achieve better readability if we split these two components into two render functions like `renderEnabled` and `renderDisabled` for instance, what do you think ?

G-Ray: and Fragment will not be required this way

bbecquet: I'm accustomed to reading stuff like that but it's true that it may be confusing, and there is a clear case here :) I'll change, thanks.

bbecquet: Done :)

src/panel/MasqStatus.js
bbecquet: Why not. That's the kind of things that is important if we want to share common practices with you :) As it requires a change in webpack config, it should be done in a separate PR though.

PR #368, opened 2019-09-09
React: migrate the app menu

bbecquet:
## Description
Migrate the app menu to React without changing its behaviour.

Noticeable things:
- It uses the React `<MasqStatus>` component previously defined in https://github.com/QwantResearch/erdapfel/pull/366

- I added a `<MasqAvatar>` component which factorizes the logic behind the Masq user image or letter fallback. It's used in the menu and the status:
Capture d’écran de 2019-09-09 16-37-24
Capture d’écran de 2019-09-09 16-37-48
Capture d’écran de 2019-09-09 16-38-08
Capture d’écran de 2019-09-09 16-38-14

- I installed the [`classnames` library](http://jedwatson.github.io/classnames/) which facilitates a lot conditional class toggling on React elements. It's used to add the transition classes.

- There are probably a lot of CSS simplifications to do in `menu.scss` to clean things up and get rid of those absolute pixels sizes everywhere, but we will probably introduce some generic text components progressively, so let's not change everything right now.


src/panel/Menu.jsx
```diff
@@ -0,0 +1,124 @@
+/* globals _ */
```
xem: why _ ?

src/scss/includes/masqAvatar.scss
```diff
@@ -0,0 +1,19 @@
...
+    line-height: 1;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+  }
+
+  &-image {
+    width: 100%;
+  }
+}
```
xem: why a new file?

bbecquet: Because it is nice for each single re-usable React component to have a single corresponding Scss file, with coherent naming. That way, it's easier to scope and manage.

It would also allow to import only the SCSS file needed by a component, directly from the JSX file, if we decide to change our CSS building process (for now we build a single big file with Sass, with many unused rules, but there are other ways).

bbecquet:
@xem thanks for the comments. I'll look into the errors you got.

bbecquet:
@xem Thanks the error report, the first warning allowed me to find a bug: the `<MasqStatus>` component was rendered event when Masq was disabled in the app settings. Now it's fixed :)

I don't know about the last 3 errors however (`transitionNode is null`). I think it comes from something else entirely, as I didn't touch this part.

PR #376, opened 2019-09-18
Qmaps 872: 4 new categories

xem:
## Description
*Add 4 PoI categories in the services menu and in the POI "anywhere" panel (search around this place)*

## Why
*We want 12 categories*

## Screenshots
image

NB: the fitness category is labeled as "sport" and uses the "basket ball" icon from the map iconfont.
Later, we will replace the icon with the one made by m.chessa (man lifting a weight)

### DIFF
--- a/src/views/service_panel.dot
+++ b/src/views/service_panel.dot
@@ -12,11 +12,7 @@
   {{? this.categories.length > 0 }}
   <div class="service_panel__categories">
     {{~ this.categories:category:index }}
-      {{?
-        !this.mustDeployCategories ||
-        (this.mustDeployCategories && !this.isDeployed && index < 12) ||
-        (this.mustDeployCategories && this.isDeployed)
-      }}
+      {{? index < 12 }}
         <button class="service_panel__category" type="button" {{= click(this.openCategory, this, category) }}>
           <div class="service_panel__category__icon" style="background: {{= category.backgroundColor }}">
             <span class="{{? index < 9 || index == 10 }}icon{{??}}category__new-icon{{?}} icon-{{= category.iconName }}"></span>
--- a/src/panel/service_panel.js
+++ b/src/panel/service_panel.js
@@ -13,15 +13,8 @@ export default class ServicePanel {
     }
     this.isMobile = Device.isMobile();
     this.categories = CategoryService.getCategories();
-    this.mustDeployCategories = this.categories.length > 8;
-    this.isDeployed = false;
     this.isDirectionActive = nconf.get().direction.enabled;
     this.active = true;
-  }
-
-  toggleCategories() {
-    this.isDeployed = !this.isDeployed;
-    this.panel.update();
   }
 
   openFavorite() {


src/views/poi_panel.dot
```diff
@@ -119,7 +119,7 @@
             {{~ this.categories:category:index }}
               <button class="service_panel__category" type="button" {{= click(this.openCategory, this, category) }}>
                 <div class="service_panel__category__icon" style="background: {{= category.backgroundColor }}">
-                  <span class="icon icon-{{= category.iconName }}"></span>
+                  <span class="{{? index < 9 || index == 10 }}icon{{??}}category__new-icon{{?}} icon-{{= category.iconName }}"></span>
```
bbecquet: Not sure I understand this rule… Why is it using a new CSS class?

src/views/service_panel.dot
```diff
@@ -14,12 +14,12 @@
     {{~ this.categories:category:index }}
       {{?
         !this.mustDeployCategories ||
-        (this.mustDeployCategories && !this.isDeployed && index + 1 < 9) ||
+        (this.mustDeployCategories && !this.isDeployed && index < 12) ||
```
bbecquet: I recently found that the whole "deployment" logic is dead code. `isDeployed` is changed only in the `toggleCategories` function, which is never called anywhere. Maybe we can get rid of all that.

src/views/poi_panel.dot
xem: Yeah, it's a bit dirty for now, as we discussed on the map Thread last week.
9 of the icons are present in the icon font shared with the map (class=icon)
3 new icons are only present in the icomoon font (class=new_icon)
It's agreed to merge like that for now, and add the 3 new icons in the shared font when we have the time (Adrien needs to teach us how to do it)

src/views/service_panel.dot
xem: ok!

xem: The whole "deploy" thing was used in an older design, and doesn't exist anymore. I removed it

src/views/poi_panel.dot
bbecquet: Alright! Thanks for the explanation :)

src/views/service_panel.dot
```diff
@@ -12,11 +12,7 @@
   {{? this.categories.length > 0 }}
   <div class="service_panel__categories">
     {{~ this.categories:category:index }}
-      {{?
-        !this.mustDeployCategories ||
-        (this.mustDeployCategories && !this.isDeployed && index < 12) ||
-        (this.mustDeployCategories && this.isDeployed)
-      }}
+      {{? index < 12 }}
```
bbecquet: I guess you could even remove this test for now, but it's no big deal :)

xem: consider it a "garde-fous" in case someone touches categories.yml

PR #379, opened 2019-09-23
Remove unused esdoc configuration and deps

amatissart:
## Description
`esdoc` was introduced a long time ago in the project but is not useful, as documentation is currently composed of static files only.

* Remove esdoc configuration
* Update dependencies


.gitignore
```diff
@@ -11,12 +11,11 @@ public/build/
 #map style
 public/mapstyle/
 
-#doumentantion
+#documentantion
```
bbecquet: Haha, there were 2 typos
documenta**N**tion

amatissart: fixed

PR #388, opened 2019-10-01
QMAPS-794: add events_enabled feature flag

xem:
## Description
Introduce a feature flags for events

## Why
To implement events in the service panel and dedicated result lists un future tickets


bbecquet:
The default setting definition is missing in https://github.com/QwantResearch/erdapfel/blob/master/config/default_config.yml

PR #393, opened 2019-10-02
QMAPS-1061: use basic-gl-style icons for categories services and health

xem:
## Description
use 2 icons from the basic-gl-style repo instead of the icomoon webfont

make grocery and services icons violet instead of yellow

## Why
so that all categories are displayed similarly in the home and the PoI panels

## Screenshots
image

### DIFF
--- a/src/views/poi_panel.dot
+++ b/src/views/poi_panel.dot
@@ -119,7 +119,7 @@
             {{~ this.categories:category:index }}
               <button class="service_panel__category" type="button" {{= click(this.openCategory, this, category) }}>
                 <div class="service_panel__category__icon" style="background: {{= category.backgroundColor }}">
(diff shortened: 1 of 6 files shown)


src/views/poi_panel.dot
```diff
@@ -119,7 +119,7 @@
             {{~ this.categories:category:index }}
               <button class="service_panel__category" type="button" {{= click(this.openCategory, this, category) }}>
                 <div class="service_panel__category__icon" style="background: {{= category.backgroundColor }}">
-                  <span class="{{? index < 9 || index == 10 }}icon{{??}}category__new-icon{{?}} icon-{{= category.iconName }}"></span>
```
amatissart: `category__new-icon` can also be removed from categories.scss

xem: sure!

amatissart:
@xem `package-lock.json` has been updated to reduce the diff size, and avoid conflicts with other PRs. Make sure to use the latest version of `npm` in your environment.

PR #394, opened 2019-10-03
Directions: highlight route alternative on summary hover

bbecquet:
## Description
Highlight the corresponding route alternative on the map when hovering its summary in the panel.

## Why
Improve the route feature UX. Route summary reacted on mouse over but the route drawing on the map didn't.

## Screenshots
Peek 03-10-2019 13-35

### DIFF
--- a/src/panel/direction/RouteResult.jsx
+++ b/src/panel/direction/RouteResult.jsx
@@ -34,9 +34,19 @@ export default class RouteResult extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.props.routes.length !== prevProps.routes.length) {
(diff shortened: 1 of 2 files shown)


src/panel/direction/RouteResult.jsx
```diff
@@ -34,11 +34,21 @@ export default class RouteResult extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.props.routes.length !== prevProps.routes.length) {
```
bbecquet: This is not really related to this feature but a fix to prevent an error when going from a vehicle mode to another one with less alternatives. We explicitly reset the selected route to the first to avoid having an active route index greater than the number of routes.

PR #406, opened 2019-10-10
Make masq avatar full height

sujeebant:
## Description
The Masq avatar should take fill the whole circle even if it is cropped.

## Why
Esthetics

## Screenshots
It is now looking like this with avatar that have a width greater then their height:

Image Pasted at 2019-10-10 10-33(1)
Image Pasted at 2019-10-10 10-33

We want it to look like this:
Screenshot from 2019-10-10 11-29-51
Screenshot from 2019-10-10 11-30-35

### DIFF
--- a/src/scss/includes/masqAvatar.scss
+++ b/src/scss/includes/masqAvatar.scss
@@ -15,5 +15,7 @@
 
   &-image {
     width: 100%;
+    height: 100%;
+    object-fit: cover;
   }
 }


src/scss/includes/masqAvatar.scss
```diff
@@ -15,5 +15,6 @@
 
   &-image {
     width: 100%;
+    height: 100%;
```
amatissart: `object-fit` could also be used to preserve the aspect ratio:

```suggestion
    height: 100%;
    object-fit: cover;
```

sujeebant: :+1:

PR #407, opened 2019-10-14
Route: Adapt roadmap to the new public transport format

bbecquet:
## Description
Changes the route road map components that assumed routes always had one "leg", to support routes with multiple legs.
As we operate mostly on "steps", it's done by facilitating iteration on all steps, ignoring the legs.

## Why
The public transport route provider changed its API, there is now one route leg by transport type section.

### DIFF
--- a/src/panel/direction/RouteVia.jsx
+++ b/src/panel/direction/RouteVia.jsx
@@ -5,20 +5,16 @@ import PublicTransportLine from './PublicTransportLine';
 const RouteVia = ({ route, vehicle }) => {
   if (vehicle !== 'publicTransport') {
     return <div className="routeVia">
-      {_('Via', 'direction')} {route.summary.replace(/^(.*), (.*)$/, '$1')}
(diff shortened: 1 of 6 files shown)


src/panel/direction/RouteVia.jsx
```diff
@@ -5,20 +5,16 @@ import PublicTransportLine from './PublicTransportLine';
...
   if (vehicle !== 'publicTransport') {
     return <div className="routeVia">
-      {_('Via', 'direction')} {route.summary.replace(/^(.*), (.*)$/, '$1')}
+      {_('Via', 'direction')} {route.legs[0].summary.replace(/^(.*), (.*)$/, '$1')}
     </div>;
   }
 
   return <div className="routeVia">
-    {route.steps
-      // removes multiple 'WALK' parts
```
bbecquet: This data sanitization hack is not needed anymore. Public transport routes now include a `summary` array with info about each leg.

PR #410, opened 2019-10-16
Public transports: display pedestrian parts with the correct style

bbecquet:
## Description
Display pedestrian parts of public transport routes with the same style as pedestrian-only routes.

As `symbol` and `line` styles can't cohabit in the same Mapbox GL layers, this required some refacto in the way route geometries are managed, so a route can be represented by multiple [data source/layer style] couples.

## Why
Increase UX cohesion between public transports and the pedestrian-only itinerary mode.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2019-10-16 14-52-36|Capture d’écran de 2019-10-16 14-51-43|

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -11,12 +11,12 @@ import { getAllSteps } from 'src/libs/route_utils';
 export default class SceneDirection {
   constructor(map) {
     this.map = map;
-    this.routeCounter = 0;
     this.routes = [];
     this.markerOrigin = null;
     this.markerDestination = null;
     this.markersSteps = [];
     this.directionPanel = window.app.directionPanel;
+    this.mapFeaturesByRoute = {};
 
     listen('set_route', ({ routes, vehicle, origin, destination, move }) => {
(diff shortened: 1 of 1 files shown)


src/adapters/scene_direction.js
```diff
@@ -169,29 +170,55 @@ export default class SceneDirection {
...
-    });
-
-    this.map.on('mouseenter', `route_${route.id}`, () => {
-      this.map.getCanvas().style.cursor = 'pointer';
-    });
+  getDataSources(route) {
+    const featureCollection = normalizeToFeatureCollection(route.geometry);
+    const sources = [];
+    if (this.vehicle === 'publicTransport') {
+      const walkFeatures = [], nonWalkFeatures = [];
```
bbecquet: This is a lot of lines for something so simple and common (separate elements of an array in two arrays depending on a criterion). I'd love to introduce Ramda for this kind of stuff [R.partition](https://ramdajs.com/docs/#partition)
But let's discuss it later :)

PR #412, opened 2019-10-16
Qmaps 1045 integrate events in services panel

xem:
## Description
add 6 events in services panel and make style look nice on PC & mobile
there's a feature flag to set: TILEVIEW_events_enabled = true
they're NOT clickable yet (it's another ticket)

## Why
we want events on the homepage

## Screenshots
image
image


src/adapters/category_service.js
```diff
@@ -36,4 +37,17 @@ export default class CategoryService {
     return matchedCategories;
   }
 
+  static getEvents() {
+
```
GuillaumeGomez: Why the empty line?

config/events.yml
```diff
@@ -0,0 +1,37 @@
...
+  color: "#e93865"
+  matcher:
+    regex: "animation"
+  icon: animation
+- label : _('sport')
+  name: sport
+  color: "#e93865"
+  matcher:
+    regex: "sport"
+  icon: sportevent
```
GuillaumeGomez: Missing newline at the end of the file.

src/panel/ServicePanel.jsx
```diff
@@ -8,11 +8,13 @@ class ServicePanel extends React.Component {
   render() {
     return <Panel
       resizable
-      title={_('Qwant Maps services', 'service panel')}
+      title=''
```
GuillaumeGomez: Is `title` still used?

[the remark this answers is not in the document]
bbecquet: I'm not convinced… For me routes, favorites, and now "good plans" are Qwant Maps "services" as well.
And for the scrolling thing, I agree it doesn't look right but could be made more pleasant, for example with a drop shadow. BTW it's how the category list is now and it looks ok.

I also think it's a good thing from a UX point of view that panels have a title which stays visible at all time, so users understand in which context they are. And it would be nice to reach for more cohesion, instead of introducing specific look and feel for each individual panel.

PR #415, opened 2019-10-17
React: migrate POI popups

bbecquet:
## Description
Implements poi map popups with React, making use of `<ReviewScore>` and `<OpeningHour>` components defined in https://github.com/QwantResearch/erdapfel/pull/397.
Also simplify and clean some stuff in the adapter opening the popup.


src/components/PoiPopup.jsx
```diff
@@ -0,0 +1,27 @@
...
+const PoiPopup = ({ poi }) => {
+  const reviews = poi.blocksByType && poi.blocksByType.grades;
+  const openingHours = poi.blocksByType && poi.blocksByType.opening_hours;
+  const address = poi.address && poi.address.label;
+
+  let displayedInfo = null;
+  if (reviews) {
+    displayedInfo = <ReviewScore reviews={reviews} poi={poi} />;
+  } else if (openingHours) {
+    displayedInfo = <OpeningHour poi={poi} />;
```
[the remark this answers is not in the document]
bbecquet: I guess this `OpeningHour` component isn't that generic yet, which is quite obvious when looking at the main CSS class it uses `.poi_panel__info__hours__status__text` :sweat_smile:
I'll refactor that.

src/scss/includes/openingHour.scss
```diff
@@ -0,0 +1,13 @@
...
+  color: #353c52;
+
+  &-circle {
+    width: 6px;
+    height: 6px;
+    display: inline-block;
+    margin: 3px 1px;
+    border-radius: 50%;
+  }
+}
```
amatissart: Old rules defined in `popup.scss` can be removed now. Is it correct ?

```diff
diff --git a/src/scss/includes/popup.scss b/src/scss/includes/popup.scss
index 7c494455..03b7a53c 100644
--- a/src/scss/includes/popup.scss
+++ b/src/scss/includes/popup.scss
@@ -39,15 +39,3 @@
 .poi_popup__address {
   font-size: 14px;
 }
-
-.poi_popup__opening {
-  font-size: 14px;
-}
-
-.poi_popup__title__hour__circle {
-  width: 6px;
-  height: 6px;
-  display: inline-block;
-  margin: 3px 1px;
-  border-radius: 50%;
-}
```

src/scss/includes/openingHour.scss
```diff
@@ -0,0 +1,13 @@
+.openingHour {
+  margin: 0;
+  font-size: 14px;
+  color: #353c52;
```
amatissart: Not specific to this PR (other hard-coded colors remain here and there), but note that an existing variable could be used here:
```suggestion
  color: $primary_text;
```

bbecquet:
@amatissart I updated the PR with your suggestions. Thanks for the missed deletions :)

PR #418, opened 2019-10-22
QMAPS-1071: use a circle as the default icon of PoIs markers

xem:
## Description
change the default icon in the PoI's markers (circle)

## Why
the default "marker2" icon is the one we want in the UI but not on the map, inside the red markers

## Screenshots
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "@qwant/merge-po": "file:local_modules/merge-po",
     "@qwant/nconf-builder": "file:local_modules/nconf_builder",
     "@qwant/po-js": "file:local_modules/po-js",
-    "@qwant/qwant-basic-gl-style": "git+https://github.com/QwantResearch/qwant-basic-gl-style.git#45e60be",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,7 +87,7 @@
     "@qwant/merge-po": "file:local_modules/merge-po",
     "@qwant/nconf-builder": "file:local_modules/nconf_builder",
     "@qwant/po-js": "file:local_modules/po-js",
-    "@qwant/qwant-basic-gl-style": "git+https://github.com/QwantResearch/qwant-basic-gl-style.git#45e60be",
+    "@qwant/qwant-basic-gl-style": "git+https://github.com/QwantResearch/qwant-basic-gl-style.git#94f3fbc",
```
amatissart: Could you update the commit to use the latest master, now that https://github.com/QwantResearch/qwant-basic-gl-style/pull/82 is merged?

And `package-lock.json` needs to be committed too.

PR #422, opened 2019-10-23
QMAPS: reenable right click

xem:
## Description
enable right click on the panels container

## Why
it's weird

### DIFF
--- a/src/views/app_panel.dot
+++ b/src/views/app_panel.dot
@@ -1,5 +1,5 @@
 <div class="side_panel__container">
-  <div class="favorite_poi_panel__container" oncontextmenu="(function(e){e.preventDefault();return false;})(event)">
+  <div class="favorite_poi_panel__container">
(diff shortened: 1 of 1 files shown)


amatissart:
This was introduced by https://github.com/QwantResearch/erdapfel/commit/d123a1dea35787b0c8e67d568c446c046e88b920
So maybe @GuillaumeGomez would have more context to explain this change ?

GuillaumeGomez:
The event was transmissed to the map and was ugly to use. Don't see why it's weird or whatever as is but it's very likely that the bug will reappear now...

amatissart:
It seems weird that it would be propagated to the map. Do you remember how to reproduce the bug exactly ?

GuillaumeGomez:
Not sure... We found it out while looking for something else but I \*think\* that it's by right-clicking on the panel directly (not on any button).

PR #423, opened 2019-10-23
React category panel

bbecquet:
## Description
Migrates the category panel to React.

Thanks to a lot of re-use of generic components (`<Panel>`, `<ItemList>`, `<OpeningHour>`, etc.), we can get rid of a lot of specific CSS and achieve better UX cohesion.

This panel does many things outside of just DOM children rendering, like toggling the special category markers on the map, telemetry, etc. so the root component `<CategoryPanel>` stays big. But a lot of things are cleanier than before, thanks to the explicit React lifecycle methods (`componentDidMount`, `componentDidUpdate` and `componentWillUnmount`)

I tried to separate things into dedicated commits for easier review. The big one is here: https://github.com/QwantResearch/erdapfel/commit/466b4158b275d623a4c4bde0f9058bd44f3c8d58

## Why
React migration.


src/panel/category/CategoryPanel.jsx
```diff
@@ -0,0 +1,227 @@
...
+import layouts from 'src/panel/layouts.js';
+import { sources } from 'config/constants.yml';
+
+const MAX_PLACES = Number(nconf.get().category.maxPlaces);
+
+export default class CategoryPanel extends React.Component {
+  static propTypes = {
+    categoryName: PropTypes.string.isRequired,
+    query: PropTypes.string,
+    bbox: PropTypes.string,
```
amatissart: `categoryName` is actually NOT required. It should be possible to use `query` without any category name.  
And a URL such as `/places/?q=myQuery` should lead to a list of results. This currently raises a JS error. (A test case is probably missing about this :disappointed: )

src/panel/category/CategoryPanel.jsx
```diff
@@ -0,0 +1,227 @@
...
+      panelContent.scrollTop = 0;
+    }
+  }
+
+  componentWillUnmount() {
+    if (!this._navToCategoryMarker) {
+      SearchInput.setInputValue('');
+      fire('remove_category_markers');
+    }
+    unListen(this.clickHandler);
```
amatissart: Removing this listener introduces a regression: it's now impossible to select another marker after a first result has been clicked (and the poi panel has replaced the category panel).

bbecquet:
This PR will be split in several others to ease review and debugging (it started with https://github.com/QwantResearch/erdapfel/pull/429), so let's close it.

PR #426, opened 2019-10-25
Map initialization: code simplification

bbecquet:
## Description
Just a bit of simplification in the map initialization code…
It shouldn't change anything functionally, the only thing that really moved is the performance time recording which could occur a bit later (but which reflects more truthfully the map loading state). If it's a problem, we can move it elsewhere.

## Why
The `scene.js` file does a lot of different things and it's quite hard to read, so let's simplify easy things.

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -7,6 +7,8 @@ import layouts from '../panel/layouts.js';
 import LatLonPoi from '../adapters/poi/latlon_poi';
 import { getRouteStyle, setActiveRouteStyle } from './route_styles';
 import { getAllSteps } from 'src/libs/route_utils';
+import Error from '../adapters/error';
+import nconf from '@qwant/nconf-getter';
 
 export default class SceneDirection {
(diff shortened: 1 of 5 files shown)


src/adapters/scene_direction.js
```diff
@@ -18,6 +20,10 @@ export default class SceneDirection {
     this.directionPanel = window.app.directionPanel;
     this.mapFeaturesByRoute = {};
 
+    const iconsBaseUrl = nconf.get().system.baseUrl + 'statics/images/direction_icons/';
+    this.addMapImage(`${iconsBaseUrl}/walking_bullet_active.png`, 'walking_bullet_active');
+    this.addMapImage(`${iconsBaseUrl}/walking_bullet_inactive.png`, 'walking_bullet_inactive');
```
amatissart: Very minor: this introduces a double slash `direction_icons//...` in the icons URLs.

bbecquet: Thanks, I fixed it!

PR #434, opened 2019-10-31
[REACT] Migrate osm contribution block to react

GuillaumeGomez:


src/components/OsmContribution.jsx
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+
+const OsmContribution = ({ poi }) => {
+  const osmContributionBlock = poi.isFromOSM && poi.isFromOSM();
+  if (!osmContributionBlock) {
+    return null;
+  }
+
+  return <div className="osm_contribute">
+    <div className="osm_contribute__logo"></div>
```
bbecquet: JSX is like XML, all empty components can use the self closing syntax `<div className="osm_contribute__logo" />`. We can install an ESlint rule for that later.

PR #435, opened 2019-10-31
[REACT] Start of "action buttons" migration to react

GuillaumeGomez:
This is the start of the migration of action buttons to react. **However**, it cannot work currently because it requires events, which prevent the static generation we use currently for most elements. Therefore, we need to insert the element inside the DOM ourselves but we need to wait for the parent DOM node to be ready to do so, which is complicated.

### DIFF
--- a/src/components/ActionButtons.jsx
+++ b/src/components/ActionButtons.jsx
@@ -1,0 +1,57 @@
+/* global _ */
+import React from 'react';
+import classnames from 'classnames';
+
+export default class ActionButtons extends React.Component {
+  constructor(props) {
+    super(props);
+    this.elem = props.elem;
+    this.poi = props.poi;
+    this.favoriteClick = this.favoriteClick.bind(this);
+  }
+
+  favoriteClick() {
+    console.log('hello!');
+    click(this.elem.toggleStorePoi, this.elem);
+  }
+
+  render() {
+    const elem = this.elem;
+    const poi = this.poi;
+    const phoneLinkStyle = elem.shouldPhoneBeHidden() ? { display: 'none' } : null;
+    const phone = poi.blocksByType && poi.blocksByType.phone && elem.shouldPhoneBeHidden() &&
+      <button className="poi_panel__action icon-icon_phone poi_phone_container_hidden"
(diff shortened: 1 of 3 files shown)


src/components/ActionButtons.jsx
```diff
@@ -0,0 +1,57 @@
+/* global _ */
+import React from 'react';
+import classnames from 'classnames';
+
+export default class ActionButtons extends React.Component {
+  constructor(props) {
+    super(props);
+    this.elem = props.elem;
```
bbecquet: You shouldn't do that in React. Each prop must be accessed via the `this.props` object, which is kept up-to-date when props are changed by the parent component. If you store and use the value like that without adding the code to update it, you'll render outdated prop values.

src/components/ActionButtons.jsx
```diff
@@ -0,0 +1,57 @@
...
+    this.poi = props.poi;
+    this.favoriteClick = this.favoriteClick.bind(this);
```
bbecquet: There is a simpler alternative to that : just declare `favoriteClick` as an arrow function.
And now your constructor is empty, so you can remove it :)

src/components/ActionButtons.jsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  favoriteClick() {
+    console.log('hello!');
+    click(this.elem.toggleStorePoi, this.elem);
+  }
+
+  render() {
+    const elem = this.elem;
```
bbecquet: According to the previous comment, you should do `const { elem, poi } = this.props;` instead.

src/components/ActionButtons.jsx
```diff
@@ -0,0 +1,57 @@
...
+    const poi = this.poi;
+    const phoneLinkStyle = elem.shouldPhoneBeHidden() ? { display: 'none' } : null;
```
bbecquet: In React you'll prefer not to render the element at all, and keep the display status as a state value in the component.

bbecquet:
Done as part of https://github.com/QwantResearch/erdapfel/pull/461

PR #436, opened 2019-10-31
[REACT] migrate accessibility block to react

GuillaumeGomez:
:)


src/views/poi_bloc/Accessibility.jsx
```diff
@@ -0,0 +1,15 @@
+/* global _ */
+import React from 'react';
+
+const AccessibilityBlock = ({ x, _y, z }) => {
+  const accessibilityList = x;
```
bbecquet: You could just do these assignments by deconstructing the params.

src/panel/poi_bloc/poi_bloc_container.js
```diff
@@ -1,14 +1,28 @@
...
 PoiBlocContainer.initBlockComponents = function() {
   PoiBlocContainer.blockComponents = constants.pois.reduce((accBlocks, poiBlock) => {
+    let builder;
+    try {
+      builder = require(`./${poiBlock.panelName}_panel`);
+    } catch (err) {
+      const name = poiBlock.panelName.charAt(0).toUpperCase() + poiBlock.panelName.slice(1);
+      const el = require(`../../views/poi_bloc/${name}`);
+      builder = {
+        default: (x, y, z) => ReactDOMServer.renderToString(<el.default x={x} y={y} z={z} /> ),
```
bbecquet: For this case it's better to use `renderToStaticMarkup` as it produces cleaner DOM https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup
And even better, using `src/libs/renderStaticReact` which does that for you by hiding the dirty detail, as using the `ReactDOMServer` on the client is really a hack.

PR #442, opened 2019-11-05
Migrate blocks to react

GuillaumeGomez:
First commit can be tested with this path: "place/osm:node:2685527894@OSullivans_Rebel_Red#map=18.00/48.8591750/2.3491722"

Second commit can be tested with this path: /place/osm:node:676860286@By_My_Car_Rivoli#map=18.00/48.8590893/2.3481345


GuillaumeGomez:
Updated.

PR #444, opened 2019-11-06
Update Masq to v0.14.3

sujeebant:
## Description
Update Masq-lib

## Why
Updates signalhubws dependency (which removes some useless dependencies such as websocket package)


amatissart:
Thanks @sujeebant. I just rebased the branch.

@xem could you check that `package-lock.json` is stable now on your side after runnning `npm install` ?

PR #445, opened 2019-11-06
migrate Hour block to react

GuillaumeGomez:
This PR isn't ready to be merged as is!

The point here is just to migrate the global style. It still misses the implementation of the toggle button to show/hide the schedules.


GuillaumeGomez:
Superseded by #458.

PR #447, opened 2019-11-06
Migrate images block to react

GuillaumeGomez:


src/views/poi_bloc/Images.jsx
```diff
@@ -0,0 +1,55 @@
+/* global _ */
+import React from 'react';
+import classnames from 'classnames';
+
+function createImage(param, classes) {
```
bbecquet: This function which renders HTML is actually… a React component :) So it would be nicer to write it the right way, so you can use it like `<Image …>`.
And there is no problem to keep it in the same file if it's used only here.

src/views/poi_bloc/Images.jsx
```diff
@@ -0,0 +1,55 @@
...
+  return <a
+    target='_blank'
+    rel='noopener noreferrer'
+    href={ param.href }
+    className={ classes }
+    style={{ backgroundImage: param.url }}
+  />;
+}
+
+function createImages(images) {
```
bbecquet: I know this is a direct transcription of the current code but I'm sure the flow logic can be improved and simplified to just look like a loop over the image array, not having to test all these special cases of 1, 2, etc. But it probably requires some CSS changes, so maybe we can merge like that and improve later.
What do you prefer?

src/views/poi_bloc/Images.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+const ImagesBlock = ({
+  block,
+  poi,
+  asString,
+}) => {
+  const images = block.images.filter(img => img.url !== poi.topImageUrl);
+  if (asString) {
+    return '';
+  } else if (!images || images.length === 0) {
```
bbecquet: These two conditions can be grouped with the previous test.

src/views/poi_bloc/Images.jsx
```diff
@@ -0,0 +1,55 @@
...
+function createImages(images) {
+  if (images.length < 2) {
+    return;
+  } else if (images.length === 2) {
+    return <div className='poi_panel__pictures_block poi_panel__pictures_second_block'>
+      { createImage(images[1], 'poi_panel__pictures_block_full') }
+    </div>;
+  }
+  return <div
+    className={ classnames(
```
bbecquet: No real need for `classnames` if you don't use its conditional class toggling capabilities.

src/views/poi_bloc/Images.jsx
GuillaumeGomez: If you have an idea on how to make it more straightforward, I'm totally fine into reworking it. I just don't know how to make it better. :-/

src/views/poi_bloc/Images.jsx
GuillaumeGomez: I voluntarily didn't group them to keep both code logics separated.

src/views/poi_bloc/Images.jsx
GuillaumeGomez: It allows to make a nicer display but I can remove it if you prefer.

src/views/poi_bloc/Images.jsx
GuillaumeGomez: Oh good point! Not used enough to react I am...

bbecquet: :) If you do that just pass `href` and `url` as separate props, so the low level component doesn't have to know how to access info.

src/views/poi_bloc/Images.jsx
bbecquet: Actually, after checking it, I would just remove the `asString` prop completely, because the `images_panel` doesn't implement `toString` so this component doesn't need to implementthat.

And for the other test, it's better to return `null`, because with the empty string React generates stuff in the DOM (historically an empty `<span>`, now comments).

src/views/poi_bloc/Images.jsx
bbecquet: `Nicer display`? You mean code? This isn't important compared to importing smth we don't really use :)

src/views/poi_bloc/Images.jsx
bbecquet: The general idea would be to always generate the images with an array loop, not caring about their number, and implementing the different cases in CSS only because I think it's possible. Or at least it's possible with just a specific class on the container.
But I don't know if you like playing with CSS, that's why I asked for your preference. I can do it later if you want :)

GuillaumeGomez:
Updated!

src/views/poi_bloc/Images.jsx
GuillaumeGomez: I'm not sure to fully understand what you have in mind so I'd prefer to let you handle it if you don't mind.

src/views/poi_bloc/Images.jsx
```diff
@@ -0,0 +1,32 @@
...
+  if (!images || images.length === 0) {
+    return null;
+  }
+
+  return <div className="poi_panel__pictures">
+    <h5 className="poi_panel__pictures_title">
+      <i className="icon-icon_camera"></i>
+      { _('Photos') }
+    </h5>
+    <div className='poi_panel__pictures_tiles'>
```
bbecquet: Can you please keep consistent attribute quotes? You used `"` above, and `'` here and below. `"` is the preferred way, used everywhere else in the code.
And yes I know this should be an ESLint rule, I'll add it later :) And I'm only asking for this change because you have other real linting errors, so you can fix it at the same time.

bbecquet: No? :crying_cat_face:

GuillaumeGomez:
Updated.

PR #448, opened 2019-11-07
Migrate internet access

GuillaumeGomez:


src/panel/poi_bloc/poi_bloc_container.js
```diff
@@ -13,7 +13,9 @@ PoiBlocContainer.initBlockComponents = function() {
     try {
       builder = require(`./${poiBlock.panelName}_panel`);
     } catch (err) {
-      const name = poiBlock.panelName.charAt(0).toUpperCase() + poiBlock.panelName.slice(1);
+      const name = poiBlock.panelName.split('_')
```
bbecquet: Maybe to avoid this hack we should just put the name in the correct form in the `constants.yml` file :man_shrugging:

src/views/poi_bloc/InternetAccess.jsx
```diff
@@ -0,0 +1,18 @@
+/* global _ */
+import React from 'react';
+
+const InternetAccessBlock = ({
+  block,
+  asString,
+}) => {
+  if (block.wifi === undefined) {
```
bbecquet: `if (!block.wifi) {` is safer for this king of test, as it manages `undefined`, `null` and the empty string.

src/views/poi_bloc/InternetAccess.jsx
```diff
@@ -0,0 +1,18 @@
...
+    return null;
+  }
+  const text = `${_('Internet access', 'poi')} : ${_('WiFi', 'poi')}`;
+  if (asString) {
+    return text;
+  }
+  return <div>{ text }</div>;
```
bbecquet: This isn't the same markup as the dot component.

src/views/poi_bloc/Wiki.jsx
```diff
@@ -0,0 +1,25 @@
...
+      <p className="poi_panel__description__ellipsis">
+        { block.description }
+      </p> }
+    <br/>
+    { block.url &&
+      <a
+        className="poi_panel__info__wiki__link"
+        rel="noopener noreferrer"
+        target="_blank"
+        href={ ExtendedString.htmlEncode(block.url) } >
```
bbecquet: I'm sure this one is not needed, especially here for a url. It will just make a mess if it includes `&` character, resulting in a wrong url.

src/views/poi_bloc/Wiki.jsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react';
+import ExtendedString from '../../libs/string';
+
+const WikiBlock = ({
+  block,
+}) => {
+  return <div className="poi_panel__info__wiki">
+    { block.description &&
+      <p className="poi_panel__description__ellipsis">
+        { block.description }
```
bbecquet: I noticed you removed `htmlEncode` form `block.description`. I don't think it's useful there, otherwise if the string can contain HTML meant to be interpreted, [dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) will be required.

GuillaumeGomez:
Updated.

PR #451, opened 2019-11-08
Really finish the category panel migration to React

bbecquet:
## Description
Finally get rid of the intermediary layer of code between `app_panel` and the React `CategoryPanel`. This layer was kept until then because the management of markers on the map wasn't entirely possible with the React life cycle methods, as the category markers could persist even if the panel was unmounted.
I solved that by moving the management of category markers to the `scene_category` file only. And the operations that can be controlled by the panel life-cycle (for example data loading) have been ketp there.
A dirty thing remaining is that the conditional removal of category markers has now to be made during panel switching, i.e. in `app_panel`.

## Why
Simplification of code.
Really instantiate a React component in `app_panel`, so we can migrate to a full React panel manager soon.


src/panel/category/CategoryPanel.jsx
```diff
@@ -41,17 +124,38 @@ export default class CategoryPanel extends React.Component {
...
+    fire('highlight_category_marker', poi, highlight);
+  }
+
   render() {
-    const { pois, dataSource, hasError, zoomIn, close, selectPoi, highlightPoiMarker } = this.props;
+    const { loading, pois, dataSource } = this.state;
+
+    if (loading) {
+      return null;
+    }
```
amatissart: As a consequence the panel will flicker when moving the map. Is it an expected behavior ?

src/panel/category/CategoryPanel.jsx
```diff
@@ -41,17 +124,38 @@ export default class CategoryPanel extends React.Component {
...
+
+    const hasError = !pois || pois.length === 0;
+    const zoomIn = !this.pois;
```
amatissart:
```suggestion
    const zoomIn = !pois;
```

src/panel/category/CategoryPanel.jsx
bbecquet: Indeed. I forgot a condition. Thanks!

bbecquet: As we don't have any loading indicator for now and the only time when the panel is not displayed is during the first load, I changed this state boolean to `initialLoading`. Now it behaves like before.

PR #455, opened 2019-11-14
Routes: path outlines

bbecquet:
## Description
Adds a darker outline effect to route paths so they stand out more visibly from the map background.

Unfortunately, Mapbox-GL hasn't been thought for that: there is no way of applying a border or shadow to line items with style only and there is now way to apply an arbitrary JS function to dynamic values extracted with style expressions.
To get around these limitations, I had to draw each line twice:
- a first one with a darker color (computed in advance using a dedicated library) and large width.
- a second one on top of the first with the normal color and a smaller width.

Capture d’écran de 2019-11-14 11-16-18

This works and looks great but this involves some data duplication and doubles the drawing job for Mapbox-GL, which may impact performances. Maybe it's worth putting this feature behind a flag we can toggle conditionally…

**What do you think?**

## Why
Improve the legibility of route paths on the map, especially for public transport lines which can have low contrast colors and for grayed out alternatives.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2019-11-14 11-29-12|Capture d’écran de 2019-11-14 11-28-46|
|Capture d’écran de 2019-11-14 11-21-47|Capture d’écran de 2019-11-14 11-26-54|

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3938,6 +3938,15 @@
         "object-visit": "^1.0.0"
       }
     },
+    "color": {
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/color/-/color-3.1.2.tgz",
+      "integrity": "sha512-vXTJhHebByxZn3lDvDJYw4lR5+uB3vuoHsuYA5AKuxRVn5wzzIfQKGLBmgdVRHKTJYeK5rvJcHnrd0Li49CFpg==",
+      "requires": {
+        "color-convert": "^1.9.1",
+        "color-string": "^1.5.2"
+      }
+    },
     "color-convert": {
       "version": "1.9.3",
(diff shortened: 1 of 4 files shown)


bbecquet:
|Version|Lignes 14 et 7|Ligne 8|
|---|---|---|
|**Actuelle**. Épaisseur 7, noir 20%|outline_7_02_b|outline_7_02_a|
|Épaisseur 7, noir 33%|outline_7_033_b|outline_7_033_a|
|Épaisseur 7, noir 33%|outline_7_05_b|outline_7_05_a|
|Épaisseur 8, noir 20%|outline_8_02_b|outline_8_02_a|
|Épaisseur 8, noir 33%|outline_8_033_b|outline_8_033_a|

PR #456, opened 2019-11-14
Migrate poi blocks

GuillaumeGomez:
The goal here is to make the code more straightforward and to facilitate the migration of the poi bloc container item.


src/views/poi_bloc/Accessibility.jsx
```diff
@@ -3,13 +3,24 @@ import React from 'react';
...
+    },
+    'toilets_wheelchair': {
+      'yes': _('Wheelchair accessible toilets'),
+      'partial': _('Partial wheelchair accessible toilets'),
+      'no': _('No wheelchair accessible toilets'),
+    },
+  };
+  const availableAccessibilities = [];
+  for (const label in labels) {
+    availableAccessibilities.push(_(labels[label][accessibilityList[label]]));
```
bbecquet: Now that everything has been regrouped in the component, do we really need this mess that tries to be generic? We can probably do that in a more readable way.
Also, I think there are now two calls to the `_` function.

GuillaumeGomez: Good catch for the double `_` calls!

About the generic part, it's still needed. We make the display depending on the argument we receive after all.

bbecquet: Ok, I guess we can improve that later if it's needed, as long as it's self contained this isn't a big deal. But I still find this pretty ugly :) (not your fault though)

PR #461, opened 2019-11-20
migrate poi panel

GuillaumeGomez:
# TODO

- [x] Restore `emptyClickOnMap` behavior
- [x] Fix store event listeners, shouldn't be registered by `<ActionButtons>`
- [x] Fix category markers not persistent anymore when coming from a category list
- [x] Fix integration tests
- [x] Fix POI map marker still displayed after close
- [x] Fix loading POI from search suggestion
- [x] Fix repositioning of map UI controls when displaying POI as card
- ~~[ ] Re-use `<Panel>` and better split card display~~ => Later
- ~~[ ] Factorize category buttons block with the one in Service panel~~ => Later


src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+    }
+  } catch (e) {
+    return false;
+  }
+  return false;
+}
+
+export default class PoiPanel extends React.Component {
+  static propTypes = {
+    poi: PropTypes.object,
```
bbecquet: The `isRequired` modifier would be useful, as this panel cannot exist without being given a POI.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+    isFromFavorite: PropTypes.bool,
+    isFromCategory: PropTypes.bool,
+    sourceCategory: PropTypes.object,
+  }
+
+  constructor(props) {
+    super(props);
+    console.log(props);
```
bbecquet: console.log

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+    });
+
+    store.eventTarget.addEventListener('poi_added', async () => {
+      if (this.props.poi && !this.props.poi.stored) {
+        this.props.poi.stored = await isPoiFavorite(this.props.poi);
+        this.update();
+      }
+    });
+
+    this.expandCollapse = () => {
```
bbecquet: This would be far more readable as individual methods. Try to have the smallest constructor possible.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+      this.setState(state => ({
+        isCollapsed: !state.isCollapsed,
+      }));
+    };
+    this.backToFavorite = () => {
+      if (!window.app) {
```
bbecquet: Why these checks? If there is a moment where `app` isn't intialized yet and this is called, we need to change smth in the flow.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+            id: this.props.poi.id,
+            source: this.props.poi.meta.source,
+            template: 'single',
+            zone: 'detail',
+            element: 'more',
+          })
+        );
+      }
+      this.card = false;
+      // re-render
```
bbecquet: If you need a re-render, update a state value.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+
+  constructor(props) {
+    super(props);
+    console.log(props);
+    this.state = { isCollapsed: true };
+    this.categories = CategoryService.getCategories();
+    this.isDirectionActive = nconf.get().direction.enabled;
+    this.isMasqEnabled = nconf.get().masq.enabled;
+
+    this.card = true;
```
bbecquet: This should probably be a state value.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+    isPoiFavorite(this.props.poi).then(x => this.props.poi.stored = x);
+    this.active = true;
```
bbecquet: This isn't needed anymore in the React panels model.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+  constructor(props) {
+    super(props);
+    console.log(props);
+    this.state = { isCollapsed: true };
+    this.categories = CategoryService.getCategories();
+    this.isDirectionActive = nconf.get().direction.enabled;
+    this.isMasqEnabled = nconf.get().masq.enabled;
+
+    this.card = true;
+    isPoiFavorite(this.props.poi).then(x => this.props.poi.stored = x);
```
bbecquet: React components shouldn't have side effects on data, meaning you must not modify what is passed as props. This should probably be done outside, or as state.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+    this.active = true;
+
+    store.onToggleStore(async () => {
+      if (this.props.poi) {
+        this.props.poi.stored = await isPoiFavorite(this.props.poi);
+        this.update();
```
bbecquet: There is no more `update` method. And my previous remark about modifying props is still valid. `stored` looks like something which should be put in state (and given a better name probably)

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+      }
+    });
+
+    store.eventTarget.addEventListener('poi_added', async () => {
```
bbecquet: Actually I remember what this is, global event on the store… This has probably nothing to do in this panel at all.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+            </button>
+          </div>
+        </div>
+        <div className="poi_panel__content">
+          <div className="poi_panel__container">
+            <div className="poi_panel__description_container" onClick={this.center}>
+               <PoiHeader poi={this.props.poi} />
+               <PoiTitleImage poi={this.props.poi} iconOnly={false} />
+            </div>
+            <div className="poi_panel__actions">
```
bbecquet: It would be great to put the actions in a dedicated component.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,345 @@
...
+              {this.isDirectionActive &&
+                <button className="poi_panel__action icon-corner-up-right"
+                  onClick={this.openDirection}>
+                  <div>{_('DIRECTIONS', 'poi')}</div>
+                </button>
+              }
+              {this.props.poi.blocksByType && this.props.poi.blocksByType.phone &&
+                this.renderPhone()}
+            </div>
+            {isAnywherePoi && this.categories &&
```
bbecquet: This is exactly the same visual elements as the category list on the service panel, but maybe the factorization can be done in a future PR. In any case, it would be more convenient as a separate component, it would make this one less polluted by category stuff.

src/adapters/poi/idunn_poi.js
```diff
@@ -25,8 +24,9 @@ export default class IdunnPoi extends Poi {
       lat: rawPoi.geometry.coordinates[LAT_INDEX],
       lng: rawPoi.geometry.coordinates[LNG_INDEX],
     };
+    rawPoi.kind = 'idunn';
     super(rawPoi.id, rawPoi.name, alternativeName, rawPoi.type, latLng, rawPoi.class_name,
-      rawPoi.subclass_name);
+      rawPoi.subclass_name, rawPoi);
```
bbecquet: For other classes, you pass a string, not an object.

src/adapters/poi/poi_store.js
```diff
@@ -30,4 +31,16 @@ export default class PoiStore extends Poi {
...
+  static deserialize(raw) {
+    const {
+      id, name, alternativeName, type, latLon, className, subClassName, bbox, blocks,
+      topImageUrl, kind,
+    } = raw;
+    if (kind === 'idunn') {
+      return new IdunnPoi(raw);
+    }
+    return new Poi(id, name, alternativeName, type, latLon, className, subClassName, bbox, blocks,
+      topImageUrl, raw);
```
bbecquet: If I understand correctly your changes in the Poi class constructor, all `raw` fields will overwrite stuff that is explicitly pass as arguments. Maybe what you need is to destructure and pass only other unnamed fields. Like:
```js
const {
  id,
  name,
  // other named fields
  ...extras,
} = raw;
// ...
return new Poi(id, name, /* named fields */, extras);
```

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+  static propTypes = {
+    poi: PropTypes.object.isRequired,
+    isFromFavorite: PropTypes.bool,
+    isFromCategory: PropTypes.bool,
+    sourceCategory: PropTypes.string,
+  }
+
+  constructor(props) {
+    super(props);
+    this.poi = props.poi;
```
bbecquet: You shouldn't do that in React, just use `this.props` directly, and if you need to compute things from props, do it in `render`. Otherwise this is the best way to have bugs from data inconsistency. For example, if `props` change with the component already mounted, which will happen all the time, the constructor won't be called so `this.poi` won't be kept up-to-date.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+    if (props.poi) {
+      this.poi = PoiStore.deserialize(props.poi);
+    }
+    this.state = {
+      isCollapsed: true,
```
bbecquet: This value doesn't seem to be used at all.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+        return;
+      }
+      fire(
+        'move_mobile_bottom_ui',
+        elem.offsetHeight + 20
+      );
+    });
+  }
+
+  expandCollapse = () => {
```
bbecquet: Related to the previous comment, I think this can be removed.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+    }));
+  }
+
+  backToFavorite = () => {
+    Telemetry.add(Telemetry.POI_BACKTOFAVORITE);
+    window.app.navigateTo('/favs');
+  }
+
+  showDetail = () => {
+    if (this.poi.meta && this.poi.meta.source) {
```
bbecquet: Here and everywhere else: `this.poi` should be replaced by `this.props.poi`.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+      data.text = _('Back to favorite');
+    } else if (this.props.isFromCategory) {
+      data.callback = this.backToList;
+      data.text = _('Back to list');
+    } else if (!this.state.card) {
+      data.callback = this.backToSmall;
+      data.text = _('Back');
+      data.className = 'poi_panel__back_mobile';
+    }
+    return <div>
```
bbecquet: This should re-use the `<Panel>` component somewhere. I know things are complicated by the mobile display which uses a card, but otherwise that's totally what the `<Panel>` component has been created for. This way you don't need to care about many classnames, the close button, etc.

src/panel/PoiPanel.jsx
```diff
@@ -0,0 +1,237 @@
...
+    isFromFavorite: PropTypes.bool,
+    isFromCategory: PropTypes.bool,
+    sourceCategory: PropTypes.string,
+  }
+
+  constructor(props) {
+    super(props);
+    this.poi = props.poi;
+    if (props.poi) {
+      this.poi = PoiStore.deserialize(props.poi);
```
bbecquet: Looks like it should be done outside of this component, so you can just assume here that `this.props.poi` is an object of the right kind.

src/panel/PoiPanel.jsx
bbecquet: Let's keep it for after this PR is merged

GuillaumeGomez:
Can't approve my own PR but :+1:. Might need some squashes though considering that a few commits are fixes of bugs added in the same PR (like the field `data` that needed to be `text`).

PR #463, opened 2019-11-20
[WIP] QMAPS-1046 : events list

xem:
## Description
- Create events yml and jsx files inspired by Category
- Perform event type check (return to home if unknown)
- Perform AJAX request to the events API endpoint
- WIP: display list items and markers properly

## Why
To have events in Qwant Maps!

## Screenshots

image

image


src/adapters/poi/idunn_poi.js
```diff
@@ -87,6 +87,36 @@ export default class IdunnPoi extends Poi {
     }
   }
 
+  static async poiEventLoad(bbox, size, category, query) {
+    const url = `${serviceConfig.idunn.url}/v1/events`;
+    const requestParams = { bbox, size };
+    if (category) {
+      requestParams['category'] = category;
+    }
+    if (query) {
```
bbecquet: So is this parameter useful after all?

src/adapters/poi/idunn_poi.js
```diff
@@ -87,6 +87,36 @@ export default class IdunnPoi extends Poi {
...
+      console.log(response);
+      response.events = response.events.map(rawPoi => new IdunnPoi(rawPoi));
+      return response;
+    } catch (err) {
+      if (err === 400 || err === 404 ) {
+        return {};
+      } else {
+        const s_requestParams = JSON.stringify(requestParams);
+        Error.sendOnce(
+          'idunn_poi', 'poiApiLoad',
```
bbecquet: `poiApiLoad` => `poiEventLoad`

src/panel/event/EventListPanel.jsx
```diff
@@ -0,0 +1,132 @@
...
+    if (panelContent) {
+      panelContent.scrollTop = 0;
+    }
+  }
+
+  componentWillUnmount() {
+    window.unListen(this.mapMoveHandler);
+  }
+
+  fitMapAndFetch() {
```
bbecquet: I wonder if this dirty stuff, among other things, can be factorized between categories and events… I really thought events were going to be more distinct than categories, but for now it looks exactly the same :|

xem: Yes, I think a lot of things will be in common... basically, everything besides the list header, list items's look and the fact that we keep osm pois or not. And as you mentioned earlier, categories need a "query" param but events don't.

src/adapters/poi/idunn_poi.js
xem: no, I cleaned it up but forgot to ûsh my last commit due to a surprise meeting :)

src/panel/app_panel.js
```diff
@@ -281,6 +290,19 @@ export default class AppPanel {
     });
   }
 
+  openEvents(params) {
+    const { type: eventName, q: query, ...otherOptions } = params;
```
bbecquet: The `q: query` part can be removed here and everywhere else (props especially), as the API doesn't support (or at least it's not a parameter used by `poiEventLoad`).

src/panel/app_panel.js
```diff
@@ -21,24 +22,28 @@ import Router from 'src/proxies/app_router';
 import Poi from 'src/adapters/poi/poi.js';
 import layouts from './layouts.js';
 import ReactPanelWrapper from 'src/panel/reactPanelWrapper';
+import events from '../../config/events.yml';
```
bbecquet: Just a note that you can also write `config/events.yml` as there is an alias defined.

src/panel/app_panel.js
```diff
@@ -254,12 +268,7 @@ export default class AppPanel {
     this.activePoiId = null;
     this.panels
       .filter(panel => panel !== panelToOpen)
-      .forEach(panel => {
-        if (panel === this.categoryPanel) {
```
bbecquet: If you remove this test, the category markers won't get removed after leaving the category panel. Instead of removing it, I think you need to adapt it for the `eventListPanel` instead.

src/panel/event/EventListPanel.jsx
```diff
@@ -0,0 +1,155 @@
...
+// import PoiCategoryItemList from '../category/PoiCategoryItemList';
+
+const eventConfig = nconf.get().events;
+const MAX_PLACES = Number(eventConfig.maxPlaces);
+
+class EventListPanel extends React.Component {
+
+  static propTypes = {
+    eventName: PropTypes.string,
+    query: PropTypes.string,
```
bbecquet: Related to the comment above, `query` is not used so can be removed.

src/panel/app_panel.js
```diff
@@ -254,12 +268,7 @@ export default class AppPanel {
...
-          fire('remove_category_markers');
-        }
-        panel.close();
-      });
+      .forEach(panel => { if (panel.close) { panel.close(); }});
```
bbecquet: All panels have a `close` method, I don't think you need this test.

src/panel/app_panel.js
xem: indeed, I think a bad manip made me remove these lines. Fixed :)

src/panel/app_panel.js
xem: indeed, I added it when I encountered "no function panel.close" errors, but the problem didn't come from here!

src/panel/event/PoiEventItem.jsx
```diff
@@ -0,0 +1,38 @@
...
+import PoiTitleImage from 'src/panel/poi/PoiTitleImage';
+
+const PoiEventItem = ({ poi, eventName }) => {
+  const address = poi.address || {};
+
+  const blocks = poi.blocks;
+  const dates_block = blocks.find(block => block.type === 'event_opening_dates') || {};
+  let date_start, date_end, dates_string;
+
+  if (dates_block) {
```
bbecquet: I thought you wanted to do that in another PR and use the Intl API…

PR #464, opened 2019-11-21
React-ify Direction point inputs

bbecquet:
## Description
Migrate the direction panel input fields.
As for the RouteResult part, it can be done separately, mounting the React component when needed in the Panel.
After that we'll be able to migrate the vehicle selector and finally the full direction panel itself.

The `<DirectionInput>` component introduced still uses the same suggestion widget as the previous version, as it works fine and is full vanilla-js. It can wait longer for a migration to React if needed.

The migration simplifies `direction_panel.js` as stuff like focus and style management are now delegated to the input components.

There is some level of complexity though to manage the state of the inputs, as there are still two distinct values to keep: the raw input string as entered by the user and the POI instance which results from the async geocoding.


src/panel/direction/DirectionInput.jsx
```diff
@@ -0,0 +1,109 @@
...
+      <div className={`itinerary_icon itinerary_icon_${pointType}`} />
+      <input
+        ref={this.inputRef}
+        id={`itinerary_input_${pointType}`}
+        className="itinerary_input"
+        type="search"
+        required
+        autoComplete="off"
+        spellCheck="false"
+        placeholder={_(pointType === 'origin' ? 'Start point' : 'End point', 'direction')}
```
amatissart: this syntax will probably not be caught by gettext script:  
`_('Start point')` and `_('End point')`  should appear separately.

src/panel/direction/DirectionInput.jsx
```diff
@@ -0,0 +1,109 @@
...
+  componentWillUnmount() {
+    if (this.suggest) {
+      this.suggest.destroy();
+    }
+  }
+
+  onChange = event => {
+    const input = event.target.value;
+    this.props.onChangePoint(input, null);
+    this.suggest.setValue(input);
```
amatissart: As far as I understand, this function is called on each keystroke.
And as a consequence, the suggestions list is flickering between requests.
Maybe due to how `suggest.setValue` resets the suggestions list ?

Actually this looks redundant with the existing mechanism in Autocomplete that creates listeners on "input" events.

src/panel/direction/DirectionInput.jsx
bbecquet: Fixed!

src/panel/direction/DirectionInput.jsx
bbecquet: Simply removing the line solved the problem, meaning the change handler registered by the autocomplete widget continues to work fine on a React-controlled input.

PR #484, opened 2019-12-06
Add GitHub action to track Webpack bundle sizes

bbecquet:
## Description
In an effort to track the evolution of our app size over time, I'd like to experiment with https://app.packtracker.io.
Among other features, this should be able to display a delta of the bundle sizes for each PR, making it easy to detect features which have a bad impact on the app size.

This requires to configure a GitHub action to hook the repo to Packtracker.
As Erdapfel is open-source, Packtracker and GitHub actions are free services.

## Screenshots
When correctly configured, we should see this kind of message on PR
assets_-LKja-OcY-MeZc8Fvr6T_-LP7HIMx5jpbhKms--BZ_-LP7HSW-xU9_TGoh8gxK_github-status

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,13 @@
+on: push
+name: packtracker.io
+jobs:
+  report:
+    name: report webpack stats
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


bbecquet:
Meh, let's play with that later -_-

PR #486, opened 2019-12-10
🔥 Remove doTjs 🔥

bbecquet:
## Description
First step of cleaning after the last feature panel has been migrated to React.
For now, let's just remove the doT.js dependency and some util classes that are now obsolete.

The next step will be to implement `app_panel` as a real React component, but this will need more work.

## Why
Since https://github.com/QwantResearch/erdapfel/pull/479 has been merged, there is no more panel implemented with doT.js in the codebase.


views/index.ejs
```diff
@@ -33,8 +33,13 @@
       (k,v) => k==='server' ? undefined : v)};` %>
   </script>
     <div id="scene_container" class="map_container"></div>
-
-    <div id="panels"></div>
+    <div class="side_panel__container">
```
bbecquet: This mark-up comes directly from the old `app_panel.dot`. I didn't want to risk breaking things so I didn't change it but it could be simplified later. The ideal target would be to have a single `react-root` node, or something like that.

PR #489, opened 2019-12-11
QMAPS-1156 show address before GPS coordinates, if available

xem:
## Description
Show "Close to [address]" before showing GPS coordinates in POI-anywhere's
But keep GPS coords in bold if no address / nearest city is available

## Why
Address is more useful than GPS coords

## Screenshots
image

image

### DIFF
--- a/src/panel/poi/PoiHeader.jsx
+++ b/src/panel/poi/PoiHeader.jsx
@@ -1,3 +1,4 @@
+/* global _ */
 import React from 'react';
 import ReviewScore from 'src/components/ReviewScore';
 import poiSubClass from 'src/mapbox/poi_subclass';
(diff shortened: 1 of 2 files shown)


src/panel/poi/PoiHeader.jsx
```diff
@@ -8,15 +9,30 @@ const PoiHeader = ({ poi }) => {
...
-    </h4>
-    {subClassName && <p className="poi_panel__description">{poiSubClass(subClassName)}</p>}
-    {address && address.label && <p className="poi_panel__address">{address.label}</p>}
-    {grades && <ReviewScore reviews={grades} poi={poi} />}
+      <h4 className="poi_panel__title">
+        <span className="poi_panel__title__main">
+          {address && address.label ? address.label : title}
+        </span>
+      </h4>
+      {<p className="poi_panel__address">{address && address.label ? title : ''}</p>}
```
bbecquet: Do we really need this node when it's empty? The previous impl. didn't add it at all (but maybe it's needed now for layout reasons).

PR #491, opened 2019-12-12
Simplify programming style for POI

bbecquet:
*:information_source: probably best reviewed commit-by-commit*

## Description
First step to move from the complex class hierarchy of POIs (POI, IdunnPoi, BragiPoi, LatLonPoi, etc.) to a simpler style inspired by functional programming.
The future goal is to manipulate POIS as plain immutable JS objects, with pure functions to manage them.

Here we simplify the different POI classes starting with their methods:
- some useless ones are simply removed
- generic ones (ex: `toUrl`) are defined as pure functions in a new module and can be imported individually
- the ones which are specific to a single use case and high-level (ex: `logGradesClick`) are moved where it's more logical

## Why
- Overall simplification of structures
- No need to test existence of a specfici function on POI instances anymore
- No need to explicitely serialize/deserialize POIs when storing them in the history or local storage
- Easier to unit test: pure functions to manipulate POI struct will be straightforward to test with clear I/O
- etc.

### DIFF
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -18,106 +18,64 @@ function roundWithPrecision(value, precision) {
 
 export default class BragiPoi extends Poi {
   constructor(feature, queryContext) {
+    const geocodingProps = feature.properties.geocoding;
+    const { id, type, label, address, city, administrative_regions } = geocodingProps;
 
     let poiClassText = '';
     let poiSubclassText = '';
 
-    if (feature.properties.geocoding.properties &&
-        feature.properties.geocoding.properties.length > 0) {
-      const poiClass = feature.properties.geocoding.properties.find(property => {
-        return property.key === 'poi_class';
-      });
-
+    if (geocodingProps.properties && geocodingProps.properties.length > 0) {
+      const poiClass = geocodingProps.properties.find(property => property.key === 'poi_class');
       if (poiClass) {
         poiClassText = poiClass.value;
       }
-      const poiSubclass = feature.properties.geocoding.properties.find(property => {
-        return property.key === 'poi_subclass';
-      });
+      const poiSubclass = geocodingProps.properties.find(property =>
+        property.key === 'poi_subclass');
       if (poiSubclass) {
         poiSubclassText = poiSubclass.value;
       }
-    }
-    let addressLabel = '';
-    if (feature.properties &&
-        feature.properties.geocoding &&
-        feature.properties.geocoding.address) {
-      addressLabel = feature.properties.geocoding.address.label;
     }
 
     /* generate name corresponding to poi type */
     let name = '';
     let alternativeName = '';
-    const adminLabel = '';
-    const resultType = feature.properties.geocoding.type;
 
-    let postcode;
-    if (feature.properties.geocoding.postcode) {
-      postcode = feature.properties.geocoding.postcode.split(';')[0];
-    }
-    const city = feature.properties.geocoding.city;
-    const country = feature.properties.geocoding.administrative_regions.find(administrativeRegion =>
-      administrativeRegion.zone_type === 'country'
-    );
-    const cityObj = feature.properties.geocoding.administrative_regions.find(administrativeRegion =>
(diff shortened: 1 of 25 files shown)


src/adapters/poi/bragi_poi.js
```diff
@@ -18,106 +18,64 @@ function roundWithPrecision(value, precision) {
...
 
-    switch (resultType) {
+    switch (type) {
     case 'poi':
-      name = feature.properties.geocoding.name;
-      alternativeName = addressLabel || cityObj && cityObj.label ||
-        [postcode, city, countryName].filter(zone => zone).join(', ');
+      name = geocodingProps.name;
+      alternativeName = address && address.label
+        || cityObj && cityObj.label
```
GuillaumeGomez: I know we have a lint about this but I really wonder if keeping the parens wouldn't be better in the end... The formatting you used makes it easier to read (and thanks for that) but parens would help me read the code. What do you think about it?

src/libs/pois.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const latLonUrlRegex = /^latlon:(-?\d*\.\d*):(-?\d*\.\d*)(?:@(.*))?/;
+  let urlData = urlParam.match(latLonUrlRegex);
+  if (urlData) {
+    const [ _, lat, lng, label ] = urlData;
+    const latLng = { lat: parseFloat(lat), lng: parseFloat(lng) };
+    return Promise.resolve(
+      new LatLonPoi(latLng, label ? ExtendedString.htmlEncode(label) : null)
+    );
+  } else {
```
GuillaumeGomez: Since you return in the `if` block, wouldn't remove the `else` clause better? I have a personal preference over less indented code but up to you!

src/libs/pois.js
```diff
@@ -0,0 +1,71 @@
...
+
+const ZOOM_BY_POI_TYPES = {
+  street: 17,
+  house: 19,
+  poi: 18,
+};
+const DEFAULT_ZOOM = 16;
+
+export function getBestZoom(poi) {
+  return ZOOM_BY_POI_TYPES[poi.type] || DEFAULT_ZOOM;
```
GuillaumeGomez: :+1:

src/adapters/poi/bragi_poi.js
bbecquet: I'm not a fan either of this rule, but moreover I'd like to avoid this kind of big boolean expression :) On the next pass I'll try to simplify these rules when parsing Idunn and Bragi POIs, so probably it won't stay long.
For the linting rule, don't hesitate to remove it. Or maybe it can be only relaxed https://eslint.org/docs/rules/no-extra-parens#nestedbinaryexpressions

src/libs/pois.js
bbecquet: I prefer it too, I'll remove the `else` block.

src/adapters/poi/bragi_poi.js
GuillaumeGomez: I'll send a PR once yours is merged then.

tests/integration.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
   globals: {
     puppeteerArguments: ['--no-sandbox', '--disable-setuid-sandbox', '--disable-dev-shm-usage'],
     APP_URL: 'http://localhost:3000/maps',
+    __config: require('@qwant/nconf-builder').get_without_check(),
```
amatissart: Is that not a bit confusing to define a configuration object here, different from the overridden configuration defined in /tests/integration/server_start.js ?

bbecquet: Ah yes, maybe, I wasn't sure about that…
The thing is, because of changes in how modules are imported, integration tests are now broken if there is no `__config` global variable (it breaks [here](https://github.com/QwantResearch/erdapfel/blob/34c34eb14cb29aa92c96f38a9d29acc8f6f64fb0/local_modules/nconf_getter/index.js#L2), when `nconf_getter` is imported by `idunn_poi`).
Maybe there is a way in `server_start` to inject the resulting config object as a global variable for the client?

PR #494, opened 2019-12-18
Add more info when someone arrives on qwant maps

GuillaumeGomez:
This PR adds more info when someone arrives on Qwant maps.

I also added the screen sizes, I think it could be interesting for mobile devices which might have a very wide range of possible sizes and help us give priorities to testing.

### DIFF
--- a/src/panel/app_panel.js
+++ b/src/panel/app_panel.js
@@ -18,6 +18,7 @@ import ReactPanelWrapper from 'src/panel/reactPanelWrapper';
 import events from 'config/events.yml';
 import { parseMapHash, parseQueryString, joinPath, getCurrentUrl } from 'src/libs/url_utils';
 import DirectionPanel from 'src/panel/direction/DirectionPanel';
+import Device from '../libs/device';
 
 const performanceEnabled = nconf.get().performance.enabled;
 const directionEnabled = nconf.get().direction.enabled;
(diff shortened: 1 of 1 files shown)


src/panel/app_panel.js
```diff
@@ -55,7 +55,16 @@ export default class AppPanel {
     }
 
     ReactDOM.render(<Menu />, document.querySelector('.react_menu__container'));
-    Telemetry.add(Telemetry.APP_START);
+    Telemetry.add(Telemetry.APP_START, null, null, {
+      'language': window.navigator.language,
+      'preferred_language': window.preferredLanguage,
```
bbecquet: It's `window.preferedLanguage` (one 'r').

src/panel/app_panel.js
```diff
@@ -55,7 +55,16 @@ export default class AppPanel {
...
+      'device_platform': window.navigator.platform,
```
bbecquet: I think these two are really borderline in terms of privacy…

src/panel/app_panel.js
```diff
@@ -55,7 +55,16 @@ export default class AppPanel {
...
+      'device_user_agent': window.navigator.userAgent,
+      'window_height': window.innerHeight,
```
bbecquet: Meh, if we really need info about app display size, maybe we can just add two of them, because I doubt we'll ever do something different with device vs. window size.
Alternative: just sending the result of `Device.isMobile()`

src/panel/app_panel.js
GuillaumeGomez: It's part of the information listed in the ticket. :-/

GuillaumeGomez:
Greatly reduced the number of returned information.

PR #503, opened 2020-01-03
Add "suggestion" in case there is no results to a query

GuillaumeGomez:
In case a user enters a search query with no results, it'll now show a little something instead of nothing.

I didn't find a sketch for this so the rendering is very "basic" but at least the logic is here.

Screenshot from 2020-01-03 16-03-12

### DIFF
--- a/src/adapters/suggest.js
+++ b/src/adapters/suggest.js
@@ -107,7 +107,7 @@ export default class Suggest {
 
         // fill the suggest with the remotes poi according to the remaining places
         const remotesLen = SUGGEST_MAX_ITEMS - nbFavorites - categories.length;
-        suggestDom += this.remotesRender(remotes.slice(0, remotesLen));
+        suggestDom += this.remotesRender(remotes.slice(0, remotesLen), query);
 
         if (favorites.length > 0) {
           suggestDom += this.favoritesRender(favorites.slice(0, query === '' ? 5 : nbFavorites));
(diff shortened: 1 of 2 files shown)


src/adapters/suggest.js
```diff
@@ -214,6 +217,23 @@ export default class Suggest {
...
   }
 
+  renderEmpty(query = '') {
+    if (query.trim() === '') {
+      return '';
+    }
+    return `
+      <div class="no_suggestion">
+        <div class="first_line">
+          ${_(`Impossible to find "${query}" in Qwant Maps...`)}
```
amatissart: This templated message is incorrect: the translation key should not depend on the "query".  
See https://github.com/QwantResearch/erdapfel/blob/master/src/components/OpeningHour.jsx#L40 for an example

src/adapters/suggest.js
```diff
@@ -214,6 +217,23 @@ export default class Suggest {
...
+        </div>
+        <div class="second_line">
+          ${_('You can add this place in OpenStreetMap so you can find it in Qwant Maps')}
+          <a href="https://learnosm.org/en/beginner/start-osm/">${_('here')}</a>.
```
amatissart: I am not convinced that this wording is appropriate. In many cases the query may refer to a place that does exist in OSM but is not indexed in our database, or under a different name/spelling, etc. Also a query may refer to a larger intention and not to a specific place.  

In my opinion this message should be less insistent.

bbecquet:
The discussion for how we deal with this case functionally is still ongoing (see QMAPS-1232). 
Moreover, the technical implementation of the suggestions changed a lot since this PR, with now a full React version, so it would be faster to just rewrite what this PR proposes than trying to merge it.

PR #515, opened 2020-01-13
Use Mapbox-GL native LLBounds.contains()

bbecquet:
## Description
Uses the [LngLatBounds.contains](https://docs.mapbox.com/mapbox-gl-js/api/#lnglatbounds#contains) method instead of custom impl.

## Why
Follow up to https://github.com/QwantResearch/erdapfel/pull/513, this method was introduced in [MapBox 1.6.0](https://github.com/mapbox/mapbox-gl-js/releases/tag/v1.6.0)

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -195,60 +195,27 @@ Scene.prototype.restoreLocation = function() {
   }
 };
 
-Scene.prototype.isBBoxInExtendedViewport = function(bbox) {
+const clamp = (min, max, value) => Math.min(max, Math.max(min, value));
 
-  // Get viewport bounds
+Scene.prototype.isBBoxInExtendedViewport = function(bbox) {
   const viewport = this.mb.getBounds();
 
-  // Compute "width", "height"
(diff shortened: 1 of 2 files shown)


bbecquet:
Putting in **WIP** because the failing test is due to our mapbox-gl-mock having too limited bounds implementation…
There are multiple possibilities:
- we can just comment out the failing test for now
- we can modify our mock fork so bounds work as they should (medium easy)
- we can finish ongoing work to ditch the mock (medium hard)

bbecquet:
@amatissart I decided to comment out the failing test, for two reasons:
- modifying the mock to have a full LLBounds implementation was too much work, with too many side effects and a dubious goal (where do we stop in mocking a map? :))
- this particular test was already unreliable in the previous implementation, testing that the new center was `{ lat: 4, lng: 3 }` whereas the `bbox` in the mock data was `[1, 2, 3, 4]`. The test was clearly written based on wrong I/O assumptions, it was just what the test suite with the mock produced.

PR #516, opened 2020-01-14
[Refacto] React root app 

bbecquet:
## Description
Some heavy changes to `app_panel`, on the road to have a fully React-ified app:
- put everything related to panel switching in a dedicated component `<PanelManager>`, which manages the currently active panel in its state
- just mount a React root component once and manage everything below with the React render cycle (as added benefit, this means we can get rid of our bridge util `reactPanelWrapper` and there is no more flashing when components update, for example when going from a Poi to another)
- try to delegate responsibilities to the right panels where it's possible
- fix the `<Information>` component, a part of the POI panel, which was breaking the React rule of not storing stuff derived from props outside of the component state, resulting in parts not re-rendering.

## Why
- Better separation of concerns
- Getting rid of boilerplate, simpler app init code, cleaner situation if we decide to introduce React router
- Now it will be easier to manage some things as app-wide context, the first use will be to make the Device testing really dynamic


src/panel/PanelManager.jsx
```diff
@@ -0,0 +1,168 @@
...
+
+  componentDidMount() {
+    this.initRouter();
+    this.initTopBar();
+    if (performanceEnabled) {
+      window.times.appRendered = Date.now();
+    }
+  }
+
+  componentDidUpdate(_prevProps, prevState) {
```
bbecquet: This code is still hard to read as it includes our very specific rules about panel switching: auto minification of the search bar and the damned stuff about persistent category markers…

src/panel/PanelManager.jsx
```diff
@@ -0,0 +1,168 @@
...
+
+    searchInput.onfocus = () => {
+      topBarHandle.classList.add('top_bar--search_focus');
+    };
+
+    searchInput.onblur = () => {
+      topBarHandle.classList.remove('top_bar--search_focus');
+    };
+
+    const minifierButton = document.querySelector('.top_bar .minifier');
```
bbecquet: Lines above could be put elsewhere…

src/panel/PanelManager.jsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+
+  render() {
+    const { ActivePanel, options, isMinified } = this.state;
+
+    return <div className={classnames('side_panel__container', {
+      'side_panel__container--hidden': isMinified }
+    )}>
+      <div className="favorite_poi_panel__container">
```
bbecquet: On the todo list: getting rid of this old classname and dom structure.

src/panel/app_panel.js
```diff
@@ -1,137 +1,55 @@
...
-      }
-    });
-
     window.onpopstate = ({ state }) => {
       this.router.routeUrl(getCurrentUrl(), state);
     };
 
-    // Route the initial URL
-    this.router.routeUrl(getCurrentUrl());
+    ReactDOM.render(
```
bbecquet: Excepting modals, this is now the only place where we mount something explicitely in the DOM.

src/scss/includes/z_index.scss
```diff
@@ -35,7 +35,7 @@
 }
 
 .top_bar {
-  z-index: 3;
+  z-index: 2;
```
amatissart: why this change ? 
As a result, this top bar is no longer visible on mobile when detailed POI panel is open:
image

bbecquet: This was apparently the result of an outdated base branch, can't reproduce since the last rebase.

PR #517, opened 2020-01-15
QMAPS-1205 Add clicked PoIs in direction panel's fields

xem:
## Description
Enable the possibility to fill the direction panel's fields by clicking on a point of the map or on an interactive PoI
If both fields are already full, close the direction panel and open the PoI panel

## Why
Fill the direction panel's form easily without having to type the name of each point and/or selecting them in the suggest list

## Screenshots
image

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -72,7 +72,6 @@ Scene.prototype.initMapBox = function() {
 
   this.popup.init(this.mb);
 
-
   window.map = { mb: this.mb };
 
   const interactiveLayers = [
@@ -111,7 +110,6 @@ Scene.prototype.initMapBox = function() {
       // This code listens to clicks on each interactive layer, and if an interactive PoI marker is clicked, instantiate a MapPoi and save it in e._mapPoi.
       // The global click listener (below) will open the corresponding PoI panel or complete the direction panel fields with it.
       this.mb.on('click', interactiveLayer, async e => {
-        e._interactiveClick = true;
         if (e.features && e.features.length > 0) {
           e._mapPoi = new MapPoi(e.features[0]);
         }
@@ -143,19 +141,10 @@ Scene.prototype.initMapBox = function() {
         poi = new LatLonPoi(e.lngLat);
       }
 
-      // Complete Direction panel if it's open, open PoI panel if it's already full
+      // If direction panel is open, tell it to fill its empty fields with this PoI
+      // If it's closed, open PoI panel
       if (document.querySelector('.directions-open')) {
-        const originField = document.querySelector('#itinerary_input_origin');
-        if (originField.value === '') {
-          fire('change_direction_point', 'origin', poi);
-        } else {
-          const destinationField = document.querySelector('#itinerary_input_destination');
-          if (destinationField.value === '') {
-            fire('change_direction_point', 'destination', poi);
-          } else {
-            window.app.navigateTo(`/place/${toUrl(poi)}`, { poi });
-          }
-        }
+        fire('set_direction_point', poi);
       } else {
         // Otherwise show PoI panel
         window.app.navigateTo(`/place/${toUrl(poi)}`, { poi });
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -16,7 +16,6 @@ import SceneEasterEgg from './scene_easter_egg';
 import Device from '../libs/device';
 import { parseMapHash, getMapHash } from 'src/libs/url_utils';
 import { toUrl, getBestZoom } from 'src/libs/pois';
-import {originDestinationCoords} from "../libs/route_utils";
 import Error from 'src/adapters/error';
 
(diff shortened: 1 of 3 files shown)


src/adapters/scene.js
```diff
@@ -107,28 +107,59 @@ Scene.prototype.initMapBox = function() {
...
+      // Select PoI (OSM or LatLon)
+      let poi;
+      if (e._mapPoi) {
+        poi = e._mapPoi;
+      } else {
+        poi = new LatLonPoi(e.lngLat);
+      }
+
+      // Complete Direction panel if it's open, open PoI panel if it's already full
+      if (document.querySelector('.directions-open')) {
```
bbecquet: The logic looks good, but it would be better to move it where the data state of the form is maintained, i.e. the `DirectionPanel` itself.
Here you need to refer directly to DOM elements, which requires to put fine knowledge of how the UI layer works in that file that isn't related to routes.

A better solution would be to send a single type of event, with no explicit `origin` or `destination` indication, and let `DirectionPanel` decide which field to fill based only on its internal state.

Among added benefits, it would make it more robust if we add more than two points in the future.

src/adapters/scene.js
```diff
@@ -107,28 +107,59 @@ Scene.prototype.initMapBox = function() {
         this.mb.getCanvas().style.cursor = '';
       });
 
+      // The OSM PoIs are on separate interactive layers.
+      // This code listens to clicks on each interactive layer, and if an interactive PoI marker is clicked, instantiate a MapPoi and save it in e._mapPoi.
+      // The global click listener (below) will open the corresponding PoI panel or complete the direction panel fields with it.
       this.mb.on('click', interactiveLayer, async e => {
         e._interactiveClick = true;
```
bbecquet: Looks like the `_interactiveClick` is now obsolete :)

src/panel/direction/DirectionPanel.jsx
```diff
@@ -180,6 +182,31 @@ export default class DirectionPanel extends React.Component {
     this.setState({ [which]: value, isDirty: true }, this.update);
   }
 
+  setDirectionPoint = poi => {
+
+    // Special case: if both origin and destination are already set, open PoI panel
```
bbecquet: Are we sure of that? I know it's in the ticket description, but it seems a bit brutal to me. Currently we just do nothing.

src/adapters/scene.js
```diff
@@ -107,28 +106,49 @@ Scene.prototype.initMapBox = function() {
...
-      }
-      // Disable POI anywhere feature on mobile until we opt for an adapted UX
-      if (Device.isMobile() || e._interactiveClick || this.routeDisplayed) {
         return;
       }
-      const poi = new LatLonPoi(e.lngLat);
-      window.app.navigateTo(`/place/${toUrl(poi)}`, { poi });
+
+      // Select PoI (OSM or LatLon)
+      let poi;
```
bbecquet: Syntax nitpicking, but it would be shorter and not less readable written as:
```js
const poi = e._mapPoi || new LatLonPoi(e.lngLat);
```

bbecquet:
Tests are failing apparently because the event listeners in `scene.js` are changed. The one attached on interactive layers doesn't do anything anymore, apart from attaching the clicked feature to the event object, letting the generic click-on-map listener handling all the rest.
This is better, making code paths easier to follow, but the mock doesn't support this. It is only able to react to a fake click on an interactive layer, and doesn't trigger the generic 'click' event after that :/

src/adapters/scene.js
```diff
@@ -107,28 +106,49 @@ Scene.prototype.initMapBox = function() {
         this.mb.getCanvas().style.cursor = '';
       });
 
+      // The OSM PoIs are on separate interactive layers.
```
bbecquet: After fiddling with tests, I think we could go one step further with this simplification, to get rid of this `_mapPoi` hack.

It makes both listeners (the interactive one and the generic one) "communicate", but it is far too fragile. It relies on the assumption that it's the same event object used by the MapBox-GL implementation to trigger both listenernes, but nothing guarantees that and it could break in the future.

By leveraging `queryRenderedFeatures`, it's possible to do everything in a single, generic listener, making everything simpler and more solid.

```js
this.mb.on('click', e => {
      const features = this.mb.queryRenderedFeatures(e.point, { layers: interactiveLayers });

      const poi = features.length > 0 ? new MapPoi(features[0]) : new LatLonPoi(e.lngLat);

      // do stuff with the POI
```

As this is called only once in a while, I don't think there are performance issues to use `queryRenderedFeatures` instead of direct listeners on the layer themselves.

Sadly this doesn't fix the tests yet, but at least we would have only one event to fake, not two.
I think this could be done in a separate PR, to be merged before this one. We would address this test problem in it. A really simple solution for that is to call a method `clickOnPoi()` instead of doing everything directly in the event listener, which requires faking events, which are badly supported by the mock.

What do you think? @xem @amatissart

xem:
closed. see PR 523

PR #499, opened 2019-12-30
Clean and simplify POI panel structure

bbecquet:
## Description
Simplify stuff about the POI Panel, trying to reduce its complexity:
- regroup component files in the same directory
- use the `<Panel>` component, when it's not in compact mode (aka. "card")
- remove a lot of useless/specific CSS
- slightly improve some UX details in the process, like the close button on the mobile card which is consistently present and never collides with the action buttons anymore (fixing a part of QMAPS-1066)

## Why
- Improve style cohesion everywhere: now all panels use the standard `<Panel>` component, ensuring visual changes to it are applied everywhere
- Simplify code readability and maintenability

PR #520, opened 2020-01-17
Manage Device as context to re-render dynamically

bbecquet:
## Description
Make the app react dynamically to device size changes ("desktop" <-> "mobile"), instead of requiring a reload for cases when we used `Device.isMobile()` in render code.
We first watch the mobile media query in JS, and update a state on the root app accordingly. We can then use the [React context mechanism](https://reactjs.org/docs/context.html) to transmit this value down to any component that would like to use it, without having to pass it explicitly down the component tree.

## Why
Fix the routes panel broken UI when the window/device size is changing from desktop size to mobile size, and vice-versa.
Now automatically re-render to the appropriate UI version.

## Screenshots
|Before|After|
|---|---|
|Peek 17-01-2020 17-29|Peek 17-01-2020 17-15|


bbecquet:
Note: if https://github.com/QwantResearch/erdapfel/pull/499 gets merged before, I'll modify this PR so the POI panel uses the Device context too.

bbecquet:
WIP to add support for the POI panel, with switching between card and full panel display.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -213,25 +213,29 @@ export default class DirectionPanel extends React.Component {
...
-
-    return <Panel
-      title={title}
-      close={this.onClose}
-      className="direction_panel"
-    >
-      {form}
-      {result}
-    </Panel>;
+    return <DeviceContext.Consumer>
```
amatissart: Probably a matter of taste, but don't you think the new  [`useContext()`](https://fr.reactjs.org/docs/hooks-reference.html#usecontext) is more readable ?

G-Ray: I agree, but following a discussion with @bbecquet , this would imply a migration of the complete class to a function component

src/panel/PanelManager.jsx
```diff
@@ -159,12 +168,14 @@ export default class PanelManager extends React.Component {
...
-    const { ActivePanel, options, isMinified } = this.state;
-
-    return <div className={classnames('panel_container',
-      { 'panel_container--hidden': isMinified }
-    )}>
-      <ActivePanel {...options} />
-    </div>;
+    const { ActivePanel, options, isMinified, isMobile } = this.state;
+
+    return <DeviceContext.Provider value={isMobile}>
```
G-Ray: Would it be possible to move this provider higher in the component tree ? This way we would only need to implement this logic only once.

src/panel/direction/DirectionPanel.jsx
amatissart: Ah ok, I thought it could be used directly in the `render()` function. Never mind.

src/panel/PanelManager.jsx
bbecquet: I can move it one level higher, but this only wouldn't solve the fundamental problem, which is that non-React code (from the map) needs to know if we are in mobile or not, so we need to make it available outside of the context mechanism.
But I'll try something, it's true that the logic is duplicated.

bbecquet: @G-Ray @amatissart In https://github.com/QwantResearch/erdapfel/pull/520/commits/1d0a6a8c11596507a187418341add31066eb11ba I've refactored the device test, so it's not duplicated and use the same shared media query.
I think we can't go further, as some non-React components will still need to access this information outside of the React Context.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -213,25 +213,29 @@ export default class DirectionPanel extends React.Component {
...
+      {isMobile => {
+        if (isMobile) {
```
G-Ray: I know you did not edited these lines in this PR, but in order to be more functional we could write :

```js
{isMobile => (isMobile
  ? <FirstComp />
  : <SecondComp />
)}
```

src/libs/device.js
```diff
@@ -1,10 +1,15 @@
...
+import React from 'react';
+
+const mediaQuery = window.matchMedia('(max-width: 640px)');
+
+export function isMobileDevice() {
+  return mediaQuery.matches;
 }
+
+export function onDeviceSizeChange(callback /* isMobile => {…} */ ) {
+  return mediaQuery.addListener(mediaQueryEvent => {
```
G-Ray: We need to removeListener once the component is unmounted to avoid memory leak

bbecquet: I voluntarily omitted it as the only case when PanelManager is unmounted is the app being exited. But maybe it's safer to do it anyway :man_shrugging:

bbecquet: Did it. I didn't see the point of an abstraction anymore, so I exposed the media query itself.

PR #521, opened 2020-01-22
Add Windows documentation

xem:
## Description
Guide for contributing to Erdapfel from Windows (10)

## Why
Not everyone has Linux 

## Screenshots
image

### DIFF
--- a/docs/src/windows.md
+++ b/docs/src/windows.md
@@ -1,0 +1,106 @@
+# Install and contribute to Erdapfel on Windows
+
+## Windows 7 / 8
+
+- Install React Dev Tools in your browser ([chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) / [Firefox](https://addons.mozilla.org/fr/firefox/addon/react-devtools/))
+- Install [VirtualBox](https://www.virtualbox.org/)
+- Mount [this xubuntu image](http://cdimages.ubuntu.com/xubuntu/releases/18.04/release/xubuntu-18.04.1-desktop-amd64.iso)
+- Allocate at least 2 CPU, 4GB of RAM and 20GB of HDD and material acceleration to the VM
+- In settings > Network, choose Bridge Adapter
+- Inside the VM, install an IDE (enable ESlint plugin), open a terminal
+- Clone erdapfel in the folder of your choice (ex: "/home/{username}/www/"): 
+```
+sudo apt install git
+git clone https://github.com/QwantResearch/erdapfel.git
+```
+- Enter the [commands](#commands) below to build and run Erdapfel
+- Retrieve your IP with ```ifconfig``` (just after "inet ...")
+<br>Your local instance of Erdapfel will be visible at **http://localhost:3000** in the VM or **http://{ip}:3000** outside of the VM
+
+# Windows 10
+
+- Install React Dev Tools in your browser ([chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) / [Firefox](https://addons.mozilla.org/fr/firefox/addon/react-devtools/))
+- Install [NodeJS LTS](https://nodejs.org/en/download/)
+- Install [Git](https://git-scm.com/downloads)
+- Install [Python](https://www.python.org/downloads/)
+- Install an IDE (enable ESlint plugin)
+- From an Admin account, enable developer mode (Start menu > Settings > Updates & Security > For developers > check "Developer mode")
+<br>And enable WSL (Settings > Apps > Programs and features > Turn Windows features on or off > enable "Windows Subsystem for Linux")
+- Then from any account, install [Ubuntu from Microsoft Store](https://www.microsoft.com/fr-fr/p/ubuntu/9nblggh4msv6?activetab=pivot:overviewtab)
+- Clone Erdapfel in the folder of your choice (ex: "C:\www\") using [Github Desktop](https://desktop.github.com/) or the terminal: 
+```
+sudo apt install git
+git clone https://github.com/QwantResearch/erdapfel.git
+```
+
+- Open a Ubuntu bash in this folder (shift + right-click > Open Linux CLI here)
+- Install tools, libraries and Python installer:
+```
+sudo apt update && sudo apt install -y build-essential git libexpat1-dev libssl-dev zlib1g-dev \
+libncurses5-dev libbz2-dev liblzma-dev \
+libsqlite3-dev libffi-dev tcl-dev linux-headers-generic libgdbm-dev \
+libreadline-dev tk tk-dev
+curl https://pyenv.run | bash
+```
+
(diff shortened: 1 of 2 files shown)


docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
...
+```
+export PATH="/home/{USERNAME}/.pyenv/bin:$PATH"
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+```
+
+- Install Python and its libraries:
+```
+pyenv install 3.6.8
+pyenv install 3.7.2
```
amatissart: Out of curiosity, why do you need two versions of Python ?

xem: it may not be mandatory:
the thing is, I had all the difficulties in the world to make WSL (linux in windows) recognize the command "python". I tried many solutions online, and this one finally paid: https://gist.github.com/monkut/35c2ef098b871144b49f3f9979032cee

(for some reason they install 2 python versions, so I kept it, but we can remove it if you think it's ok)

amatissart: I think the linked solution describes how it's possible to switch between python versions. As far as I know, our install does not depend on a specific version of Python. So, either of them would probably be enough.

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
+# Install and contribute to Erdapfel on Windows
+
+## Windows 7 / 8
+
+- Install React Dev Tools in your browser ([chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) / [Firefox](https://addons.mozilla.org/fr/firefox/addon/react-devtools/))
```
bbecquet: I think we should keep the doc straightforward, and limit it to installing and running the project on Windows.

As React Dev tools aren't related to this topic (and aren't even mandatory to develop), we shouldn't mention them here.
It could be added in a `Recommended tools` section, in another, platform-agnostic doc related to contribution.

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
...
+- Install [VirtualBox](https://www.virtualbox.org/)
+- Mount [this xubuntu image](http://cdimages.ubuntu.com/xubuntu/releases/18.04/release/xubuntu-18.04.1-desktop-amd64.iso)
+- Allocate at least 2 CPU, 4GB of RAM and 20GB of HDD and material acceleration to the VM
+- In settings > Network, choose Bridge Adapter
+- Inside the VM, install an IDE (enable ESlint plugin), open a terminal
```
bbecquet: Same remark for the IDE, this is off-topic. If someone wants to contribute with Notepad and no realtime ESLint hint, technically they can :)

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
+# Install and contribute to Erdapfel on Windows
+
+## Windows 7 / 8
```
bbecquet: As Win7 is old and not supported anymore, and Win8 is quasi non-existant, I would consider Windows 10 the default case and put it first. Especially since it doesn't require full Linux virtualization.

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
...
+sudo apt install git
+git clone https://github.com/QwantResearch/erdapfel.git
+```
+- Enter the [commands](#commands) below to build and run Erdapfel
+- Retrieve your IP with ```ifconfig``` (just after "inet ...")
+<br>Your local instance of Erdapfel will be visible at **http://localhost:3000** in the VM or **http://{ip}:3000** outside of the VM
+
+# Windows 10
+
+- Install React Dev Tools in your browser ([chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) / [Firefox](https://addons.mozilla.org/fr/firefox/addon/react-devtools/))
```
bbecquet: Same remark here for React Dev Tools.

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
...
+- Install [NodeJS LTS](https://nodejs.org/en/download/)
+- Install [Git](https://git-scm.com/downloads)
+- Install [Python](https://www.python.org/downloads/)
+- Install an IDE (enable ESlint plugin)
```
bbecquet: … and here for IDE. It can be removed.

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
...
+```
+- Enter the [commands](#commands) below to build and run Erdapfel
+- Retrieve your IP with ```ifconfig``` (just after "inet ...")
+<br>Your local instance of Erdapfel will be visible at **http://localhost:3000** in the VM or **http://{ip}:3000** outside of the VM
+
+# Windows 10
+
+- Install React Dev Tools in your browser ([chrome](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi) / [Firefox](https://addons.mozilla.org/fr/firefox/addon/react-devtools/))
+- Install [NodeJS LTS](https://nodejs.org/en/download/)
+- Install [Git](https://git-scm.com/downloads)
```
bbecquet: Is it required, as we install it via APT or Github desktop later?

docs/src/windows.md
```diff
@@ -0,0 +1,106 @@
...
+# Commands
+
+- Install tools (you only need to do it once):
+```
+sudo apt install python3
+sudo apt install curl
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.2/install.sh | bash # see https://github.com/creationix/nvm#install-script
+source ~/.bashrc
+nvm install 8.9.0
+npm i webpack -g
```
bbecquet: `webpack` and `webpack-cli` are listed in the local packages of Erdapfel, I don't think you need to install them globally (but maybe I'm overlooking smth)

docs/src/windows.md
xem: Agreed: I'm commenting this section altogether. It won't be visible in the help page, but still present in the doc's source code just in case

docs/src/windows.md
xem: idk: these commands were in the previous install guide I made when I joined the project :D

docs/src/windows.md
```diff
@@ -44,26 +24,45 @@ curl https://pyenv.run | bash
...
-pip install pip --upgrade
-pip install pipenv
 pyenv global 3.7.2
 pip install pip --upgrade
 pip install pipenv
 ```
 
 - Then enter the [commands](#commands) below to build and run Erdapfel
 
+<!--
```
bbecquet: Instead of commenting it, which adds useless lines of doc to maintain, I would just explain briefly that Windows 7/8 isn't "officially" supported and not recommended, and if you *really* want to run Erdapfel on it you'll need a Linux VM. No need to dive into VM config details.

It could be nice hovewer to explain, for example at the beginning, why Erdapfel doesn't work natively on Windows and always requires Linux virtualization…

xem: sure

PR #523, opened 2020-01-27
QMAPS-1205 handle POI click when direction panel is open

xem:
## Description
Enable the possibility to fill the direction panel's fields by clicking on a point of the map or on an interactive PoI
If both fields are already full, close the direction panel and open the PoI panel

## Why
Fill the direction panel's form easily without having to type the name of each point and/or selecting them in the suggest list

## Screenshots
image


src/libs/pois.js
```diff
@@ -18,7 +17,7 @@ export function toAbsoluteUrl(poi) {
   const baseUrl = window.baseUrl;
   const lat = poi.latLon.lat.toFixed(7);
   const lon = poi.latLon.lng.toFixed(7);
-  const mapHash = `#map=${getBestZoom(poi.zoom)}/${lat}/${lon}`;
+  const mapHash = `#map=${getBestZoom(poi)}/${lat}/${lon}`;
```
bbecquet: Next time, let's put that kind of bug fix in a separate PR, even if it's only one line.

xem: np!

src/adapters/scene.js
```diff
@@ -160,22 +160,17 @@ Scene.prototype.initMapBox = function() {
 };
 
 Scene.prototype.clickOnMap = function(lngLat, clickedFeature) {
-  // Disable POI anywhere feature on mobile until we opt for an adapted UX
-  if (isMobileDevice() && !clickedFeature) {
```
bbecquet: I think you removed too much stuff here during the rebase. We still want specific behavior on mobile when not in direction mode, when clicking anywhere on the map.

PR #527, opened 2020-01-30
Use qwant-maps-fonts module

G-Ray:
## Description
Remove current build scripts to build fonts, and pull https://www.npmjs.com/package/qwant-maps-fonts/ instead

## Why
:rocket: increase npm install speed
Node.js :one: :two: support


package.json
```diff
@@ -55,16 +54,16 @@
     "jest": "^24.1.0",
     "jest-extended": "^0.11.2",
     "json-loader": "0.5.7",
-    "klokantech-noto-sans": "git+https://github.com/QwantResearch/fonts.git#f25a802",
     "mapbox-gl": "^1.6.1",
     "mapbox-gl-js-mock": "https://github.com/QwantResearch/mapbox-gl-js-mock.git#7eaee7b",
     "masq-lib": "git+https://github.com/QwantResearch/masq-lib.git#v0.14.3",
-    "node-sass": "^4.9.2",
+    "node-sass": "^4.13.1",
```
bbecquet: Is this related to the font stuff?

G-Ray: In order to fix compat with node >= 12, I changed this in a separate commit that the reset of the changes

PR #531, opened 2020-02-05
QMAPS-1189 show three different error messages in directions panel

xem:
## Description
Show more specific erroe messages in Directions

## Why
To know if the service is down temporarily, or if there's really no result

## Screenshots

image

image

image

NB: one sentence is already translated, hence the bilingual screenshots

### DIFF
--- a/src/adapters/direction_api.js
+++ b/src/adapters/direction_api.js
@@ -93,6 +93,11 @@ export default class DirectionApi {
       return { data: response.data, error: 0 };
     }
 
+    // Mock response / direction not proxified by qwant
+    if (directionConfig.api === 'mapbox') {
+      return { data: response, error: 0 };
+    }
+
     // Empty response
     return { data: null, error: -1 };
   }
--- a/src/adapters/direction_api.js
+++ b/src/adapters/direction_api.js
@@ -74,7 +74,13 @@ export default class DirectionApi {
     try {
       response = await Ajax.get(directionsUrl, directionsParams, { timeout });
     } catch (e) {
-      response = e;
+      if (e >= 400 && e < 600) {
+        // Use the error codes 4xx and 5xx to display different error messages
+        response = { data: null, error: true, errorCode: e };
+      } else {
+        // Other errors are catched silently
+        return;
+      }
     }
     if (directionConfig.api === 'qwant' && response.data) {
       response = response.data;
--- a/src/adapters/direction_api.js
+++ b/src/adapters/direction_api.js
@@ -74,16 +74,16 @@ export default class DirectionApi {
     try {
       response = await Ajax.get(directionsUrl, directionsParams, { timeout });
     } catch (e) {
-      if (e >= 400 && e < 600) {
+      if (Number.isInteger(e) && e >= 400 && e < 600) {
         // Use the error codes 4xx and 5xx to display different error messages
-        response = { data: null, error: true, errorCode: e };
+        response = { data: null, error: e };
       } else {
-        // Other errors are catched silently
-        return;
+        // Other errors
+        response = { data: null, error: -1 };
       }
     }
     if (directionConfig.api === 'qwant' && response.data) {
-      response = response.data;
+      response = { data: response.data, error: 0 };
     }
     return response;
   }
--- a/src/adapters/direction_api.js
+++ b/src/adapters/direction_api.js
@@ -76,16 +76,25 @@ export default class DirectionApi {
     } catch (e) {
       if (Number.isInteger(e) && e >= 400 && e < 600) {
         // Use the error codes 4xx and 5xx to display different error messages
-        response = { data: null, error: e };
+        return { data: null, error: e };
       } else {
         // Other errors
-        response = { data: null, error: -1 };
+        return { data: null, error: -1 };
       }
     }
-    if (directionConfig.api === 'qwant' && response.data) {
-      response = { data: response.data, error: 0 };
+
+    // Valid response
+    if (
+      directionConfig.api === 'qwant'
+      && response.data
+      && response.data.routes
+      && response.data.routes.length > 0
+    ) {
+      return { data: response.data, error: 0 };
     }
-    return response;
+
+    // Empty response
+    return { data: null, error: -1 };
   }
 }
 
--- a/src/adapters/direction_api.js
+++ b/src/adapters/direction_api.js
@@ -74,9 +74,9 @@ export default class DirectionApi {
     try {
       response = await Ajax.get(directionsUrl, directionsParams, { timeout });
     } catch (e) {
-      return;
+      response = e;
     }
-    if (directionConfig.api === 'qwant') {
+    if (directionConfig.api === 'qwant' && response.data) {
       response = response.data;
     }
     return response;
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -56,7 +56,7 @@ export default class DirectionPanel extends React.Component {
         (props.poi && Poi.deserialize(props.poi)) || persistentPointState.destination || null,
       isLoading: false,
       isDirty: true, // useful to track intermediary states, when API update call is not made yet
-      error: 0,
+      error: false,
       routes: [],
       activePreviewRoute: null,
       isInitializing: true,
@@ -120,12 +120,12 @@ export default class DirectionPanel extends React.Component {
           isActive: i === 0,
           id: i,
         }));
-        this.setState({ isLoading: false, error: 0, routes });
+        this.setState({ isLoading: false, error: false, routes });
         window.execOnMapLoaded(() => {
           fire('set_route', { routes, vehicle, origin, destination });
         });
       } else {
-        this.setState({ isLoading: true, error: directionResponse });
+        this.setState({ isLoading: true, error: true, errorCode: directionResponse });
         fire('clean_route');
       }
     } else {
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -114,22 +114,26 @@ export default class DirectionPanel extends React.Component {
         destination,
         vehicle,
       );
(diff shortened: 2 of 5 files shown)


src/panel/direction/RouteResult.jsx
```diff
@@ -73,7 +73,20 @@ export default class RouteResult extends React.Component {
     if (this.props.error) {
       return <div className="itinerary_no-result">
         <span className="icon-alert-triangle" />
-        <div>{_('Qwant Maps found no results for this itinerary.')}</div>
+        <div>{
+          this.props.error >= 500 && this.props.error < 600
+            ? _('The service is temporarily unavailable, please try again later.', 'direction')
+            : this.props.error === 422
+              ? _('Qwant Maps found no results for this itinerary.', 'direction')
+              : ''
```
amatissart: What if the `directionResponse` contains an empty list of routes ?   
As far as I understand, `this.props.error` will contain the response object, and the error message will no longer be visible.

amatissart: Here is the result
image

With a React warning
```
Warning: Failed prop type: Invalid prop `error` of type `object` supplied to `RouteResult`, expected `number`.
    in RouteResult (created by DirectionPanel)
    in DirectionPanel (created by PanelManager)
    in div (created by PanelManager)
    in PanelManager
```

bbecquet: I agree there is an inconsistency for empty results.

Maybe we can change returns from direction_api so error cases are easier to parse and we don't need to make assumptions such as `if (directionResponse && directionResponse.routes && directionResponse.routes.length > 0)` to separate valid results from errors in the Panel.

For example, always return an object, with an optional `error` field, and never return the error itself. Even if the API returns "correctly" but with an empty list of routes, fill the error property so it falls in a generic "no route found" case.

xem: I agree, I lacked a bit of global knowledge of this part of erdapfel to address this issue in the best possible way. I also thinked about including the optional error code in the same object that contains the data if there's no error... I'll try to make this the right way, and will let you check if it's ok :)

src/panel/direction/RouteVia.jsx
```diff
@@ -7,22 +7,24 @@ const RouteVia = ({ route, vehicle }) => {
     return <div className="routeVia">
       {_('Via', 'direction')} {route.legs[0].summary.replace(/^(.*), (.*)$/, '$1')}
     </div>;
+  } else if (route.summary) {
```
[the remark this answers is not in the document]
xem: indeed, at one moment I had "route.summary is undefined" but it's probably caused by another file I had changed incorrectly

src/adapters/direction_api.js
```diff
@@ -74,9 +74,15 @@ export default class DirectionApi {
     try {
       response = await Ajax.get(directionsUrl, directionsParams, { timeout });
     } catch (e) {
-      return;
+      if (e >= 400 && e < 600) {
+        // Use the error codes 4xx and 5xx to display different error messages
+        response = { data: null, error: true, errorCode: e };
+      } else {
+        // Other errors are catched silently
+        return;
```
bbecquet: Now that we have a `error` flag available for this use, this path should use it too to avoid data flow inconsistency.

src/adapters/direction_api.js
```diff
@@ -74,9 +74,15 @@ export default class DirectionApi {
...
+      }
     }
-    if (directionConfig.api === 'qwant') {
+    if (directionConfig.api === 'qwant' && response.data) {
       response = response.data;
```
bbecquet: Now that you established that the response shape is `{ data, error, errorCode }`, it would be better if the working case used it as well.
Otherwise you have now 3 different cases as result from this function:
- in case of a known error, an object `{ data: null, error: true, errorCode: … }`.
- in case of an unknown error, `undefined`
- in case of a working result, an object `{ routes: …, message: …, code: …}`
This makes managing all the cases more complex.

src/panel/direction/RouteResult.jsx
```diff
@@ -73,7 +74,21 @@ export default class RouteResult extends React.Component {
...
+        <div>{
+          this.props.errorCode >= 500 && this.props.errorCode < 600
+            ? _('The service is temporarily unavailable, please try again later.', 'direction')
+            : _('Qwant Maps found no results for this itinerary.', 'direction')
+        }</div>
+        {
+          this.props.vehicle === 'publicTransport' &&
+          <div>{
+            _(
+              'For information, public transport mode is currently in test version for some ' +
```
bbecquet: We had some discussion about that sentence in the Jira ticket. See with @Charlotte-Ziegler what is the chosen wording.

src/panel/direction/RouteVia.jsx
```diff
@@ -7,22 +7,22 @@ const RouteVia = ({ route, vehicle }) => {
     return <div className="routeVia">
       {_('Via', 'direction')} {route.legs[0].summary.replace(/^(.*), (.*)$/, '$1')}
     </div>;
+  } else {
```
bbecquet: Why this else? The early return construct was on purpose.

src/panel/direction/RouteResult.jsx
```diff
@@ -14,6 +14,7 @@ export default class RouteResult extends React.Component {
     vehicle: PropTypes.string,
     isLoading: PropTypes.bool,
     error: PropTypes.bool,
+    errorCode: PropTypes.number,
```
bbecquet: I wonder if we couldn't use a single `errorCode` prop in place of `error` *and* `errorCode`. With something like `-1` for unspecified errors and `0` for no error.
A best practice in React is to work with the minimum viable state, to avoid inconsistencies between state values that have some relation.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -125,7 +125,7 @@ export default class DirectionPanel extends React.Component {
           fire('set_route', { routes, vehicle, origin, destination });
         });
       } else {
-        this.setState({ isLoading: true, error: true });
+        this.setState({ isLoading: true, error: true, errorCode: directionResponse });
```
bbecquet: I don't understand how the errorCode is in the `directionResponse` here :thinking: I thought it would be in the `errorCode` value of the object… Ensuring the shape of the return is always the same would prevent that.
It would also simplify the test of a valid response.

src/panel/direction/RouteVia.jsx
xem: the else has always been there in the file's history, but ok :)

bbecquet: Ah? In that case I'm sorry. I just didn't understand these lines of diff which did nothing, but I may have misread it.

xem: np, I had just added a superfluous test ("else if(...) { "), instead of the "else" that was already here, and finally removed it

src/adapters/direction_api.js
xem: ok, it will be just { data, error } then

src/panel/direction/DirectionPanel.jsx
```diff
@@ -114,22 +114,26 @@ export default class DirectionPanel extends React.Component {
         destination,
         vehicle,
       );
-      if (directionResponse && directionResponse.routes && directionResponse.routes.length > 0) {
-        const routes = directionResponse.routes.map((route, i) => ({
+      if (directionResponse
```
bbecquet: It was supposed to be simpler, not more complex :P Why not just testing `directionResponse.error === 0` since you made all the job to uniformize it in the API?

src/panel/direction/RouteResult.jsx
```diff
@@ -13,8 +13,7 @@ export default class RouteResult extends React.Component {
     destination: PropTypes.string,
     vehicle: PropTypes.string,
     isLoading: PropTypes.bool,
-    error: PropTypes.bool,
-    errorCode: PropTypes.number,
+    error: PropTypes.number,
```
bbecquet: This is correct but now you need to change the initial `error` state of `DirectionPanel` from `false` to `0` to avoid this warning: 

Capture d’écran de 2020-02-11 16-09-28

src/adapters/direction_api.js
```diff
@@ -74,10 +74,16 @@ export default class DirectionApi {
     try {
       response = await Ajax.get(directionsUrl, directionsParams, { timeout });
     } catch (e) {
-      return;
+      if (Number.isInteger(e) && e >= 400 && e < 600) {
+        // Use the error codes 4xx and 5xx to display different error messages
+        response = { data: null, error: e };
```
bbecquet: I would just return here and below, instead of rewriting the response and risking it to overwritten below.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -114,22 +114,26 @@ export default class DirectionPanel extends React.Component {
...
           id: i,
         }));
-        this.setState({ isLoading: false, error: false, routes });
+        this.setState({ isLoading: false, error: 0, routes });
         window.execOnMapLoaded(() => {
           fire('set_route', { routes, vehicle, origin, destination });
         });
       } else {
-        this.setState({ isLoading: true, error: true });
+        this.setState({ isLoading: true, error: directionResponse.error });
```
bbecquet: I know `isLoading` was set to true before, but this looks like a bug.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -114,22 +114,25 @@ export default class DirectionPanel extends React.Component {
...
       } else {
-        this.setState({ isLoading: true, error: true });
+        // Error or empty response
+        this.setState({ isLoading: false, error: directionResponse.error });
         fire('clean_route');
       }
     } else {
-      this.setState({ isDirty: false, routes: [] });
+      // When both fields are not filled yet or not filled anymore
+      this.setState({ isLoading: false, isDirty: false, error: 0, routes: [] });
```
amatissart: we should also set `error: 0` on line 111, to avoid showing an obsolete error when a new request is triggered.

PR #532, opened 2020-02-05
Performances bench

G-Ray:
## Description
- Dust off existing bench (moved from `dev/` to `bench/` + fixes)
- Generate reproductible `.har` (network requests) and tracing from puppeteer, which can be imported from developer tools.
- Added webpack-bundle-analyzer


build/webpack.config.js
```diff
@@ -79,6 +80,23 @@ const sassChunkConfig = () => {
...
+    new webpack.NormalModuleReplacementPlugin(/mapbox-gl--ENV/, function(resource) {
+      if (buildMode === 'test') {
+        resource.request = resource.request.replace('--ENV', '-js-mock');
+      } else {
+        resource.request = resource.request.replace('--ENV', '');
+      }
+    }),
+    ...buildMode === 'production'
+      ? [
+        new BundleAnalyzerPlugin({
```
bbecquet: If I'm not wrong, you don't seem to use it for now, and maybe it's not worth doing that at each prod build.

G-Ray: a `report.html` is generated into the `public/build` folder. I agree that moving this analyze out from the build process would be better. We can write a npm script to complete this task, what do you think ?

G-Ray: updated

PR #535, opened 2020-02-07
Declare and use a generic Button component

bbecquet:
## Description
Introduce a new shared `<Button>` component, to display this type of buttons in a unified way, with simplified support for icons, better alignment and text ellipsis, etc.
Capture d’écran de 2020-02-07 12-04-52.
Use it on the Poi card and the Route result cards.

## Why
Simplify CSS and avoid introducing look and feel inconsistencies.
This is supposed to be part of the generic Qwant design system, but it's unclear if and when we'll use it on Maps. So until then, just declare a very simple React component which avoids re-declaring the same-ish CSS everytime we want to have good looking buttons. And if we switch to the real design system, it will be straightforward to replace this one by the real stuff.

### DIFF
--- a/src/components/ui/Button.jsx
+++ b/src/components/ui/Button.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import classnames from 'classnames';
+
+const Button = ({ children, icon, variant, className = '', type = 'button', onClick, ...rest }) =>
+  <button
+    type={type}
+    className={classnames('button', { [`button--${variant}`]: variant }, className)}
+    onClick={onClick}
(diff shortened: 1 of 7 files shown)


src/components/ui/Button.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import classnames from 'classnames';
+
+const Button = ({ children, icon, variant, className = '', type = 'button', onClick, ...rest }) =>
```
G-Ray: :+1:

PR #540, opened 2020-02-10
[Mobile routes] Buttons to access details and preview

bbecquet:
## Description
Uniformize action buttons on mobile route results:
- On cards, display "details" and "share" buttons
- The "preview" button is now available from the "details" fullscreen panel. (but still not available on public transports, as it's not ready)

## Screenshots
|Buttons on public transport cards|Buttons on car cards|Title + preview button in details|
|---|---|---|
|localhost_3000_routes__origin=latlon_48 79272_2 35662 destination=latlon_48 83006_2 34942 mode=driving(iPhone 6_7_8 Plus)|localhost_3000_routes__origin=latlon_48 79272_2 35662 destination=latlon_48 83006_2 34942 mode=driving(iPhone 6_7_8 Plus) (1)|localhost_3000_routes__origin=latlon_48 79272_2 35662 destination=latlon_48 83006_2 34942 mode=driving(iPhone 6_7_8 Plus) (2)|

### DIFF
--- a/src/scss/includes/utils.scss
+++ b/src/scss/includes/utils.scss
@@ -1,0 +1,3 @@
+.u-bold {
+  font-weight: bold;
+}
--- a/src/components/ui/Button.jsx
+++ b/src/components/ui/Button.jsx
@@ -9,7 +9,7 @@ const Button = ({ children, icon, variant, className = '', type = 'button', onCl
     {...rest}
   >
     {icon && <span className={`button-icon icon-${icon}`} />}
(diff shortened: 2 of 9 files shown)


src/scss/includes/utils.scss
```diff
@@ -0,0 +1,3 @@
+.u-bold {
```
bbecquet: Introduces a first generic utility class, to facilitate simple inline styling without having to target everything with a specific classname.

PR #541, opened 2020-02-11
[Mobile roadmaps] Interactive header style with scrolling

bbecquet:
## Description
Add a dynamic effect to the roadmap header to highlight it when the user has scrolled the view.
Listen to scroll events using the React [`useEffect` hook](https://reactjs.org/docs/hooks-effect.html) and throttle the handler using the [lodash implementation](https://lodash.com/docs/4.17.15#throttle) to prevent lagging.

## Why
UX improvement.

## Screenshots

simplescreenrecorder-2020-02-10_17 42 52

### DIFF
--- a/src/panel/direction/MobileRouteDetails.jsx
+++ b/src/panel/direction/MobileRouteDetails.jsx
@@ -1,13 +1,31 @@
 /* global _ */
-import React from 'react';
+import React, { useState, useRef, useEffect } from 'react';
 import RouteVia from './RouteVia';
 import RoadMap from './RoadMap';
 import Button from 'src/components/ui/Button';
 import { formatDuration } from 'src/libs/route_utils';
+import classnames from 'classnames';
+import throttle from 'lodash.throttle';
 
 const MobileRouteDetails =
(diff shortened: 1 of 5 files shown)


src/panel/direction/MobileRouteDetails.jsx
```diff
@@ -1,10 +1,29 @@
...
+({ id, route, origin, destination, vehicle, toggleDetails }) => {
+  const panelElement = useRef(null);
+  const [scrolledDown, setScrolledDown] = useState(false);
+
+  useEffect(() => {
+    const handleScroll = throttle(() => {
+      setScrolledDown(panelElement.current.scrollTop > 0);
+    }, 100, { leading: true, trailing: true });
+    panelElement.current.addEventListener('scroll', handleScroll, { passive: true });
+    return () => window.removeEventListener('scroll', handleScroll);
```
amatissart: Should not that be
```suggestion
    return () => panelElement.current.removeEventListener('scroll', handleScroll);
```
?

bbecquet: Oops, indeed. Thank you!

PR #545, opened 2020-02-14
[Mobile routes] Make result panel resizable

bbecquet:
## Description
Add resizable behavior to the route result panel on mobile.
The biggest part of this PR is actually simplifying the `<Panel>` component lifecycle for the management of the resize events, as well as the corresponding CSS.
Hopefully this makes it easier to read, but more importantly this fixes some UX problems with mobile panels, like the handler "jumping" below the cursor/finger and animations feeling awkward.

This is hard to validate with automatic testing, so if reviewers could play a bit with it on mobile devices that would be great :)

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -23,9 +23,14 @@ export default class Panel extends React.Component {
     minimizedTitle: PropTypes.node,
     resizable: PropTypes.bool,
     initialSize: PropTypes.string,
+    marginTop: PropTypes.number,
     close: PropTypes.func,
     className: PropTypes.string,
     white: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    marginTop: 0,
   }
 
   constructor(props) {
@@ -108,7 +113,8 @@ export default class Panel extends React.Component {
       size = !previousSize ? 'minimized' : null;
     } else {
       // the resize handler was really dragged-n-dropped
-      const sizes = [0, window.innerHeight / 2, window.innerHeight];
+      const maxSize = window.innerHeight - this.props.marginTop;
+      const sizes = [0, maxSize / 2, maxSize];
       const positionIndex = getClosestIndex(sizes, currentHeight);
       if (positionIndex === 0) {
         size = 'minimized';
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -39,11 +39,11 @@ export default class Panel extends React.Component {
   }
 
   componentDidMount() {
-    this.updateMobileMapUI(this.panelDOMElement.offsetHeight);
+    this.updateMobileMapUI();
   }
 
   componentDidUpdate() {
-    this.updateMobileMapUI(this.panelDOMElement.offsetHeight);
+    this.updateMobileMapUI();
   }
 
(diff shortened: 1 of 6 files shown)


src/components/ui/Panel.jsx
```diff
@@ -91,41 +94,41 @@ export default class Panel extends React.Component {
...
-    clearTimeout(this.timer);
 
-    const nativeEvent = event.nativeEvent;
-    const clientY = nativeEvent.changedTouches
-      ? nativeEvent.changedTouches[0].clientY
-      : nativeEvent.clientY;
-    const sizes = [0, window.innerHeight / 2, window.innerHeight];
-    const positionIndex = getClosestIndex(sizes, clientY);
+    document.removeEventListener('mousemove', this.moveCallback);
+    document.removeEventListener('touchmove', this.moveCallback);
```
G-Ray: Should we also remove the listeners in componenWillUnmount, just in case ?

bbecquet: Good idea, I'll add that. Thanks!

src/components/ui/Panel.jsx
```diff
@@ -91,41 +96,41 @@ export default class Panel extends React.Component {
...
-    } else if (positionIndex === 0) {
-      size = 'maximized';
+    if (event.timeStamp - this.interactionStarted < CLICK_RESIZE_TIMEOUT_MS) {
+      // the resize handler was only clicked, provide 'smart' resize
+      size = !previousSize ? 'minimized' : null;
+    } else {
+      // the resize handler was really dragged-n-dropped
+      const sizes = [0, window.innerHeight / 2, window.innerHeight];
+      const positionIndex = getClosestIndex(sizes, currentHeight);
+      if (positionIndex === 0) {
```
G-Ray: What about positionIndex === 1 ?

src/components/ui/Panel.jsx
```diff
@@ -91,41 +96,41 @@ export default class Panel extends React.Component {
...
+      }
     }
 
-    this.holding = false;
     this.setState({
-      isTransitioning: true,
+      holding: false,
       size,
       currentHeight: null,
+      isTransitioning: true,
```
G-Ray: I thing `isTransitioning` is not used anymore ?

src/components/ui/Panel.jsx
bbecquet: It's the medium position, i.e. the default case, which would result in `size=null`.

src/components/ui/Panel.jsx
bbecquet: It is for now. We need to call `updateMobileMapUI` after a CSS transition is done on the panel. It's currently done through `componentDidUpdate` when this state value is changed (see the lines below).
But you're right to spot it. It's an old hack with ugly coupling between CSS and JS… I'll try find an alternative using https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event

src/components/ui/Panel.jsx
G-Ray: ok, shouln't we asign `size = null` in this case ?

bbecquet: It's already the initial value, line 105.

PR #551, opened 2020-02-18
Mobile direction focus icons

xem:
## Description
Add a return button and an indication icon when a mobile itinerary field is focused

## Why
To be able to know which field we are in, and to be able to return to the itinerary form easily

## Screenshots
image
image

### DIFF
--- a/src/panel/direction/DirectionInput.jsx
+++ b/src/panel/direction/DirectionInput.jsx
@@ -106,6 +106,8 @@ export default class DirectionInput extends React.Component {
         onKeyPress={this.onKeyPress}
       />
       <div className="icon-x itinerary__field__clear" onMouseDown={this.clear} />
+      <div className="itinerary_field_return">&lt;==</div>
+      <div className="itinerary_field_icon">icon</div>
     </div>;
   }
 }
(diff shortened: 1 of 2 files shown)


amatissart:
I wonder how difficult it would be to simplify these css changes using flexbox.  
What do you think @bbecquet ?

bbecquet:
> I wonder how difficult it would be to simplify these css changes using flexbox.
> What do you think @bbecquet ?

It would be a less fragile layout and really easy to do with flexbox, but as things are done for now it would require a small change in the `<DirectionInput>` component to put a special class on the container div when the field is focused.

Right now this is the `<input>` element itself with is given a fixed, 100% width and top: 0 position, which is why we are forced to use these complex ~ selectors everywhere and use absolute positioning for all the siblings.
IMO it reflects more closely how things work if the parent container, not the input, gets a special display state. We would have code easier to read and maintain. Do you think it's worth the JS change?

**Note: I just discovered it's theoretically doable with CSS only with the new `:focus-within` pseudo-class https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within, but we may have support issue on some browsers**

bbecquet:
Switching to flexbox would also simplify the desktop layout, as for now it uses absolute positioning too for the icon and cross.

xem:
In general, I prefer using CSS (even if it involves sibling selectors, which I don't find very complex) than JS when it's possible.

In this case, indeed, the field and its neighbours have custom look and behavior on PC and mobile...
I'd rather keep the CSS solution, but if you prefer that I flag the parent container when the field is focused, it's ok for me too.

xem:
sure!

PR #552, opened 2020-02-18
Fix windows doc

xem:
## Description
Explain how to re-set DNS on windows 10's doc

## Why
Linux subsystem randomly forgets what dns to use
The public doc sets 8.8.8.8, but internally we use our own dns

### DIFF
--- a/docs/src/windows.md
+++ b/docs/src/windows.md
@@ -94,3 +94,11 @@ npm start
 ```
 npm watch
 ```
+
+- If W10's Linux Bash stops resolving domain names, here's a fix ([source](https://github.com/microsoft/WSL/issues/3268#issuecomment-485096972)):
+
+```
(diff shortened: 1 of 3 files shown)


docs/src/windows.md
```diff
@@ -93,4 +93,12 @@ npm start
 
 ```
 npm watch
+```
+
+- If W10's Linux Bash stops resolving domain names, here's a fix ([source](https://github.com/microsoft/WSL/issues/3268#issuecomment-485096972)):
```
GuillaumeGomez: I'm not sure if this is really relevant to put it here...

GuillaumeGomez:
There are also changes on JS/CSS files, is it linked to this PR?

GuillaumeGomez:
It seems like you included commits from #551.

PR #553, opened 2020-02-19
QMAPS-doc explain how to fix random DNS errors on W10

xem:
## Description
Explain how to re-set DNS on windows 10's doc

## Why
Linux subsystem randomly forgets what dns to use
The public doc sets 8.8.8.8, but internally we use our own dns

### DIFF
--- a/docs/src/windows.md
+++ b/docs/src/windows.md
@@ -94,3 +94,11 @@ npm start
 ```
 npm watch
 ```
+
+- If W10's Linux Bash stops resolving domain names, here's a fix ([source](https://github.com/microsoft/WSL/issues/3268#issuecomment-485096972)):
+
+```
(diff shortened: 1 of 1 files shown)


docs/src/windows.md
```diff
@@ -93,4 +93,12 @@ npm start
 
 ```
 npm watch
+```
+
+- If W10's Linux Bash stops resolving domain names, here's a fix ([source](https://github.com/microsoft/WSL/issues/3268#issuecomment-485096972)):
+
+```
+sudo rm /etc/resolv.conf
+echo "nameserver 8.8.8.8" | sudo tee /etc/resolv.conf
```
amatissart: Maybe we could use a placeholder here, to avoid favoring any particular DNS server :)

PR #554, opened 2020-02-19
migrate views from suggest.js to react

G-Ray:
## Description
*What this PR does. If needed, explain choices you made, complex parts, architectural changes, etc.*

## Why

## Screenshots


src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+          this.suggestList = [];
+
+          if (categoryResponse) {
+            // this.suggestList = this.suggestList.concat(categoryResponse);
+            this.setState({
+              categories: categoryResponse,
+            });
+          }
+
+          this.bragiPromise = null;
```
bbecquet: I think it's useless.

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+        this.categoryPromise = this.props.withCategories ?
+          CategoryService.getMatchingCategories(term) : null;
+
+        try {
+          const [bragiResponse, storeResponse, categoryResponse] = await Promise.all([
+            this.bragiPromise, this.historyPromise, this.categoryPromise,
+          ]);
+
+          if (!bragiResponse) {
+            return resolve(null);
```
bbecquet: It looks like it will ignore suggestions from store and category. Is it what we want?

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+            });
+          }
+
+          this.bragiPromise = null;
+          this.suggestList = this.suggestList.concat(bragiResponse);
+          this.suggestList = this.suggestList.concat(storeResponse);
+
+          // console.log(this.suggestList);
+
+          resolve(this.suggestList);
```
bbecquet: This whole promise could be simplified a lot so we see clearer :) And we don't need to store suggestList anymore.

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
-      </div>`;
+  render() {
+    const { prefixes } = this.props;
+    const { categories, pois, favorites } = this.state;
+    // console.log('pois.length', pois.length);
+
+    const items = [
+      ...prefixes,
+      ...categories,
+      ...pois.length > 0 ? pois.slice(0, SUGGEST_MAX_ITEMS - categories.length) : [],
```
bbecquet: Nit: You don't need the test on length, `[ ].slice(0, x)` will always return `[ ]`.

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+      ...favorites.filter(item => item.name.toUpperCase().includes(this.props.inputValue.toUpperCase())),
```
bbecquet: For clarity I would extract and name this filter as a pure function taking the favorite object and the inputValue.

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+          });
+        }}
+        onInputValueChange={value => {
+          console.log('onInputValueChange', value);
+          this.setState({ typedValue: value });
+          this.source(value);
+          this.props.onChange(value);
+        }}
+        selectedItem={this.state.selectedItem}
+        onStateChange={({ type, highlightedIndex }, stateAndHelpers) => {
```
bbecquet: Same for some of these inline functions.

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+        }}
+      >
+        {({
+          getInputProps,
+          getItemProps,
+          getMenuProps,
+          highlightedIndex,
+          getRootProps,
+        }) =>
+          <div>
```
bbecquet: Nit: could be a Fragment

src/adapters/suggest.js
```diff
@@ -1,258 +1,284 @@
...
+                    });
+                  }
+                },
+                onFocus: () => this.setState({ isOpen: true }),
+              }))}
+            </div>
+            <ul
+              {...getMenuProps()}
+              className={classNames('autocomplete_suggestions', this.props.className)}
+              style={{ display: 'block' }}
```
bbecquet: Is it needed?

src/panel/direction/DirectionInput.jsx
```diff
@@ -15,34 +15,33 @@ class DirectionInput extends React.Component {
...
+    if (this.props.claimFocus) {
+      this.focus();
+    }
   }
 
-  componentWillUnmount() {
-    if (this.suggest) {
-      this.suggest.destroy();
+  componentDidUpdate() {
+    if (this.props.claimFocus) {
```
bbecquet: We changed how we managed this in recent merges, don't wait too long before rebase :)

src/panel/direction/DirectionInput.jsx
```diff
@@ -75,28 +76,38 @@ class DirectionInput extends React.Component {
...
-          ? _('Start point', 'direction')
-          : _('End point', 'direction')}
-        value={this.props.value}
-        onChange={this.onChange}
-        onKeyPress={this.onKeyPress}
+      <div className={`itinerary_icon itinerary_icon_${pointType}`} />
+      <Suggest
+        className="direction_suggestions"
+        inputValue={this.props.value}
+        onChange={value => {this.onChange(value); }}
```
bbecquet: Why not just `this.onChange` like before?

src/ui_components/search_input.js
```diff
@@ -13,20 +15,102 @@ const MAPBOX_RESERVED_KEYS = [
...
+      <>
+        <Suggest
+          className="search_form__list"
+          withCategories
+          onChange={value => {
+            this.setState({ value });
+          }}
+          inputValue={this.state.value}
+          onSelect={item => this.selectItem(item, true)}
+          prefixes={[]}
```
bbecquet: I would add that as default value in Suggest so we can just omit it.

bbecquet:
As a general comment, it would be great to decouple the suggest component and item rendering from the code producing the suggestions. Things would be more cleaner and also easier to test in isolation.
I realize now (sorry, a bit late in the refacto…) it should have been the first step, even before diving in the react-ification.

G-Ray:
closing this in favor of https://github.com/QwantResearch/erdapfel/pull/583

PR #556, opened 2020-02-20
[Routes] Add a 'TEST' label to public transport option

bbecquet:
## Description
Add a 'TEST' label to the public transport mode button.

I also took the opportunity to improve markup & CSS a bit (better element and class name, flex for more adaptative layout, instead of alignment using absolute paddings & margins).
Notably, now the vehicle buttons are truly `<button>` elements, which enables keyboard navigation for free, whereas it was impossible before. I just added special styling to show the currently keyboard-focused button.

## Screenshots
|Before|After|
|---|---|
|maps dev qwant ninja_maps_routes__mode=driving pt=true(iPhone 6_7_8)|localhost_3000_routes__mode=driving pt=true(iPhone 6_7_8)|
|maps dev qwant ninja_maps_routes__mode=driving pt=true(iPhone 6_7_8) (1)|localhost_3000_routes__mode=driving pt=true(iPhone 6_7_8) (1)|
|maps dev qwant ninja_maps_routes__mode=driving pt=true|localhost_3000_routes__mode=driving pt=true|

### DIFF
--- a/src/panel/direction/VehicleSelector.jsx
+++ b/src/panel/direction/VehicleSelector.jsx
@@ -1,16 +1,23 @@
+/* global _ */
 import React from 'react';
 import classnames from 'classnames';
 import { getVehicleIcon } from 'src/libs/route_utils';
 
 const VehicleSelector = ({ vehicles, activeVehicle, onSelectVehicle }) =>
-  <div className="itinerary_vehicles">
-    {vehicles.map(vehicle => <span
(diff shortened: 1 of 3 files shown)


bbecquet:
> The new buttons layout with public transports disabled is probably unintentional
> image
> 
> When focused, the public transports button border is oddly positioned
> image

Both issues have been fixed in the last push.

PR #559, opened 2020-02-24
Add loading placeholder for category panel with re-usable component

bbecquet:
## Description
Introduce a visible loading status for the category panel, to avoid just hiding the panel while we fetch the first POIs, which resulted in the app feeling non-responsive to the user.
To avoid continuous toggling of loading state when we move the map, this loading state is used only when we enter the panel, not for other loading phases which are still done in the background.

Capture d’écran de 2020-02-24 13-58-47
For this loading status, we re-use the same kind of text placeholders already used during route computations.
Capture d’écran de 2020-02-24 13-59-08

To ensure consistency of these text placeholder, I introduced a new `<PlaceholderText>` component which standardizes and simplifies markup/style production. 

## Why
Improve UX feedback so the user knows something is happening.

## Screenshots
|Before|After|
|---|---|
|Peek 24-02-2020 12-08|Peek 24-02-2020 12-07|
|---|---|
|Peek 24-02-2020 13-51|Peek 24-02-2020 13-50|

### DIFF
--- a/src/components/ui/PlaceholderText.jsx
+++ b/src/components/ui/PlaceholderText.jsx
@@ -1,0 +1,12 @@
+import React from 'react';
+
+const PlaceholderText = ({ length = 10, tagName = 'div', className }) => {
+  const DomTag = tagName;
+  return <DomTag className={className}>
+    <span className="u-placeholder u-placeholder--text">
+      {Array.from({ length }).map(() => '_').join('')}
+    </span>
+  </DomTag>;
+};
+
+export default PlaceholderText;
--- a/src/panel/category/PoiItemListPlaceholder.jsx
+++ b/src/panel/category/PoiItemListPlaceholder.jsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { ItemList, Item } from 'src/components/ui/ItemList';
+import PlaceholderText from 'src/components/ui/PlaceholderText';
+
+const PoiItemPlaceholder = () =>
+  <div className="category__panel__item">
+    <div className="poiTitleImage u-placeholder" />
+    <PlaceholderText length={25} tagName="h3" className="category__panel__name"/>
+    <PlaceholderText length={15} tagName="p" className="category__panel__type" />
+    <PlaceholderText length={25} tagName="p" className="category__panel__address" />
+    <div className="openingHour"><PlaceholderText length={15} /></div>
+  </div>;
+
+const PoiItemListPlaceholder = ({ nbItems = 6 }) =>
+  <ItemList className="category__panel__items category__panel__items--placeholder">
+    {Array.from({ length: nbItems }).map((_item, index ) => <Item key={index}>
+      <PoiItemPlaceholder />
+    </Item>)}
+  </ItemList>;
+
+export default PoiItemListPlaceholder;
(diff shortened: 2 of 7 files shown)


src/components/ui/PlaceholderText.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const PlaceholderText = ({ length = 10, tagName = 'div', className }) => {
+  const DomTag = tagName;
+  return <DomTag className={className}>
```
xem: I assume this is used to make a container that is a div by default, but that can be modified by changing the tagName param.
Is DomTag necessary? Why not directly &lt;tagName className={className}>?

src/panel/category/PoiItemListPlaceholder.jsx
```diff
@@ -0,0 +1,21 @@
...
+    <div className="poiTitleImage u-placeholder" />
+    <PlaceholderText length={25} tagName="h3" className="category__panel__name"/>
+    <PlaceholderText length={15} tagName="p" className="category__panel__type" />
+    <PlaceholderText length={25} tagName="p" className="category__panel__address" />
+    <div className="openingHour"><PlaceholderText length={15} /></div>
+  </div>;
+
+const PoiItemListPlaceholder = ({ nbItems = 6 }) =>
+  <ItemList className="category__panel__items category__panel__items--placeholder">
+    {Array.from({ length: nbItems }).map((_item, index ) => <Item key={index}>
```
xem: Isn't "Array(nbItems)" simpler?

bbecquet: `Array(nbItems)` is a strange thing and is not clearly defined in the JS API. Here it wouldn't work as it doesn't return a real array but "something" with 5 slots…
As a result, the `.map( )` function doesn't transform the items and the resulting stuff is not what we want.

You can try in a browser console.

Firefox considers it's an array filled with `undefined` 
Capture d’écran de 2020-02-24 15-12-45

and Chrome tells it's `empty`
Capture d’écran de 2020-02-24 15-24-10

xem: ah yes, ughh JS

src/components/ui/PlaceholderText.jsx
bbecquet: It's related to the way JSX is transpiled to vanilla-JS. JSX needs dynamic component names to start with a capital letter, otherwise it's considered as the html element itself.
So if you write `<tagName />` in JSX, it will render the DOM node `<tagName>`, and we don't want that :)

xem: That's very unintuitive :D 
So... just to be sure, it would work if we had:
```
const PlaceholderText = ({ length = 10, TagName = 'div', className }) => {
  return <TagName className={className}>
```
?

src/panel/category/PoiItemListPlaceholder.jsx
bbecquet: In many projects it's common to use util functions to do that instead of using `Array.from`. For example, if we used the Ramda util library https://ramdajs.com/docs/#repeat.
If we do that often maybe it will be nice to implement a common util function.

PR #561, opened 2020-02-25
Qmaps 1308: add origin/destination pins on the map while completing itinerary fields

xem:
## Description
- Add start/end pin on field input or map click
- Center the map on the marker
- The destination pin can be set first if the destination field is focused
- The pin also appears when one field is set and the page is refreshed

## Why
We didn't have a visual clue on the map representing the field that was already filled

## Screenshots
image
image

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -53,6 +53,32 @@ export default class SceneDirection {
     listen('unhighlight_step', step => {
       this.unhighlightStep(step);
     });
+
+    listen('set_origin', pos => {
+      this.setOrigin(pos);
+    });
+
+    listen('set_destination', pos => {
+      this.setDestination(pos);
+    });
+  }
+
+  setOrigin = pos => {
+    const originMarker = createMarker(pos, 'itinerary_marker_origin', { draggable: true })
+      .addTo(this.map)
+      .on('dragend', event => this.refreshDirection('origin', event.target.getLngLat()));
+    this.routeMarkers.push(originMarker);
+  }
+
+  setDestination = pos => {
+    const destinationMarker = createMarker(pos,
+      'itinerary_marker_destination', {
+        draggable: true,
+        anchor: 'bottom',
+      })
+      .addTo(this.map)
+      .on('dragend', event => this.refreshDirection('destination', event.target.getLngLat()));
+    this.routeMarkers.push(destinationMarker);
   }
 
   addMarkerSteps(route) {
@@ -112,20 +138,8 @@ export default class SceneDirection {
 
     const { origin, destination } = originDestinationCoords(mainRoute);
 
-    const originMarker = createMarker(origin, 'itinerary_marker_origin', { draggable: true })
-      .addTo(this.map)
-      .on('dragend', event => this.refreshDirection('origin', event.target.getLngLat()));
-
-    const destinationMarker = createMarker(destination,
-      'itinerary_marker_destination', {
-        draggable: true,
-        anchor: 'bottom',
-      })
-      .addTo(this.map)
-      .on('dragend', event => this.refreshDirection('destination', event.target.getLngLat()));
-
-    this.routeMarkers.push(originMarker);
-    this.routeMarkers.push(destinationMarker);
+    this.setOrigin(origin);
+    this.setDestination(destination);
   }
 
   displayRoute() {
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -54,33 +54,35 @@ export default class SceneDirection {
       this.unhighlightStep(step);
     });
 
-    listen('set_origin', pos => {
-      this.setOrigin(pos);
+    listen('set_origin', poi => {
+      this.setOrigin(poi);
     });
 
-    listen('set_destination', pos => {
-      this.setDestination(pos);
+    listen('set_destination', poi => {
+      this.setDestination(poi);
     });
   }
 
-  setOrigin = pos => {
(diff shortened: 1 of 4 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -200,16 +221,19 @@ export default class DirectionPanel extends React.Component {
       return;
     }
 
-    // If origin field is empty, set it
-    // or, if destination field is empty, set it
-    if (persistentPointState.origin === null) {
+    // If destination field is empty and focused, set it
+    // else, if origin field is empty (focused or not), set it
+    // else, if destination field is empty (focused or not), set it
+    if (persistentPointState.destination === null && this.focusedField === 'destination') {
```
amatissart: This new `focusedField` state value is a bit odd. Ideally, we would like to avoid using fired events to propagate state.
And as far as I can see, it doesn't handle "blur" events.

Why not this solution:
```suggestion
    if (persistentPointState.destination === null && document.activeElement.id === 'itinerary_input_destination') {
```
It's still a bit hackish though.
The "React way" would probably be to use a Ref attached the input (they are already used in DirectionForm.jsx), or to propagate the state with a prop function.

xem: unfortunately, as soon as you click the map to set a marker, document.activeElement is the canvas and not the field :(
But if there's a React way, I'm interested!

src/adapters/scene_direction.js
```diff
@@ -53,6 +53,34 @@ export default class SceneDirection {
...
+      this.setDestination(pos);
+    });
+  }
+
+  setOrigin = pos => {
+    const originMarker = createMarker(pos, 'itinerary_marker_origin', { draggable: true })
+      .addTo(this.map)
+      .on('dragend', event => this.refreshDirection('origin', event.target.getLngLat()));
+    this.routeMarkers.push(originMarker);
+    window.map.mb.flyTo({ center: pos });
```
amatissart: This `flyTo` ignores the paddings used by the application (on mobile for example, the marker will not be centered because of the direction panel), and uses a potentially slow animation depending on the start position.
I would suggest to reuse the functions defined in scene.js

src/panel/direction/DirectionPanel.jsx
amatissart: hmm, you're right.  
The best way I can think of would be to define a new prop `onInputFocus` (similar to `onChangePoint`) in DirectionInput.jsx

bbecquet: Good job figuring a way to make it work, as that focus part is really tricky and already caused me much trouble :+1: 

Events aren't how to do that in React, as we already have all the pieces in place with the component hierarchy and refs. As a rule of thumb we should restrain custom events to cases where two pieces of the app can't communicate otherwise, namely React UI and the map view.

I think I would use a `onInputFocus` callback like @amatissart  proposed, store the focused field in state and just check it in `setDirectionPoint`.
And as usual, don't hesitate to do it in another PR so we can discuss this point specifically and merge the main feature without waiting.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -134,6 +150,11 @@ export default class DirectionPanel extends React.Component {
       // When both fields are not filled yet or not filled anymore
       this.setState({ isLoading: false, isDirty: false, error: 0, routes: [] });
       fire('clean_routes');
+      if (origin) {
```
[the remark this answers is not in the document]
xem: I'm afraid I can't factorize these snippets so easily, can I?

restorePoints does two "if" tests and waits for the map to be loaded in every case

```
      if (origin) {
          fire('set_origin', origin);
        });
      }
      if (destination) {
          fire('set_destination', destination);
        });
      }
```

while computeRoutes does an if/else and doesn't need to test that the map has loaded

```
      if (origin) {
        fire('set_origin', origin);
      } else if (destination) {
        fire('set_destination', destination);
      }
```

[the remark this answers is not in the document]
bbecquet: Ok I hadn't seen the `else if`, my bad.

src/adapters/scene_direction.js
```diff
@@ -53,6 +53,36 @@ export default class SceneDirection {
...
+
+  setOrigin = poi => {
+    const originMarker = createMarker(
+      poi.latLon,
+      'itinerary_marker_origin', { draggable: true }
+    )
+      .addTo(this.map)
+      .on('dragend', event => this.refreshDirection('origin', event.target.getLngLat()));
+    this.routeMarkers.push(originMarker);
+    fire('fit_map', poi);
```
bbecquet: As you suggested, passing an extra argument to keep the current zoom in the `fitMap` method would be great.

src/adapters/scene_direction.js
```diff
@@ -53,6 +53,36 @@ export default class SceneDirection {
...
+
+  setOrigin = poi => {
+    const originMarker = createMarker(
+      poi.latLon,
+      'itinerary_marker_origin', { draggable: true }
+    )
+      .addTo(this.map)
+      .on('dragend', event => this.refreshDirection('origin', event.target.getLngLat()));
+    this.routeMarkers.push(originMarker);
+    fire('fit_map', poi, false);
```
bbecquet: `setOrigin` and `setDestination` are also called by `updateMarkers`, so it explains the bug of zooming on the end point when hovering on an alternative.
I think the right way is to separate the `fire('fit_map', …)` from the actual marker change. Why not putting them in the listeners above?

PR #572, opened 2020-03-05
Improve resizable panel responsiveness

bbecquet:
## Description
Tries a new, simpler heuristic for the resizable panel responsiveness to touch gestures.
Previously, we defined 3 pixel values, corresponding to the 3 size states of the panel (`'default'`, `'minimized'` and `'maximized'`). The new size set after dragging and releasing the handle was determined by where the user released this handle, the closest pixel value giving the new size.

It had two related UX problems:
 - it required big gestures to cross the height thresholds. For example when going from `default` to `maximized`, the user had to drag the panel a long way, which was worse for big screens.
 - it wasn't responsive to quick swipe movements, which are common across many apps, appearing broken.

This PR proposes to fix that by using only the direction of the movement and a small threshold to trigger it. 

It also tries to improve code clarity with more explicit naming and simpler flow.

## Why
With the current implementation, the panel requires a lot of finger movement to be maximized, and sometimes doesn't appear to respond logically.

## Screenshots
Hard to showcase as gif, you'll have to test by yourself on a real phone :)

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -6,24 +6,31 @@ const getEventClientY = event => event.changedTouches
   ? event.changedTouches[0].clientY
   : event.clientY;
 
-// Pixel threshold to consider vertical resize moves
-const HEIGHT_RESIZE_THRESHOLD_PX = 50;
+// Pixel threshold to consider vertical swipes
+const SWIPE_THRESHOLD_PX = 50;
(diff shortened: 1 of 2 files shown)


bbecquet:
@amatissart I've made min-to-max and max-to-min resizing possible, introducing special cases in the resize function when the user releases the gesture in the bottom or top.

PR #576, opened 2020-03-09
Fix Favorite button keeping previous state

bbecquet:
## Description
Ensure the "save POI to favorite" button is up-to-date with the real favorite status of the POI displayed.

## Why
There is a currently a bug: when going from POI to POI (for example by clicking different places on the map), the favorite button keeps the saved/unsaved state of the first POI clicked in the series, or of the last which actually changed, ignoring the real state of the current POI.

This happens because the `ActionButtons` instance is now re-used by React, instead of being unmounted/remounted each time like it was during the migration phase.

### DIFF
--- a/src/panel/poi/ActionButtons.jsx
+++ b/src/panel/poi/ActionButtons.jsx
@@ -2,91 +2,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import Telemetry from '../../libs/telemetry';
-import { openAndWaitForClose as openMasqFavModalAndWaitForClose }
-  from 'src/modals/MasqFavoriteModal';
-import Store from '../../adapters/store';
-import { isFromPagesJaunes } from 'src/libs/pois';
-
-const store = new Store();
-
-async function isPoiFavorite(poi) {
-  try {
-    const storePoi = await store.has(poi);
-    return !!storePoi;
-  } catch (e) {
-    return false;
-  }
-}
 
 export default class ActionButtons extends React.Component {
   static propTypes = {
     poi: PropTypes.object.isRequired,
     isDirectionActive: PropTypes.bool,
-    isMasqEnabled: PropTypes.bool,
-    isFromCategory: PropTypes.bool,
-    isFromFavorite: PropTypes.bool,
     openDirection: PropTypes.func,
     openShare: PropTypes.func,
-  }
-
-  state = {
-    showPhoneNumber: !isFromPagesJaunes(this.props.poi),
-    poiIsInFavorite: false,
-  };
-
-  componentDidMount() {
-    this.storeAddHandler = listen('poi_added_to_favs', poi => {
-      if (poi === this.props.poi) {
-        this.setState({ poiIsInFavorite: true });
-      }
-    });
-
-    this.storeRemoveHandler = listen('poi_removed_from_favs', poi => {
-      if (poi === this.props.poi) {
-        this.setState({ poiIsInFavorite: false });
-      }
-    });
-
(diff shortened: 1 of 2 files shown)


bbecquet:
> What about the `showPhoneNumber` state value ?
> Should not it be reset too when another Poi is selected ?

Indeed, same kind of problem here.
The proper way to fix that is to make the `ActionButtons` component stateless and replace its local state by props. Meaning this state has to be controlled upstream. I'll refactor it. I'll probably move the fav state also.

Thanks!

xem:
> 
> 
> > What about the `showPhoneNumber` state value ?
> > Should not it be reset too when another Poi is selected ?
> 
> Indeed, same kind of problem here.
> The proper way to fix that is to make the `ActionButtons` component stateless and replace its local state by props. Meaning this state has to be controlled upstream. I'll refactor it. I'll probably move the fav state also.
> 
> Thanks!

I understand the problem and solution, though, I wonder why these buttons states aren't updated properly if all the info in the panel is correctly updated...?

bbecquet:
> I understand the problem and solution, though, I wonder why these buttons states aren't updated properly even if all the info in the panel is correctly updated...?

That's because we did something we shouldn't have in React, that is spreading parts of what should have been a single state in multiple parts along the component tree.
React will:
 - render stuff using `props` and `state`
 - reuse any component instance already in the tree, without unmounting/remounting it, keeping its internal state if it has one.

Here that means when going from POI to POI, this is the same `PoiPanel` and `ActionButtons` instances that are re-used, to minimize DOM updates. But as `ActionButtons` had stored stuff as state, this state was persistent and re-used to determine what to render, even if the `poi` prop changed.

The real source of the problem is `poiIsInFavorite` and the initial value of `showPhoneNumber` are functionally dependent (in the math meaning) from the Poi object, so they should have been managed at the same level, to guarantee that if the poi changes, they also change.

That's the concept of "single source of truth", which is central to React philosophy and should ensure consistency between the data and the UX.
And that's why we tend to [move state up the tree (React doc)](https://reactjs.org/docs/lifting-state-up.html), favoring big components at the top to manage the state and dumb, stateless, functional components down the tree to split the UX in really re-usable parts.

We should have done that in the first place. But during the progressive migration from doT.js we overlooked this. We didn't detect it because during the transition period, going from a POI to another involded a full unmouting/remounting of the tree. Now we always have a single React tree for the app and we should be careful about this kind of stuff.

bbecquet:
I turned `ActionButtons` into a stateless component by moving all the state stuff upwards. It should solve all problems.

xem:
Thanks a lot for the explanation @bbecquet !
It involves a lot of extra code, but seems right.
A test "show info on poi click" fails though, is it due to this refact?

bbecquet:
> A test "show info on poi click" fails though, is it due to this refact?

Yeah I still don't understand why :thinking: At first I thought it was because the rule to display phone number was wrong (it was :D), but it still fails. I'm investigating.

xem:
could it be an error in the test? for example if the test has always clicked a poi, then another poi and checked some exact values in the poi panel, it's possible that the test has always checked something wrong :p

bbecquet:
> could it be an error in the test? for example if the test has always clicked a poi, then another poi and checked some exact values in the poi panel, it's possible that the test has always checked something wrong :p

Yup, that's something that happened a lot, I'm exploring this lead too :)

bbecquet:
FYI, the failing test was due to typos after the refacto, like passing `this.isPhoneNumberVisible` instead of `this.state.isPhoneNumberVisible`. (@amatissart I also fixed the other ones)

PR #577, opened 2020-03-11
Fix benchmark URL overload

GuillaumeGomez:
The `set` method works but only if it has been overwrote once. In here, if you have an environment variable, the bench `set` call won't work.

### DIFF
--- a/bench/performances.js
+++ b/bench/performances.js
@@ -113,15 +113,11 @@ async function wait(time) {
 async function serverStart() {
   const App = require(path.join(__dirname, '../bin/app'));
   const configBuilder = require('@qwant/nconf-builder');
-
-  configBuilder.set('store:name', 'local_store');
-
-  configBuilder.set('mapStyle:poiMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`);
-  configBuilder.set('mapStyle:baseMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`);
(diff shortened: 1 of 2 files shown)


bench/performances.js
```diff
@@ -116,12 +116,12 @@ async function serverStart() {
 
   configBuilder.set('store:name', 'local_store');
 
-  configBuilder.set('mapStyle:poiMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`);
-  configBuilder.set('mapStyle:baseMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`);
   configBuilder.set('performance:enabled', true);
 
 
   const config = configBuilder.get();
```
amatissart: As no valid is url is defined in the default configuration, this would fail before the values could be overidden.
```suggestion
  const config = configBuilder.get_without_check();
```

bench/performances.js
```diff
@@ -116,12 +116,12 @@ async function serverStart() {
 
   configBuilder.set('store:name', 'local_store');
 
-  configBuilder.set('mapStyle:poiMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`);
-  configBuilder.set('mapStyle:baseMapUrl', `["http://localhost:${PORT}/fake_pbf/{z}/{x}/{y}.pbf"]`);
   configBuilder.set('performance:enabled', true);
```
amatissart: why not move all these overrides to use a single mechanism and remove this confusing usage of  `.set()` ?

PR #580, opened 2020-03-16
Qmaps 1354 - show addresses in destination panel fields in case of latlon pois

xem:
## Description
Complete destination fields with latlon POI's addresses when it's possible
- when clicking on the map
- when restoring pois on page reload

## Why
It's more readable than GPS coordinates

## Screenshots
chelou2

### DIFF
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -39,8 +39,8 @@ export default class DirectionForm extends React.Component {
     }
   }
 
-  onChangePoint = (which, textInput, point) => {
-    this.props.onChangeDirectionPoint(which, point);
+  onChangePoint = (which, value, point) => {
+    this.props.onChangeDirectionPoint(which, value, point);
   }
 
   onReverse = () => {
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -11,8 +11,6 @@ export default class DirectionForm extends React.Component {
     destination: PropTypes.object,
     onChangeDirectionPoint: PropTypes.func.isRequired,
     onReversePoints: PropTypes.func.isRequired,
-    onEmptyOrigin: PropTypes.func.isRequired,
-    onEmptyDestination: PropTypes.func.isRequired,
     vehicles: PropTypes.array.isRequired,
     onSelectVehicle: PropTypes.func.isRequired,
     activeVehicle: PropTypes.string.isRequired,
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -11,6 +11,8 @@ export default class DirectionForm extends React.Component {
     destination: PropTypes.object,
     onChangeDirectionPoint: PropTypes.func.isRequired,
     onReversePoints: PropTypes.func.isRequired,
+    onEmptyOrigin: PropTypes.func.isRequired,
+    onEmptyDestination: PropTypes.func.isRequired,
     vehicles: PropTypes.array.isRequired,
     onSelectVehicle: PropTypes.func.isRequired,
     activeVehicle: PropTypes.string.isRequired,
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -15,11 +15,8 @@ export default class DirectionForm extends React.Component {
     onSelectVehicle: PropTypes.func.isRequired,
     activeVehicle: PropTypes.string.isRequired,
     isInitializing: PropTypes.bool,
-  }
-
-  state = {
-    originInputText: '',
-    destinationInputText: '',
+    originInputText: PropTypes.string,
+    destinationInputText: PropTypes.string,
   }
 
   constructor(props) {
@@ -28,20 +25,11 @@ export default class DirectionForm extends React.Component {
     this.destinationRef = React.createRef();
   }
 
-  static getDerivedStateFromProps(props, state) {
-    const { origin, destination } = props;
-    return {
-      originInputText: origin ? origin.getInputValue() : state.originInputText,
-      destinationInputText: destination ? destination.getInputValue() : state.destinationInputText,
-    };
-  }
-
   componentDidUpdate(prevProps) {
     if (isMobileDevice() || this.props.isInitializing) {
       return;
     }
-    const { originInputText, destinationInputText } = this.state;
-    const { origin, destination } = this.props;
+    const { origin, destination, originInputText, destinationInputText } = this.props;
     if (!originInputText && (prevProps.destination !== destination || prevProps.isInitializing)) {
       this.originRef.current.focus();
     } else if (!destinationInputText && prevProps.origin !== origin) {
@@ -50,25 +38,21 @@ export default class DirectionForm extends React.Component {
   }
 
   onChangePoint = (which, textInput, point) => {
-    if (which === 'origin') {
-      this.setState({ originInputText: textInput });
-    } else {
-      this.setState({ destinationInputText: textInput });
-    }
     this.props.onChangeDirectionPoint(which, point);
   }
 
   onReverse = () => {
-    this.setState(prevState => ({
-      originInputText: prevState.destinationInputText,
-      destinationInputText: prevState.originInputText,
-    }));
     this.props.onReversePoints();
   }
 
   render() {
-    const { originInputText, destinationInputText } = this.state;
-    const { vehicles, activeVehicle, onSelectVehicle } = this.props;
+    const {
+      vehicles,
+      activeVehicle,
+      onSelectVehicle,
+      originInputText,
+      destinationInputText,
+    } = this.props;
 
     return <div className="itinerary_form">
       <div className="itinerary_fields">
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -117,12 +117,18 @@ export default class DirectionPanel extends React.Component {
       if (origin) {
         window.execOnMapLoaded(() => {
           fire('set_origin', origin);
+          if (!destination) {
+            fire('fit_map', origin);
+          }
         });
         this.setTextInput('origin', origin);
       }
       if (destination) {
         window.execOnMapLoaded(() => {
           fire('set_destination', destination);
+          if (!origin) {
+            fire('fit_map', destination);
+          }
         });
         this.setTextInput('destination', destination);
       }
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -74,6 +74,7 @@ export default class DirectionPanel extends React.Component {
     document.body.classList.add('directions-open');
     this.dragPointHandler = listen('change_direction_point', this.changeDirectionPoint);
     this.setPointHandler = listen('set_direction_point', this.setDirectionPoint);
+    this.refreshAddressesHandler = listen('get_addresses', this.refreshAddresses);
   }
 
   componentWillUnmount() {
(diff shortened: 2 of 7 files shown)


src/panel/direction/DirectionForm.jsx
```diff
@@ -11,15 +11,14 @@ export default class DirectionForm extends React.Component {
     destination: PropTypes.object,
     onChangeDirectionPoint: PropTypes.func.isRequired,
     onReversePoints: PropTypes.func.isRequired,
+    onEmptyOrigin: PropTypes.func.isRequired,
```
bbecquet: These two functions are required… but unused. Looks like obsolete WIP code you forgot to remove. (And it reminds me we can add an ESLint rule to highlight that)

src/panel/direction/DirectionPanel.jsx
```diff
@@ -206,13 +245,27 @@ export default class DirectionPanel extends React.Component {
...
       origin: previousState.destination,
       destination: previousState.origin,
+      originInputText: this.state.destinationInputText,
+      destinationInputText: this.state.originInputText,
       isDirty: true,
     }), this.update);
   }
 
-  changeDirectionPoint = (which, value) => {
+  emptyOrigin = () => {
```
bbecquet: According to the comment above, looks like you can delete these two functions.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -157,6 +177,25 @@ export default class DirectionPanel extends React.Component {
         fire('set_destination', destination);
       }
     }
+
+    let poiInfo;
```
bbecquet: By putting the reverse geocoding here, you postpone it to after the route is computed and displayed, whereas it could be done in parallel.
The two geocoding themselves could be done in parallel, but using two `await` make it in practice synchronized.
As a result, it's slower than needed.

This is also a bit strange to find these operations in a method called `computeRoutes` which explicitly tries to limit its scope to a single purpose. IMO it should be done in the `changeDirectionPoint`, so everything can be decoupled and can run more effectively.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -92,6 +95,16 @@ export default class DirectionPanel extends React.Component {
     Promise.all(poiRestorePromises).then(([ origin, destination ]) => {
       persistentPointState.origin = origin;
       persistentPointState.destination = destination;
+
+      // Show the raw POI name/latlon while waiting for computeRoutes to complete
+      if (origin) {
+        this.setState({ originInputText: origin.name });
```
bbecquet: Even if React can combine some operations behind the scene, it's better to regroup `setState` calls where it's possible, to limit useless re-renders.
Here, this can be grouped with the existing `setState` below.

xem:
> 
> 
> Some comments, mostly about improving state management.
> 
> I also noticed a bug when testing:
> If I compute a route, then empty a field (with the X button), I can't click on the map to fill it again. Clicks are just ignored.

fixed :)

src/panel/direction/DirectionPanel.jsx
xem: Hi,
I agree both requests must be made in parallel, I'm trying to fix it.
Regarding the place where the code should be, I initially chose computeRoutes because it seemed to be the only function called on field completion and on page reload, which is not the case for changeDirectionPoint (yet).
Maybe changeDirectionPoint only should compute addresses, and computeRoutes should call changeDirectionPoint, and if it does that, it could stop doing setStates directly... 🤔 
So I'm moving the code around hoping to find something more elegant, but I may need your help to confirm what is the nicest place/way to do it.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -92,6 +95,14 @@ export default class DirectionPanel extends React.Component {
     Promise.all(poiRestorePromises).then(([ origin, destination ]) => {
       persistentPointState.origin = origin;
       persistentPointState.destination = destination;
+
+      // Show the raw POI name/latlon while waiting for computeRoutes to complete
+      this.setState({
+        originInputText: origin ? origin.name : '',
+        destinationInputText: destination ? destination.name : '',
```
amatissart: To be consistent with the current implementation, addresses and streets should be rendered as labels (with city names)
```suggestion
        originInputText: origin ? origin.getInputValue() : '',
        destinationInputText: destination ? destination.getInputValue() : '',
```

src/panel/direction/DirectionPanel.jsx
amatissart: Another side effect of calling Idunn here, is that new requests are executed on each compute (eg when another vehicle is selected). That should not be necessary.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -157,6 +175,25 @@ export default class DirectionPanel extends React.Component {
         fire('set_destination', destination);
       }
     }
+
+    let poiInfo;
+    if (origin) {
+      if (origin.type === 'latlon') {
+        poiInfo = await IdunnPoi.poiApiLoad(origin);
```
amatissart: Note that `IdunnPoi.poiApiLoad` returns an instance of `IdunnPoi` that could probably be used as `origin` or `destination` state directly.  
Then, the response would be persisted and would't require other requests on next computes.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -206,13 +259,26 @@ export default class DirectionPanel extends React.Component {
     this.setState(previousState => ({
       origin: previousState.destination,
       destination: previousState.origin,
+      originInputText: this.state.destinationInputText,
```
bbecquet: `setState` is a potentially asynchronous operation, that's why when you set values based on other state values, you use this function form which gives safe access to the current state through it's first argument. So it's safer to do `previousState.destinationInputText`, like it's done above.
Here is a link to the explanation in the React doc. https://fr.reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

tests/integration/tests/direction.js
```diff
@@ -91,7 +91,7 @@ test('destination', async () => {
   const directionStartInput = await page.evaluate(() =>
     document.getElementById('itinerary_input_origin').value
   );
-  expect(directionStartInput).toEqual('');
+  expect(directionStartInput).toEqual(' ');
```
bbecquet: Why a single space character?

src/panel/direction/DirectionPanel.jsx
xem: k, can tou confirm me that I should also use previousState on lines 105 & 110 ?

src/panel/direction/DirectionPanel.jsx
```diff
@@ -206,13 +259,26 @@ export default class DirectionPanel extends React.Component {
...
+    persistentPointState[which] = point;
+    this.setState({
+      [which]: point,
+      isDirty: true,
+      [which + 'InputText']: value || '',
+    }, this.update);
+
+    // Retrieve addresses
+    if (which === 'origin') {
+      this.getAddresses(persistentPointState.origin, null);
```
bbecquet: I still don't get the point of the `getAddresses` function taking the two points. Among other things, it forces you to make these calls with `null` which look unnatural. As the action on both points is exactly the same, why not having a simpler function?

src/panel/direction/DirectionPanel.jsx
```diff
@@ -80,6 +83,35 @@ export default class DirectionPanel extends React.Component {
...
+          ? IdunnPoi.poiApiLoad(origin)
+          : Promise.resolve(origin)
+        : null,
+      destination
+        ? destination.type === 'latlon'
+          ? IdunnPoi.poiApiLoad(destination)
+          : Promise.resolve(destination)
+        : null,
+    ];
+    Promise.all(inputTextPromises).then(([ origin, destination ]) => {
```
bbecquet: If I'm not wrong, there is no need for this forced synchronization. Any of two points text can be updated independantly from the other, as soon as we get the response.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -80,6 +83,35 @@ export default class DirectionPanel extends React.Component {
     document.body.classList.remove('directions-open');
   }
 
+  getAddresses = (origin, destination) => {
+    const inputTextPromises = [
+      origin
```
bbecquet: We should find a way to avoid these unclear nested ternaries.
For example:
```js
origin && origin.type === 'latlon' ? IdunnPoi.poiApiLoad(origin) : origin
```
And sorry about the `Promise.resolve` that I used elsewhere, you can remove it if you want.
I thought it was needed but it can be omitted with `Promise.all`, any non-Promise value is considered as a Promise resolving with this value.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -92,6 +124,16 @@ export default class DirectionPanel extends React.Component {
     Promise.all(poiRestorePromises).then(([ origin, destination ]) => {
       persistentPointState.origin = origin;
       persistentPointState.destination = destination;
+
+      // Show the raw POI name/latlon while waiting for computeRoutes to complete
+      if (origin || destination) {
+        this.setState({
+          originInputText: ' ',
```
bbecquet: Why a single space character instead of an empty string like it's already done in the initial state?

src/panel/direction/DirectionPanel.jsx
xem: I agreen but if I make this function perform only one request, I don't know how to call it twice without making the two function calls (and the two API calls) synchronous...

src/panel/direction/DirectionPanel.jsx
xem: to avoid opening the suggest while the address is loading (didn't find how to disable it yet)

src/panel/direction/DirectionPanel.jsx
bbecquet: That's my point, I don't see why you need to synchronize the calls. Is there a UX reason?
Because otherwise, it's like remote work: the less synchronous stuff the better :smile:

src/panel/direction/DirectionPanel.jsx
bbecquet: Maybe a less hackish solution would be a dedicated state value, indicating that the form is still initializing and preventing the suggest to appear?

src/panel/direction/DirectionPanel.jsx
```diff
@@ -120,8 +145,15 @@ export default class DirectionPanel extends React.Component {
...
-      this.setState({ isDirty: false, isLoading: true, error: 0, routes: [] });
+      this.setState({
+        isDirty: false,
+        isLoading: true,
+        error: 0,
+        routes: [],
+      });
       const currentQueryId = ++this.lastQueryId;
+      fire('set_origin', origin);
+      fire('set_destination', destination);
```
amatissart: Why is this change required ?

xem: the idea was to show the start/end markers without waiting for the route to be computed (which can take a few seconds). If I remove these lines, and compute a route between two points clicked on the map, the last marker I clicked appears at the same time as the route, which is too late imo...

xem: ...Do you want me to make sure any preexisting marker is cleared before that? The way I did it, one of the markers may get doubled (harmlessly)

amatissart: I am not really worried about the markers, and I agree that the markers should be visible as soon as possible.  

But a side-effect of this change is that `set_origin` and `set_destination` will call `fitMap` before the route is computed.
So selecting the origin or destination that will trigger the compute, will actually cause the map position to jump twice, before the route is visible.

xem: OH, indeed. That's wrong. I guess I could add a fitMap = false parameter to setOrigin/Destination... is it ok for you?

amatissart: Yes, it should be ok :+1:

xem: ... but wait, I don't see anything that resembles fitMap() in these functions or in the functions called by them! 🤔 all they do is create a marker and add it to the map and in routeMarkers...

xem: actually I removed the fitMap entirely from set_origin and set_destination as there were zero cases where the fitMap was needed when calling them.
FitMap is only necessary in one case: when only one point/field is restored from the url, so I added the fitMap in this special case.

PR #582, opened 2020-03-17
[Clean-up] Improve code for suggestion fetching

bbecquet:
## Description
Separate the code producing address suggestion results from the presentation layer.
Also clean-up code and comment non-explicit stuff.

## Why
Improve maintainability.
Ease migration of the suggestion widget to React.

### DIFF
--- a/src/adapters/suggest.js
+++ b/src/adapters/suggest.js
@@ -3,10 +3,8 @@
 import Autocomplete from '../vendors/autocomplete';
 import IconManager from '../adapters/icon_manager';
 import ExtendedString from '../libs/string';
-import BragiPoi from './poi/bragi_poi';
 import PoiStore from './poi/poi_store';
 import Category from './category';
-import CategoryService from './category_service';
 import nconf from '@qwant/nconf-getter';
 
 const geocoderConfig = nconf.get().services.geocoder;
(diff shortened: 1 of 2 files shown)


src/adapters/suggest_sources.js
```diff
@@ -0,0 +1,55 @@
...
+      const categoryPromise = withCategories ? CategoryService.getMatchingCategories(term) : [];
+
+      try {
+        const [geocoderSuggestions, favorites, categories] =
+          await Promise.all([ geocoderPromise, favoritePromise, categoryPromise ]);
+
+        // For now don't display anything when there is no geocoder match
+        if (geocoderSuggestions.length === 0) {
+          return resolve([]);
+        }
```
amatissart: This is actually not equivalent to the current implementation, although arguably it could have been made clearer :)

`geocoderSuggesstions` will be `null` if the geocoder query has been aborted (by successive keystrokes), and that's why the current implementation calls `resolve(null)` in that case to avoid refresh the autocomplete results with an obsolete request.

An empty list should not be caught here, as some suggestions might exist among categories of favorites.

src/adapters/suggest_sources.js
```diff
@@ -0,0 +1,55 @@
...
+    const { lat, lon } = window.map.mb.getCenter();
+    return { lat, lon, zoom };
+  }
+  return {};
+}
+
+export function suggestResults(term, { withCategories, useFocus, focusMinZoom = 11 } = {}) {
+  let geocoderPromise;
+  let promise;
+  if (term === '') {
```
G-Ray: I think we can improve readability further by opting for early returns

bbecquet: I thought about it but as the current implementation requires attaching an 'abort' function to the returned promise in any case, I don't think that's possible.
BTW, I expect we'll be able to clean more things once everything is plugged to your new implementation.

src/adapters/suggest_sources.js
bbecquet: Thanks for the explanation! I had the wrong analysis and conclusion. I'll just restore the previous behavior and change the comment.

src/adapters/suggest.js
```diff
@@ -154,9 +107,6 @@ export default class Suggest {
 
   async onSubmit() {
     if (this.pending) {
-      if (this.bragiPromise) {
-        this.bragiPromise.abort();
-      }
```
amatissart: This change has actually a small side-effect:
without this abort, the suggestions results may be briefly re-rendered after the submit action occurred (and the pending request returned its results).

Peek 19-03-2020 11-10

amatissart: Arf, nevermind. This behavior is also present on "master".

PR #583, opened 2020-03-17
Add SuggestsDropdown ui component

G-Ray:
## Description
Replace autocomplete logic with SuggestsDropdown.

I had some flaky tests on my comptuer, but I restarted the CI builds multiple times successfully


src/components/ui/SuggestItem.jsx
```diff
@@ -0,0 +1,54 @@
...
+const SuggestItem = ({
+  className = '',
+  suggest,
+  isHighlighted = false,
+  onClick,
+  ...rest
+}) =>
+  <>
+    {suggest.categoryLabel &&
+      <h3 className="autocomplete_suggestion__category_title">
```
bbecquet: Rendering that here implies we'll create a `<ul>` with some `<h3>` children, which is invalid HTML.
Maybe a better architecture would be to inject a special object at the right place in the data array (before favs or categories), and render them as real list items with no interaction. The only difficulty would be how it affects indices in keyboard nav, but just 'jumping' these items should be easy enough.

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -0,0 +1,78 @@
...
+    <ul
+      className={classnames('autocomplete_suggestions', className)}
+      style={{ display: 'block', width: '100%' }}
+    >
+      {suggests.map((suggest, index) =>
+        <SuggestItem
+          isHighlighted={highlighted === index}
+          key={index}
+          suggest={suggest}
+          onClick={() => onSelect(index)}
```
bbecquet: What do you think of working with item objects instead of indices, to have a cleaner API? That way indices can be thought as an internal implementation detail and decoupled from the data itself.

tests/integration/tests/autocomplete.js
```diff
@@ -71,7 +71,9 @@ test('keyboard navigation', async () => {
   responseHandler.addPreparedResponse(mockAutocomplete, /autocomplete/);
   await page.goto(APP_URL);
   await autocompleteHelper.typeAndWait(TypedSearch);
+  await page.waitFor(100);
```
G-Ray: Quickfix to remove

src/adapters/suggest.js
```diff
@@ -62,11 +62,33 @@ export default class Suggest {
...
+        const elem = document.createElement('div');
+        elem.setAttribute('id', 'react-suggests-' + tagSelector);
+        this.searchInputDomHandler.parentNode.insertBefore(
+          elem,
+          this.searchInputDomHandler.nextSibling
         );
+
+        const typedValue = this.searchInputDomHandler.value;
+
+        ReactDOM.render(
```
G-Ray: This will be mounted each time \renderItems` is called. I guess performances should be the same as static rendering this way.

src/components/ui/SuggestItem.jsx
```diff
@@ -10,23 +11,29 @@ const ItemLabels = ({ firstLabel, secondLabel }) =>
     <div className="autocomplete_suggestion__second_line">{secondLabel}</div>
   </div>;
 
-const GeolocationItem = () =>
+const GeolocationItem = ({ isHighlighted }) =>
```
bbecquet: To avoid managing highlighted state on each component, you could have them render only the specific content of each item, and put in the parent the rendering of the standard list item container as a `<li>` with the `autocomplete_suggestion` and optional `selected` list.

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -0,0 +1,78 @@
...
+          onHighlight(null);
+        }
+      }
+
+      if (key === 'Enter') {
+        if (highlighted !== null) {
+          onSelect(suggestItems[highlighted]);
+        }
+      }
+    };
```
bbecquet: The `Esc` key also has to be managed to close the dropdown. Or maybe in the parent as it is the one controlling the drowdown display status?

src/components/ui/SuggestItem.jsx
bbecquet: Sorry I should have probably done that in the PR where I separated data and rendering.

src/components/ui/SuggestsDropdown.jsx
G-Ray: It is handled elsewhere, so I should have not changed the existing behavior

G-Ray: nope, I lied, There was a case to handle. I fixed it thanks

src/components/ui/SuggestItem.jsx
G-Ray: I agree but this would lead us to a css refactoring, I prefer to let the style untouched in this PR

tests/integration/tests/autocomplete.js
```diff
@@ -71,7 +71,10 @@ test('keyboard navigation', async () => {
   responseHandler.addPreparedResponse(mockAutocomplete, /autocomplete/);
   await page.goto(APP_URL);
   await autocompleteHelper.typeAndWait(TypedSearch);
+  await page.waitForSelector('.autocomplete_suggestions');
+  await page.waitFor(100);
   await autocompleteHelper.pressDown();
+  await page.waitFor(100);
```
G-Ray: this seems to be requirer in order to wait for the event to be completed. Let me know if you have a better idea

amatissart:
I haven't digged into the details yet, but the behavior looks good.  
One issue is that the suggestions remain visible when the app is launched with a query in the URL (eg: http://localhost:3000/?q=marseille )

G-Ray:
> I haven't digged into the details yet, but the behavior looks good.
> One issue is that the suggestions remain visible when the app is launched with a query in the URL (eg: http://localhost:3000/?q=marseille )

thanks, It should be fixed

tests/integration/tests/autocomplete.js
bbecquet: I guess it could be replaced by a selector on the `highlighted` status of the suggest item which is supposed to be selected after a first key down. Maybe we can define a function which press and wait?
But if the test is reliable I would be ok to move on with this implementation and try to refactor it separately.

PR #584, opened 2020-03-19
[Suggest refacto] More data vs. rendering separation + introducing React 

bbecquet:
*More easily reviewed one commit at a time*

## Description
Follow up to https://github.com/QwantResearch/erdapfel/pull/582
Goes one step further in the suggest refacto, by removing stuff from the current suggest component so the migration itself has to change less things.

Instead of building a big DOM string piece-by-piece for the drop-down content, focus on building the list of items to put in the suggest, trying to simplify the logic in the process, *then* render each of it.

For the render itself, replace old template string by new elementary React component and uses the `renderStaticReact` trick to render them as a DOM string, so the suggest component doesn't see the difference.
It still requires to render some attributes specific to the actual implementation (the 'id' and 'value' data attributes), but we'll soon be able to remove them and quickly make everything cleaner.
Also I didn't change the DOM or CSS to keep exactly the same behavior, but they'll be improved later. Let's advance step by step :)

## Why
Make things easier to connect to @G-Ray's ongoing work in https://github.com/QwantResearch/erdapfel/pull/583

## Screenshots
No visual or UX change.

### DIFF
--- a/src/components/SuggestItem.jsx
+++ b/src/components/SuggestItem.jsx
@@ -1,0 +1,82 @@
+/* global _ */
+import React from 'react';
+import NavigatorGeolocalisationPoi from 'src/adapters/poi/specials/navigator_geolocalisation_poi';
+import IconManager from '../adapters/icon_manager';
+import Category from 'src/adapters/category';
+
+const ItemLabels = ({ firstLabel, secondLabel }) =>
+  <div className="autocomplete_suggestion__lines_container">
+    <div className="autocomplete_suggestion__first_line">{firstLabel}</div>
+    <div className="autocomplete_suggestion__second_line">{secondLabel}</div>
+  </div>;
+
+const GeolocationItem = () =>
+  <div
+    className="autocomplete_suggestion itinerary_suggest_your_position"
+    data-id="geolocalisation"
+    data-val={_('Your position', 'direction')}
+  >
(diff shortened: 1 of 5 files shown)


src/components/SuggestItem.jsx
```diff
@@ -0,0 +1,82 @@
...
+        style={{ color: icon ? icon.color : '' }}
+        className={`autocomplete-icon icon icon-${icon.iconClass}`}
+      />
+      <ItemLabels firstLabel={name} secondLabel={alternativeName} />
+    </div>
+  );
+};
+
+const SeparatorLabel = ({ label }) =>
+  <h3 className="autocomplete_suggestion__category_title" onMouseDown="return false;">
```
amatissart: the `onmousedown` attribute is not rendered in the static markup.  
As a consequence mouse events on this separator will make the input field lose focus.

bbecquet: Ah right, thanks. I guess I'll have to disable these events by targeting DOM nodes after the rendering, making the refacto a bit less clean, but that's temporary.

bbecquet: @amatissart Here is a fix https://github.com/QwantResearch/erdapfel/pull/584/commits/02178cf362d5d35e6f099962e98487996205bd63

PR #587, opened 2020-03-26
Fix Direction input tests to match the real behavior

bbecquet:
## Description
Fix integration tests of the direction panel so they match the actual behavior of the form initialization:
when POIs are passed as origin and/or destination in the URL, we call the Idunn API to retrieve the real POI objects and we fill input field with their real names.
This requires mocking Idunn calls, like we do in the POI panel. As I needed two POIs, I re-used the existing one (Musée d'Orsay :)) and created another mock.

## Why
Better testing and will ease the completion of this other PR https://github.com/QwantResearch/erdapfel/pull/580

### DIFF
--- a/tests/integration/tests/direction.js
+++ b/tests/integration/tests/direction.js
@@ -4,6 +4,8 @@ import ResponseHandler from '../helpers/response_handler';
 const ROUTES_PATH = 'routes';
 const mockAutocomplete = require('../../__data__/autocomplete.json');
 const mockMapBox = require('../../__data__/mapbox.json');
+const mockPoi1 = require('../../__data__/poi.json');
+const mockPoi2 = require('../../__data__/poi2.json');
 
 let browser;
 let page;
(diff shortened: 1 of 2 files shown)


tests/integration/tests/direction.js
```diff
@@ -80,14 +82,13 @@ test('route flag', async () => {
...
-
 test('destination', async () => {
-  expect.assertions(3);
-  await page.goto(`${APP_URL}/${ROUTES_PATH}/?destination=latlon:47.4:7.5@Monoprix Nice`);
+  responseHandler.addPreparedResponse(mockPoi1, new RegExp(`places/${mockPoi1.id}`));
 
+  expect.assertions(2);
+  await page.goto(`${APP_URL}/${ROUTES_PATH}/?destination=${mockPoi1.id}`);
   await page.waitForSelector('#itinerary_input_origin');
-  const smallToolBar = await page.waitForSelector('.top_bar--small');
```
bbecquet: In the three cases I removed the testing of this stuff for simplification, as it's not related to the input fields and is already tested in another case.

bbecquet:
> Note that one more mocked request will probably be required in #580

Yup, this is on purpose that I let this one out, so @xem can adapt it to his PR needs himself :) I think it will be easier now that the other problems are solved.

PR #588, opened 2020-03-30
[Covid 19] Display data from caresteouvert.fr

bbecquet:
## Description
This PR uses the new `covid19` block on french POIs, coming from the ["Ça reste ouvert" project](https://caresteouvert.fr), to display information on the lockdown in the POI panel.

It defines and uses a new feature flag `covid19.enabled`, to be able to test and remove the feature easily. 

Then we define a new `Covid19` component to manage this special block and rules. I chose to make some refacto on the way to be able to re-use the existing time table display of opening times in this block. This makes the PR grows bigger than I hoped, but make the whole thing easier at the end.

Outside from the block, several other things are adjusted:
 - we don't display the current open status in poi popups
 - if we have covid19 data, we don't display the current open status in the mobile POI Card 
 - if we have covid19 data, we don't display the current open status as header of "normal" opening hours

PJ POIs have a covid19 block of status `unkown` and no URL to contribute to Ça reste ouvert, so we just display some very neutral info.

## Screenshots
|Situation|Screenshot|
|---|---|
|open, with specified hours|Capture d’écran de 2020-03-30 17-57-03|
|The same in PJ|Capture d’écran de 2020-03-30 17-58-22|
|open, with no specified hours|Capture d’écran de 2020-03-30 17-57-14|
|closed|Capture d’écran de 2020-03-30 17-58-01|
|unknown, but existing in CRO*|Capture d’écran de 2020-03-30 17-57-48

### DIFF
--- a/src/panel/poi/PoiCard.jsx
+++ b/src/panel/poi/PoiCard.jsx
@@ -21,14 +21,16 @@ class PoiCard extends React.Component {
   }
 
   render() {
-    const { poi, closeAction, showDetails, openDirection } = this.props;
+    const { poi, closeAction, showDetails, openDirection, covid19Enabled } = this.props;
+
+    const hideOpeningHour = covid19Enabled && poi.blocks.find(b => b.type === 'covid19');
 
     return <div className="poi_card" ref={this.cardRef}>
       <div className="poi_card__description_container">
         <PoiTitleImage poi={poi} iconOnly={true} />
         <div>
           <PoiHeader poi={poi} />
-          <OpeningHour poi={poi} />
+          {!hideOpeningHour && <OpeningHour poi={poi} />}
         </div>
       </div>
       <div className="poi_card__action_container">
--- a/src/panel/poi/PoiCard.jsx
+++ b/src/panel/poi/PoiCard.jsx
@@ -23,7 +23,8 @@ class PoiCard extends React.Component {
   render() {
     const { poi, closeAction, showDetails, openDirection, covid19Enabled } = this.props;
 
-    const hideOpeningHour = covid19Enabled && poi.blocks.find(b => b.type === 'covid19');
+    const hideOpeningHour = covid19Enabled
+      && poi.blocks && poi.blocks.find(b => b.type === 'covid19');
 
     return <div className="poi_card" ref={this.cardRef}>
       <div className="poi_card__description_container">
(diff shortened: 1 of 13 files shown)


src/panel/poi/blocks/Covid19.jsx
```diff
@@ -0,0 +1,77 @@
...
+    schedule = opening_hours && new OsmSchedule(opening_hours, scheduleMessages);
+    content = <Fragment>
+      <div className="covid19-status covid19-status--open">{covidStrings.statusOpen}</div>
+      {schedule && <div className="covid19-timeTableContainer">
+        <i className="icon-icon_clock" />
+        <TimeTable title={<OpeningHour poi={poi} />} schedule={schedule} />
+      </div>}
+      {!schedule && <div className="covid19-changeWarning">{covidStrings.hoursMayChange}</div>}
+      {additionalInfo}
+      {source}
```
amatissart: The link to caresteouvert.fr may be useful independently from the status (and if`contribute_url` is not null), especially if no data is available

src/panel/poi/blocks/Covid19.jsx
```diff
@@ -0,0 +1,77 @@
...
+    msg: 'Fermé',
+    color: '#8c0212',
+  },
+};
+
+const getContent = (poi, { status, opening_hours, note, contribute_url }) => {
+  const additionalInfo = note &&
+    <div className="covid19-note">
+      <i className="icon-icon_info" />
+      <span dangerouslySetInnerHTML={{ __html: note }}/>
```
amatissart: I see no good reason to inject html here, as `note` contains just text anyway, and is possibly unsafe.

src/panel/poi/blocks/Covid19.jsx
bbecquet: I agree with you and that's the topic of ongoing discussions… I'll try to expand the cases where we can promote contribution.

src/panel/poi/PoiCard.jsx
```diff
@@ -21,14 +21,16 @@ class PoiCard extends React.Component {
   }
 
   render() {
-    const { poi, closeAction, showDetails, openDirection } = this.props;
+    const { poi, closeAction, showDetails, openDirection, covid19Enabled } = this.props;
+
+    const hideOpeningHour = covid19Enabled && poi.blocks.find(b => b.type === 'covid19');
```
amatissart: `poi.blocks` may not exist here (for example for BragiPoi from geocoder results).
```suggestion
    const hideOpeningHour = covid19Enabled && poi.blocks && poi.blocks.find(b => b.type === 'covid19');
```

bbecquet: Thank you!

PR #589, opened 2020-03-31
covid19 fix: Opening hours title does not use hours from the new block

amatissart:
## Description
The opening hours title could be wrong, because it used the default opening hours, instead of those that appear in the new block.

### DIFF
--- a/src/panel/poi/blocks/Covid19.jsx
+++ b/src/panel/poi/blocks/Covid19.jsx
@@ -43,7 +43,7 @@ const getContent = (poi, { status, opening_hours, note, contribute_url }) => {
       <div className="covid19-status covid19-status--open">{covidStrings.statusOpen}</div>
       {schedule && <div className="covid19-timeTableContainer">
         <i className="icon-icon_clock" />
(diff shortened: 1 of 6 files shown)


src/panel/poi/blocks/Covid19.jsx
```diff
@@ -43,7 +43,7 @@ const getContent = (poi, { status, opening_hours, note, contribute_url }) => {
       <div className="covid19-status covid19-status--open">{covidStrings.statusOpen}</div>
       {schedule && <div className="covid19-timeTableContainer">
         <i className="icon-icon_clock" />
-        <TimeTable title={<OpeningHour poi={poi} />} schedule={schedule} />
+        <TimeTable title={<OpeningHour openingHours={opening_hours} />} schedule={schedule} />
```
bbecquet: This should allow some simplification, by not requiring `poi` to be passed at all in this function and even this component. Much cleaner indeed!

amatissart: Done

PR #590, opened 2020-03-31
[Covid-19] Fix display of 24/7 opening hours in Covid mode

bbecquet:
## Description
Don't show 24/7 normal opening status when Covid flag is enabled.
Unfortunalety required *a lot* of refacto on components managing hours to be able to manage every case…
The best option from the very start would have been to manage all collapsible POI info blocks in a generic component, but that would touch too many unrelated parts.
Let's keep that for a future cleaning.

## Why
24/7 are managed as a special case with collapse/expand action.

## Screenshots
|Before|After (expanded)|
|---|---|
|Capture d’écran de 2020-03-31 16-25-14|Capture d’écran de 2020-03-31 16-24-25|

### DIFF
--- a/src/panel/poi/blocks/TimeTable.jsx
+++ b/src/panel/poi/blocks/TimeTable.jsx
@@ -12,12 +12,12 @@ function showHour(day) {
   return _('Closed', 'hour block');
 }
 
-function showHours(displayHours) {
+const Days = ({ days }) => {
   const dayNumber = new Date().getDay();
 
   return <table>
     <tbody>
-      {displayHours.map((day, i) =>
+      {days.map((day, i) =>
         <tr key={i} className={
           classnames({ 'currentDay': (i + 1) % 7 === dayNumber })
         }>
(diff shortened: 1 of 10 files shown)


src/panel/poi/blocks/TimeTable.jsx
```diff
@@ -11,44 +12,55 @@ function showHour(day) {
...
+  return <div className={classnames('timetable', {
+    'timetable--collapsable': collapsable,
+    'timetable--collapsed': isCollapsed,
+  })}>
+    <div className="timetable-status" onClick={() => {
+      if (collapsable) {
+        setCollapsed(!isCollapsed);
+      }
+    }}>
+      <div className="timetable-status-text">{header}</div>
```
amatissart: As a consequence, the default header ("open" or "closed") font-size has changed from 15px to 14px.

bbecquet: I added an explicit rule for that https://github.com/QwantResearch/erdapfel/pull/590/commits/dc7c5ac9a6060e0a3fc3b73a961bae62de1b96a8
And I hope we'll have some time soon to clean the mess which is font-size hierarchy in the app :)

PR #592, opened 2020-03-31
[Covid-19] Clickable tooltip with legal message

bbecquet:
WIP

## Screenshots
Capture d’écran de 2020-03-31 18-02-29


bbecquet:
Implemented as a simple message in https://github.com/QwantResearch/erdapfel/pull/593

PR #594, opened 2020-04-03
Fix back to list from query

G-Ray:
when instancing the app with a query in url, like `q=park`, a list of pois is returned.
If we select one of them, the poi is focused on the map. If we click on "go back to list", the application's panel crashes and there is no way to recover from this state without a page reload.
This is due to the way `backToList` build an url only for a category.

This PR fixes this behavior by encapsulating query/category into a `poiFilters` object. This way we can extends it as we need to add more filters, and we can build any url with appropriate filters.

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -166,7 +166,7 @@ Scene.prototype.initMapBox = function() {
 
   listen('map_mark_poi', (poi, options) => {
     this.ensureMarkerIsVisible(poi, options);
-    if (!options.isFromCategory) {
+    if (!options.resource || !options.resource.category) {
       this.addMarker(poi, options);
     }
   });
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -166,7 +166,7 @@ Scene.prototype.initMapBox = function() {
 
   listen('map_mark_poi', (poi, options) => {
     this.ensureMarkerIsVisible(poi, options);
-    if (!options.resource || !options.resource.category) {
+    if (!options.poiFilters || !options.poiFilters.category) {
       this.addMarker(poi, options);
     }
   });
--- a/src/adapters/scene_category.js
+++ b/src/adapters/scene_category.js
@@ -19,12 +19,12 @@ export default class SceneCategory {
     listen('highlight_category_marker', (poi, highlight) => {
       this.highlightPoiMarker(poi, highlight);
     });
-    listen('click_category_poi', (poi, categoryName) => {
-      this.selectPoi(poi, categoryName);
+    listen('click_category_poi', (poi, resource) => {
+      this.selectPoi(poi, resource);
     });
   }
 
-  selectPoi = (poi, categoryName) => {
+  selectPoi = (poi, resource) => {
     const previousMarker = document.querySelector('.mapboxgl-marker.active');
     if (previousMarker) {
       previousMarker.classList.remove('active');
@@ -37,14 +37,13 @@ export default class SceneCategory {
           template: 'multiple',
           zone: 'list',
           element: 'item',
-          category: categoryName,
+          category: resource.category,
         })
       );
     }
     window.app.navigateTo(`/place/${toUrl(poi)}`, {
       poi,
-      isFromCategory: true,
-      sourceCategory: categoryName,
+      resource,
       centerMap: true,
     });
     this.highlightPoiMarker(poi, true);
@@ -59,7 +58,9 @@ export default class SceneCategory {
         poi.marker_id = `marker_${id}`;
         marker.onclick = function(e) {
           e.stopPropagation();
-          fire('click_category_poi', poi, categoryName);
+          fire('click_category_poi', poi, {
+            category: categoryName,
+          });
         };
         marker.onmouseover = function(e) {
           fire('open_popup', poi, e);
--- a/src/adapters/scene_category.js
+++ b/src/adapters/scene_category.js
@@ -19,12 +19,12 @@ export default class SceneCategory {
     listen('highlight_category_marker', (poi, highlight) => {
       this.highlightPoiMarker(poi, highlight);
     });
-    listen('click_category_poi', (poi, resource) => {
-      this.selectPoi(poi, resource);
+    listen('click_category_poi', (poi, poiFilters) => {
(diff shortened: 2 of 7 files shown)


src/panel/poi/PoiPanel.jsx
```diff
@@ -233,7 +234,7 @@ export default class PoiPanel extends React.Component {
         text: _('Back to favorite'),
         className: 'poi_panel__back_to_list',
       };
-    } else if (isFromCategory) {
+    } else if (resource && (resource.category || resource.query)) {
```
bbecquet, line 237: You can put a default prop value for `resource` in this component too so you can remove the existence tests everywhere and have shorter lines (the next one breaks linter rules)

src/adapters/scene.js
```diff
@@ -166,7 +166,7 @@ Scene.prototype.initMapBox = function() {
 
   listen('map_mark_poi', (poi, options) => {
     this.ensureMarkerIsVisible(poi, options);
-    if (!options.isFromCategory) {
+    if (!options.resource || !options.resource.category) {
```
bbecquet, line 169: I'm not a fan of this `resource` name, it sounds really too vague and looks like it describes a single piece of data, not filtering criteria.
What about something like `poiFilters`, `poiParams`, `criteria`?

src/adapters/scene_category.js
```diff
@@ -19,12 +19,12 @@ export default class SceneCategory {
     listen('highlight_category_marker', (poi, highlight) => {
       this.highlightPoiMarker(poi, highlight);
     });
-    listen('click_category_poi', (poi, categoryName) => {
-      this.selectPoi(poi, categoryName);
+    listen('click_category_poi', (poi, resource) => {
```
bbecquet, line 22: If we refactor the filters a lot of things will have to be rename to stop referring to 'category'. Like this event.
But I guess it's ok to wait for the next PR and keep this one focused on the bug fix.

src/panel/poi/PoiPanel.jsx
```diff
@@ -180,7 +178,10 @@ export default class PoiPanel extends React.Component {
   backToList = () => {
     Telemetry.add(Telemetry.POI_BACKTOLIST);
     fire('restore_location');
-    window.app.navigateTo(`/places/?type=${this.props.sourceCategory}`);
+    const uri = this.props.resource.category
```
bbecquet, line 181: I'm not 100% sure `q` and `category` can't be present at the same time. And anyways, as we are going to more genericity of this feature, with more possible paramers, it would be safer to have a function that generates the query string based on all that is present in the `resource` object.

src/adapters/scene.js
G-Ray: I agree, I was not sure about how to name this value last name and went with `resource`, trying to keep the actual behavior first. I renamed to `poiFilters`

src/panel/poi/PoiPanel.jsx
G-Ray: Yep, I did this in `CategoryPanel.jsx ` already but missed this one, thanks

src/panel/category/CategoryPanel.jsx
```diff
@@ -90,16 +98,16 @@ export default class CategoryPanel extends React.Component {
...
+      query
     );
     this.setState({
       pois: places,
       dataSource: source,
       initialLoading: false,
     });
 
-    fire('add_category_markers', places, this.props.categoryName);
+    fire('add_category_markers', places, this.props.poiFilters.categoryName);
```
bbecquet, line 110: I think that's `category`, not `categoryName` here.

src/panel/category/CategoryPanel.jsx
```diff
@@ -112,7 +120,7 @@ export default class CategoryPanel extends React.Component {
           template: 'multiple',
           zone: 'list',
           element: 'phone',
-          category: this.props.categoryName,
+          category: this.props.poiFilters.categoryName,
```
bbecquet, line 123: Also here.

src/panel/poi/PoiPanel.jsx
G-Ray: ok, I added a simple loop to add params

src/panel/poi/PoiPanel.jsx
```diff
@@ -178,9 +180,32 @@ export default class PoiPanel extends React.Component {
...
+    const { poiFilters } = this.props;
+
     Telemetry.add(Telemetry.POI_BACKTOLIST);
     fire('restore_location');
-    window.app.navigateTo(`/places/?type=${this.props.sourceCategory}`);
+
+    let uri = '/places/';
+
+    // Reconstruct uri
+    for (const filterName of Object.keys(poiFilters)) {
```
bbecquet, line 191: That would be cleaner and easier to manage in a pure function, like `getFiltersAsQueryString`, which would return a string like `type=Foo&q=Bar`.
Or maybe better, adding a generic function in `url_utils` to add a set of query string parameters to any url, and just  renaming the keys in this file.

src/panel/poi/PoiPanel.jsx
```diff
@@ -178,9 +180,32 @@ export default class PoiPanel extends React.Component {
...
+      uri += uri[uri.length - 1] === '/' ? '?' : '&';
+
+      // TODO use keys as query param ?
+      if (filterName === 'category') {
+        uri += 'type=';
+      } else if (filterName === 'query') {
+        uri += 'q=';
+      }
+
+      uri += poiFilters[filterName];
```
bbecquet, line 205: By the way, there is a bug here, values which key isn't `category` or `query` will be just concat'd without their key.

src/panel/poi/PoiPanel.jsx
bbecquet: That's the perfect use case for `URLSearchParams`, which can be initialized with an object:
```js
const params = new URLSearchParams({type: 'restaurant' , q: 'McDo'});
console.log(params.toString()); // "type=restaurant&q=McDo"
```

G-Ray: yep, but I had issues doing this, will retry

bbecquet:
Could you add a description to the PR before merging please?

PR #599, opened 2020-04-06
QMAPS-1432 modify the mobile service panel

xem:
## Description
update the content and look of service mobile panel (direction icons first, no favourites button)

## Screenshots
image


src/panel/ServicePanelMobile.jsx
```diff
@@ -0,0 +1,104 @@
...
+      </div>
+
+      <hr/>
+
+      <h3>{_('Services nearby', 'service panel')}</h3>
+
+      <CategoryList className="service_panel__categories" />
+
+      {
+        nconf.get().events.enabled &&
```
G-Ray, line 79: Can we extract some of this pieces of code into component shared between desktop and mobile ?
I think we can at least share `service_panel__events`

src/panel/ServicePanel.jsx
```diff
@@ -1,69 +1,27 @@
...
-          onClick={() => { window.app.navigateTo('/favs'); }}>
-          <div className="service_panel__action__icon">
-            <span className="icon-icon_star"/>
-          </div>
-          <div className="service_panel__action__title">{ _('Favorites', 'service panel')}</div>
-        </button>
-
-      </div>
-    </Panel>;
+    if (isMobileDevice()) {
```
bbecquet, line 10: If you use this function for that, it won't re-render dynamically if the device changes (like it could happen for a tablet changing orientation).
For that, I introduced a React context in this PR https://github.com/QwantResearch/erdapfel/pull/520, which will update the component automatically if the device changes.
You can find how it's used:
 - here for a class component: https://github.com/QwantResearch/erdapfel/blob/master/src/panel/direction/DirectionPanel.jsx#L329
 - here for a functional component: https://github.com/QwantResearch/erdapfel/blob/master/src/panel/direction/Route.jsx#L11. It may be the occasion to convert your file to a functional component.

src/panel/ServicePanelMobile.jsx
```diff
@@ -0,0 +1,104 @@
...
+import CategoryList from 'src/components/CategoryList';
+import nconf from '@qwant/nconf-getter';
+
+const directionConf = nconf.get().direction;
+
+class ServicePanelMobile extends React.Component {
+  render() {
+    return <Panel
+      resizable
+      title=""
```
bbecquet, line 15: I think you can just omit it if it's empty.

src/panel/ServicePanelMobile.jsx
```diff
@@ -0,0 +1,104 @@
...
+        {
+          nconf.get().direction.enabled &&
+          <Fragment>
+            <button
+              onClick={() => { window.app.navigateTo('/routes/?mode=driving'); }}
+              className="service_panel__action service_panel__item__direction">
+              <div className="service_panel__action__icon">
+                <span className="icon-drive"/>
+              </div>
+              <div className="service_panel__action__title">{_('by car', 'service panel')}</div>
```
bbecquet, line 33: We need to include translations for these new strings in this PR.

src/scss/includes/panels/service_panel.scss
```diff
@@ -133,22 +134,19 @@
...
-    font-size: 23px;
-    line-height: 36px;
-  }
-}
-
 @media (max-width: 640px){
 
   .panel.service_panel {
-    height: 215px;
+    height: 235px;
```
bbecquet, line 140: At first I thought it was just a screenshot artefact but there is a white space at the bottom of the screen. I think this hardcoded height is wrong, as it doesn't take the reduced handle height into account.

src/panel/ServicePanelMobile.jsx
amatissart: Usually we wait for a branch to be merged to apply new translations separately (Travis will upload the new translation keys to Transifex after merging to master), although that's not ideal.

src/panel/ServicePanelMobile.jsx
```diff
@@ -0,0 +1,104 @@
...
+              </div>
+              <div className="service_panel__action__title">{_('on foot', 'service panel')}</div>
+            </button>
+            <button
+              onClick={() => { window.app.navigateTo('/routes/?mode=cycling'); }}
+              className="service_panel__action service_panel__item__direction">
+              <div className="service_panel__action__icon">
+                <span className="icon-bike"/>
+              </div>
+              <div className="service_panel__action__title">{_('on bike', 'service panel')}</div>
```
amatissart, line 65: "by bike" would probably be more correct ?

src/panel/ServicePanelMobile.jsx
xem: I vote for having the translations commits happening only on Master. o/

bbecquet: Ah yes, I always forget the process…
That means if we want to merge in master only code that can be deployed to production as soon as possible, we have to anticipate this. Merging some PR with new keys before?

src/panel/ServicePanelMobile.jsx
xem: yes. I googled the proverbs for each vehicle and thought the answer was "on" but "by" is indeed better

src/panel/ServicePanel.jsx
xem: > It may be the occasion to convert your file to a functional component.

I don't think so: in the near future desktop and mobile HTML/CSS will be reunited and won't need 2 separate subfiles like they do now, so there's no need to lose time on that, no?

bbecquet: As you wish. In this case there is no real consequence except code clarity. I personnaly think `useContext` is a simpler thing to use than a context consumer with is child function, but we may have different opinions.
As long as you're at ease with the code.

src/scss/includes/panels/service_panel.scss
```diff
@@ -133,22 +134,23 @@
...
+    }
+
+    .panel-content {
+      height: calc(100% - 26px);
+    }
   }
 
   .panel.service_panel .panel-resizeHandle {
-    min-height: 50px;
+    min-height: 25px;
```
amatissart, line 153: I think that this change makes the handle significantly harder to grab, and causes missed touches.

Also (I am not sure it's related), in somes cases the panel content will scroll during the unfold, which is unexpected.
Peek 08-04-2020 11-18

src/panel/ServicePanelMobile.jsx
```diff
@@ -0,0 +1,89 @@
...
+  render() {
+    return <Panel
+      resizable
+      minimizedTitle={_('Unfold to see quick actions', 'service panel')}
+      className="service_panel"
+      white
+    >
+      <div className="service_panel__actions">
+
+        <h3>{_('Itineraries', 'service panel')}</h3>
```
amatissart, line 26: wording: `itineraries` :arrow_right: `directions`

src/panel/ServicePanelMobile.jsx
```diff
@@ -0,0 +1,89 @@
...
+              && <button
+                onClick={() => {
+                  window.app.navigateTo('/routes/?mode=publicTransport');
+                }}
+                className="service_panel__action service_panel__item__direction">
+                <div className="service_panel__action__icon">
+                  <span className="icon-public-transport"/>
+                </div>
+                <div className="service_panel__action__title">
+                  {_('transport', 'service panel')}
```
amatissart, line 51: wording: `transport` :arrow_right: `transit`

src/scss/includes/panels/service_panel.scss
xem: I don't reproduce the weird scroll effect in any browser, pretty sure it's unrelated to a couple of sizes modified in CSS though :)
OK: I'll make the grab area bigger.

PR #600, opened 2020-04-07
Misc suggest simplifications

bbecquet:
## Description
Some simplications following the suggest dropdown migration to React. 
Mostly removes a lot of CSS due to over-specificity and cleans random bits from the old implementation.

## Why
Doing some clean-up before adding a new type of suggestion from NLU intentions.

### DIFF
--- a/src/components/ui/SuggestItem.jsx
+++ b/src/components/ui/SuggestItem.jsx
@@ -17,11 +17,14 @@ const GeolocationItem = () =>
   </div>;
 
 const CategoryItem = ({ category }) => {
-  const { label, alternativeName, color, backgroundColor } = category;
+  const { id, label, alternativeName, color, backgroundColor } = category;
   const icon = category.getIcon();
 
   return (
-    <div className="autocomplete_suggestion autocomplete_suggestion--category">
+    <div
+      className="autocomplete_suggestion autocomplete_suggestion--category"
+      data-id={id}
+    >
       <div
         style={{ color, backgroundColor }}
         className={`autocomplete-icon icon icon-${icon.iconClass}`}
--- a/src/components/ui/SuggestItem.jsx
+++ b/src/components/ui/SuggestItem.jsx
@@ -1,58 +1,51 @@
 /* global _ */
 import React from 'react';
-import classNames from 'classnames';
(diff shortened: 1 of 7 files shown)


G-Ray:
The "geolocation" entry is no longer skewed when hovered :+1:

src/components/ui/SuggestItem.jsx
```diff
@@ -66,24 +50,24 @@ const PoiItem = ({ poi, isHighlighted }) => {
 };
 
 const SeparatorLabel = ({ label }) =>
-  <h3 className="autocomplete_suggestion__category_title">
+  <h3 className="autocomplete_separator_label">
     {label}
   </h3>;
 
-const SuggestItem = ({ item, isHighlighted }) => {
```
G-Ray, line 73: oh, I left this in the last merge ? sorry

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -95,7 +93,6 @@ const SuggestsDropdown = ({
     <ul
       className={classnames('autocomplete_suggestions', className)}
       style={style}
-      ref={dropDownRef}
```
G-Ray, line 98: :+1: +1

src/components/ui/SuggestItem.jsx
```diff
@@ -1,61 +1,45 @@
...
+        className={`autocomplete-icon icon icon-${icon.iconClass}`}
       />
       <ItemLabels firstLabel={label} secondLabel={alternativeName} />
     </div>
   );
 };
 
-const PoiItem = ({ poi, isHighlighted }) => {
-  const { id, name, className, subClassName, type, alternativeName } = poi;
+const PoiItem = ({ poi }) => {
```
G-Ray, line 37: I think we can even go further merging PoiItem and CategoryItem, they rally don't seem visally différent . This can be done in another PR I guess

bbecquet: Maybe, but there are some differences in the icons, which are always a tricky part in the app because of the variety of icon fonts, sizes, etc.
As we'll have to introduce a new item type for NLU intentions, maybe further simplification will become obvious.

PR #604, opened 2020-04-10
CategoryPanel: Fetch data and fit on props updates

G-Ray:
Fetch Data when CategoryPanel has props updates (bbox or poiFilters.category)

### DIFF
--- a/src/panel/category/CategoryPanel.jsx
+++ b/src/panel/category/CategoryPanel.jsx
@@ -45,16 +45,19 @@ export default class CategoryPanel extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { bbox } = this.props;
-    const { category } = this.props.poiFilters;
+    const { bbox, poiFilters } = this.props;
 
     const panelContent = document.querySelector('.panel-content');
     if (panelContent) {
       panelContent.scrollTop = 0;
     }
 
-    if (category && category !== prevProps.poiFilters.category) {
-      this.fetchData();
+    // Check for a new, or changed poiFilter
+    for (const key in poiFilters) {
+      if (poiFilters[key] !== prevProps.poiFilters[key]) {
+        this.fetchData();
+        break;
+      }
     }
 
     if (bbox && bbox !== prevProps.bbox) {
--- a/src/panel/category/CategoryPanel.jsx
+++ b/src/panel/category/CategoryPanel.jsx
@@ -32,23 +32,30 @@ export default class CategoryPanel extends React.Component {
   }
 
   componentDidMount() {
-    const { category } = this.props.poiFilters;
-
     this.mapMoveHandler = listen('map_moveend', this.fetchData);
-
-    if (category) {
-      Telemetry.add(Telemetry.POI_CATEGORY_OPEN, null, null, { category });
-      const { label } = CategoryService.getCategoryByName(category);
-      SearchInput.setInputValue(label.charAt(0).toUpperCase() + label.slice(1));
-    }
-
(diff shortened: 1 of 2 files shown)


src/panel/category/CategoryPanel.jsx
```diff
@@ -32,24 +32,31 @@ export default class CategoryPanel extends React.Component {
   }
 
   componentDidMount() {
-    const { category } = this.props.poiFilters;
-
     this.mapMoveHandler = listen('map_moveend', this.fetchData);
-
-    if (category) {
```
bbecquet, line 39: As `componentDidUpdate` is not called after the first render, I'm not sure you can safely move that in it.

src/panel/category/CategoryPanel.jsx
```diff
@@ -32,24 +32,31 @@ export default class CategoryPanel extends React.Component {
...
+      if (category) {
+        Telemetry.add(Telemetry.POI_CATEGORY_OPEN, null, null, { category });
+        const { label } = CategoryService.getCategoryByName(category);
+        SearchInput.setInputValue(label.charAt(0).toUpperCase() + label.slice(1));
+      }
+      this.fetchData();
+    }
+
+    if (bbox !== prevProps.bbox) {
+      this.fitMapAndFetch();
```
bbecquet, line 58: You probably need to wrap it in a `window.execOnMapLoaded` call like it's done in the didMount to avoid errors or calls without effects if the user interacts with the panel while the map is still loading.
I know this is a mess, I have plans to remove that, but not yet.

G-Ray: ok, done

src/panel/category/CategoryPanel.jsx
G-Ray: Yes you 're right, and we surely telemetry to be send only on firt open

src/panel/category/CategoryPanel.jsx
```diff
@@ -45,11 +44,22 @@ export default class CategoryPanel extends React.Component {
...
+  componentDidUpdate(prevProps) {
+    const { bbox } = this.props;
+    const { category } = this.props.poiFilters;
+
     const panelContent = document.querySelector('.panel-content');
     if (panelContent) {
       panelContent.scrollTop = 0;
     }
+
+    if (category && category !== prevProps.poiFilters.category) {
```
bbecquet, line 56: If there is no category in the new props (so `category` is null), we never pass this test. This means if there was an active category before, the points won't be refreshed…
To anticipate for `query` and other fields of `poiFilters`, it's probably safer to compare `poiFilters` as a whole.

G-Ray: Ok I update by comparing if any key has changed instead

src/panel/category/CategoryPanel.jsx
```diff
@@ -45,11 +44,25 @@ export default class CategoryPanel extends React.Component {
...
+  componentDidUpdate(prevProps) {
+    const { bbox, poiFilters } = this.props;
+
     const panelContent = document.querySelector('.panel-content');
     if (panelContent) {
       panelContent.scrollTop = 0;
     }
+
+    // Check for a new, or changed poiFilter
+    for (const key in poiFilters) {
```
bbecquet, line 56: This assumes the same keys are present in both objects. It's the case for now just because there is a single render point for `CategoryPanel` which always pass `category` and `query`. But I'm quite sure in the future we'll have need for something more generic like a shallowCompare function. Let's not forget it then.

PR #606, opened 2020-04-14
[NLU] Display category(+place) suggestion from NLU instead of category 

bbecquet:
## Description
Use NLU in place of regexp matching to propose suggestions from detected intentions as first suggest item. First refactor some parts of the suggest/geocoder loop to introduce this new type of suggestion item.
 - separate the geocoder API calls from the `BragiPOI` class
 - move suggest data logic to the suggest_sources lib, where calls to data sources are done, instead of in the presentation component. 
 - use server-side intention detection in place of category regexp matching if the API returns some. Limit to intentions that can be mapped to a category (and optionaly a place).   
 - introduce a new `Intention` class and a new type of corresponding `SuggestItem`
 - when clicking on an intention suggestion, direct to the CategoryPanel with the detected category and optional bbox.

## Screenshots
Capture d’écran de 2020-04-14 18-22-19

### DIFF
--- a/src/adapters/geocoder.js
+++ b/src/adapters/geocoder.js
@@ -47,18 +47,17 @@ export function getGeocoderSuggestions(term, { lat, lon, zoom } = {}) {
       query.nlu = 'true';
     }
     suggestsPromise = ajax.get(geocoderConfig.url, query);
-    suggestsPromise.then(suggests => {
-      let ranking = 0;
-      const bragiResponse = suggests.features.map(feature => {
-        ranking += 1;
+    suggestsPromise.then(({ features, intentions }) => {
+      const pois = features.map((feature, index) => {
         const queryContext = new QueryContext(
           term,
-          ranking,
+          index + 1, // ranking
           query.lang,
           { lat, lon, zoom }
         );
         return new BragiPoi(feature, queryContext);
       });
+      const bragiResponse = { pois, intentions };
       bragiCache[cacheKey] = bragiResponse;
       resolve(bragiResponse);
     }).catch(error => {
--- a/src/adapters/geocoder.js
+++ b/src/adapters/geocoder.js
@@ -1,0 +1,78 @@
+import ajax from 'src/libs/ajax';
+import nconf from '@qwant/nconf-getter';
+import QueryContext from 'src/adapters/query_context';
+import BragiPoi from 'src/adapters/poi/bragi_poi';
+
+const serviceConfigs = nconf.get().services;
+const geocoderConfig = serviceConfigs.geocoder;
+const geocoderFocusPrecision = geocoderConfig.focusPrecision;
+
+if (!window.__bragiCache) {
+  window.__bragiCache = {};
+}
+
+function roundWithPrecision(value, precision) {
+  const rounded = Math.round(value * (1 / precision)) * precision;
+  return rounded.toFixed(3);
+}
+
+export function getGeocoderSuggestions(term, { lat, lon, zoom } = {}) {
+  let cacheKey = term;
+  if (lat !== undefined && lon !== undefined) {
+    lat = roundWithPrecision(lat, geocoderFocusPrecision);
+    lon = roundWithPrecision(lon, geocoderFocusPrecision);
+    cacheKey += `;${lat};${lon}`;
+  }
+  /* cache */
+  if (cacheKey in window.__bragiCache) {
+    const cachePromise = new Promise(resolve => {
+      resolve(window.__bragiCache[cacheKey]);
+    });
+    cachePromise.abort = () => {};
+    return cachePromise;
+  }
+  /* ajax */
+  let suggestsPromise;
+  const queryPromise = new Promise(async (resolve, reject) => {
+    const query = {
+      'q': term,
+      'limit': geocoderConfig.maxItems,
+    };
+    if (lat !== undefined && lon !== undefined) {
+      query.lat = lat;
+      query.lon = lon;
+    }
+    if (geocoderConfig.useLang) {
+      query.lang = window.getLang().code;
+    }
+    if (geocoderConfig.useNlu) {
+      query.nlu = 'true';
+    }
+    suggestsPromise = ajax.get(geocoderConfig.url, query);
+    suggestsPromise.then(suggests => {
+      let ranking = 0;
+      const bragiResponse = suggests.features.map(feature => {
+        ranking += 1;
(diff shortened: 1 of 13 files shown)


src/scss/includes/search_form.scss
```diff
@@ -182,6 +182,17 @@ input[type="search"] {
   }
 }
 
+.autocomplete_suggestion--intention {
+  text-transform: capitalize;
```
amatissart, line 186: Very minor, but on the second line at least I find this disturbing to read.
image

bbecquet: I agree, I hate this, we should juste capitalize the first letter of the whole string. I did that first but got entangled with tests and removed it. I'll restore it.

src/adapters/geocoder.js
```diff
@@ -0,0 +1,79 @@
...
+
+const serviceConfigs = nconf.get().services;
+const geocoderConfig = serviceConfigs.geocoder;
+const geocoderFocusPrecision = geocoderConfig.focusPrecision;
+
+const bragiCache = {};
+
+function roundWithPrecision(value, precision) {
+  const rounded = Math.round(value * (1 / precision)) * precision;
+  return rounded.toFixed(3);
```
G-Ray, line 15: ok, we can even use `digits` parameter with default value of 3

src/libs/url_utils.js
```diff
@@ -22,7 +22,7 @@ export function getMapHash(zoom, lat, lng) {
 export function parseQueryString(queryString) {
   const params = {};
   new URLSearchParams(queryString).forEach((value, key) => {
-    params[key] = value;
+    params[key] = decodeURIComponent(value);
```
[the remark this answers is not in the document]
bbecquet: Indeed, I don't remember why I thought I needed it. I'll remove it.

src/adapters/geocoder.js
G-Ray: also, maybe we should move this reusable function in a lib

bbecquet: Why not if we extract that to a shared function, here I just copy-pasted the most part of the bragi_poi file.

src/adapters/intention.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default class Intention {
+  constructor({ filter, description }) {
+    this.category = CategoryService.getCategoryByName(filter.category);
+    this.name = this.category ? this.category.getInputValue() : '';
+    this.bbox = filter.bbox;
+    this.place = description.place;
+  }
+
+  toQueryString = () => {
```
G-Ray, line 11: perfect time to reuse `buildQueryString` from `url_utils.js`. There is not much added value in this case though, but it's a start. However, please note `toQueryString` already return a `?` if there is at least one param

G-Ray: we ca also extends `buildQueryString` to ignore undefined values btw

src/ui_components/search_input.js
```diff
@@ -102,6 +103,9 @@ export default class SearchInput {
     } else if (selectedItem instanceof Category) {
       window.app.navigateTo(`/places/?type=${selectedItem.name}`,
         {}, { replace: replaceUrl });
+    } else if (selectedItem instanceof Intention) {
```
G-Ray, line 106: As the triggered action is always the same (calling `window.app.navigateTo`), I think we should just set two vars in each if statement : `url` and `options`, and call `window.app.navigateTo(url, options, { replace: replaceUrl });` at the end

src/adapters/intention.js
bbecquet: I did that :) https://github.com/QwantResearch/erdapfel/pull/606/commits/b690f19579f5ddfd73e16359639fc7c72f94f514

src/scss/includes/search_form.scss
bbecquet: I fixed it :+1:

src/scss/includes/search_form.scss
```diff
@@ -198,6 +205,10 @@ input[type="search"] {
       color: $secondary_text;
       font-size: 14px;
     }
+
+    &:first-letter {
+      text-transform: capitalize;
+    }
```
amatissart, line 212: I understand why we need this rule for intention suggestions.  
But for other types of suggestions, don't you think that the API response should be the source of truth for capitalization of place names, store brands, etc ?

bbecquet: Ok, I'll put it only on categories and suggestions then.

src/adapters/geocoder.js
```diff
@@ -0,0 +1,79 @@
...
+    if (lat !== undefined && lon !== undefined) {
+      query.lat = lat;
+      query.lon = lon;
+    }
+    if (geocoderConfig.useLang) {
+      query.lang = window.getLang().code;
+    }
+    if (geocoderConfig.useNlu) {
+      query.nlu = 'true';
+    }
```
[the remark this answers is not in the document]
bbecquet: Right, let's not hit our APIs uselessly. I can add a parameter quite easily.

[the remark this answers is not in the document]
bbecquet: I did it in this commit, I thought it was important enough to be in this PR https://github.com/QwantResearch/erdapfel/pull/606/commits/1e9a5cd510be4f4d31dbbf7a3e4a9f99d2e465a4

src/components/ui/SuggestItem.jsx
```diff
@@ -16,6 +17,18 @@ const GeolocationItem = () =>
...
 
+const IntentionItem = ({ intention }) => {
+  const { category, place } = intention;
+  const placeString = place
+    ? `${_('Close to')} ${place.properties.geocoding.name}`
+    : _('Search around this place');
+
+  return <div className="autocomplete_suggestion autocomplete_suggestion--intention">
+    <div className="autocomplete-icon" />
+    <ItemLabels firstLabel={category.name} secondLabel={placeString} />
```
amatissart, line 28: The category `name` is not localized here.
```suggestion
    <ItemLabels firstLabel={category.label} secondLabel={placeString} />
```

bbecquet: Damn it, once again! :roll_eyes: Thanks, I pushed your fix.

bbecquet: Damn it, once again! :roll_eyes: Thanks, I pushed your fix.

PR #608, opened 2020-04-15
Qmaps 1447 - animate geolocation and direction buttons on mobile

xem:
## Description
Make buttons disappear progressively during unfold and appear during fold

## Screenshots
boutons


src/adapters/scene.js
```diff
@@ -399,4 +407,30 @@ Scene.prototype.moveMobileGeolocationButton = function(bottom = 0) {
   }
 };
 
+Scene.prototype.mobileGeolocationButtonVisibility = function(visible) {
+  if (isMobileDevice()) {
```
G-Ray, line 411: I would advice to use an early return instead, as it is a pre-condition to execute the rest of this function
```suggestion
  if (!isMobileDevice()) {
    return 
  }
```

src/adapters/scene.js
```diff
@@ -399,4 +407,30 @@ Scene.prototype.moveMobileGeolocationButton = function(bottom = 0) {
...
+      if (visible) {
+        item.classList.remove('hidden');
+      } else {
+        item.classList.add('hidden');
+      }
+    }
+  }
+};
+
+Scene.prototype.mobileDirectionButtonVisibility = function(visible) {
```
G-Ray, line 423: `mobileDirectionButtonVisibility` and `mobileGeolocationButtonVisibility` are nearly identical, can we define and reuse a more generic function like `setMobileDirectionButtonVisibility(selector, isVisible)` ?

src/scss/includes/direction.scss
```diff
@@ -817,18 +819,18 @@ input:valid:focus ~ .itinerary__field__clear {
...
-    background: #fff;
     border-radius: 50%;
     box-shadow: rgba(0,0,0,.2) 0 2px 12px 0;
     position: fixed;
     bottom: 66px;
     right: 9px;
-    color: #5c6f84;
     font-size: 25px;
+    opacity: 1;
+    transition: opacity .3s;
```
bbecquet, line 833: You used 0.2s as duration on the other button. If there is no reason for this difference it should be the same value.

src/components/ui/Panel.jsx
```diff
@@ -66,6 +66,9 @@ export default class Panel extends React.Component {
 
   componentDidMount() {
     this.updateMobileMapUI();
+    this.setState({
```
bbecquet, line 69: This isn't used anymore in this implementation.

bbecquet, line 69: It's generally not a good idea to modify the state synchronously in `componentDidMount` as it will trigger immediately a new render after the first one.
See also next comment.

src/components/ui/Panel.jsx
```diff
@@ -83,6 +86,20 @@ export default class Panel extends React.Component {
       window.execOnMapLoaded(() => {
         fire('move_mobile_bottom_ui', height);
       });
+
+      if (height > this.state.defaultHeight) {
```
bbecquet, line 90: If don't think this is realiable as the Panel can be passed an `initialSize` prop to render it first minimized or maximized. If the initial size is minized and you swipe up, this first test will be considered true, hiding both buttons whereas we are transitioning to default.
Instead of this `defaultHeight` set at first mount, we should set a value when the transtion starts and compare it during the swipe. I think this is exactly what's already done in `holdResizer`, which stores this value in `this.startHeight`, as we need it for the algo controling the Panel resize.

src/adapters/scene.js
xem: sure, but in the function only, the 4 events remain the same to avoid copying the css selector in every "fire" :)
thanks

PR #610, opened 2020-04-15
Fix map_control_group buttons hover on mobile

G-Ray:
Geoloc, itinerary and rotation buttons were subject to a misbehavior of the `hover` effect on mobile: When long tapping on a button, an `hover` style remained, even after raising the finger from the screen.

This PR fixes this behavior by setting the necessary styles on a `hover + active` state. So the finger needs to press the button for this style to be effective.

We can go further using touch events, because this css only solution will "disable" the button if the finger moves on the buttons by some pixels. What do you think ?

After

### DIFF
--- a/src/scss/includes/direction.scss
+++ b/src/scss/includes/direction.scss
@@ -833,12 +833,6 @@ input:valid:focus ~ .itinerary__field__clear {
     /* Attempt to fix rare layout bug on iOS */
     transform: translate3d(0,0,0);
 
-    &:hover {
-      background: #353c52;
-      color: #fff;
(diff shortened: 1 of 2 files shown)


src/scss/includes/direction.scss
```diff
@@ -833,12 +833,6 @@ input:valid:focus ~ .itinerary__field__clear {
     /* Attempt to fix rare layout bug on iOS */
     transform: translate3d(0,0,0);
 
-    &:hover {
```
G-Ray, line 836: Let me know if you think this is necessary

PR #612, opened 2020-04-17
Set poi zoom level to 16.5

G-Ray:
Set zoom levels to `16.5` pois of types: `street`, `house` and `poi`.
I let the `DEFAULT_ZOOM` to 16, let me know if we should set it to 16.5 too.


bbecquet:
> I let the DEFAULT_ZOOM to 16, let me know if we should set it to 16.5 too.

Yeah maybe we could just have a single DEFAULT_ZOOM to 16.5, that's cleaner and simpler and would avoid strange zoom in/out actions for the user.

PR #613, opened 2020-04-17
[Fix] Make sure category label is always removed from search input

bbecquet:
## Description
Mini bug fix: ensure we always remove the category label that is displayed in the search input field when we leave the category panel.
It was only removed when the panel was closed with the cross button, but it can be left also by clicking on the map.


G-Ray:
tested and approved :+1:

PR #616, opened 2020-04-20
Hide direction button on mobile when a poi is selected

G-Ray:
Call 
```js
fire('mobile_direction_button_visibility', false);
```
when a PoiPanel is mounted in order to hide the itinerary button when a poi is selected

### DIFF
--- a/src/panel/poi/PoiPanel.jsx
+++ b/src/panel/poi/PoiPanel.jsx
@@ -61,6 +61,7 @@ export default class PoiPanel extends React.Component {
   }
 
   componentDidMount() {
+    fire('mobile_direction_button_visibility', false);
     this.loadPoi();
     this.storeAddHandler = listen('poi_added_to_favs', poi => {
       if (poi === this.state.fullPoi) {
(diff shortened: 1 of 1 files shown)


src/panel/poi/PoiPanel.jsx
```diff
@@ -61,6 +61,7 @@ export default class PoiPanel extends React.Component {
   }
 
   componentDidMount() {
+    fire('mobile_direction_button_visibility', false);
```
bbecquet, line 64: I expected we'd restore it in `componentWillUnmount`. Is it unnecessary? If so, it looks like a side-effect.

G-Ray: it should not be necessary because the panel is resized when we quit a POI panel, hence the button animation is triggered.
However you are right it should probably be safer to trigger button visibility in componentDIdMount in case this behavior change in the futur.

G-Ray: I update the code, thanks

PR #621, opened 2020-04-24
[Misc clean-up] Don't attach custom event functions to window namespace

bbecquet:
## Description
A piece of tech clean-up I wanted to do for a long time.
We declare functions to fire/listen to custom events, but they are declared as globals, attached to the `window` object.
There is no need to do that. This pollutes the global namespace and requires special ESLint config, whereas JS modules provide the right way to import that as needed.
Also, if we need to do server side rendering with our React components, any use of `window` will break.


src/libs/customEvents.js
```diff
@@ -0,0 +1,25 @@
+
+function fire(name, params, ...additionalParams) {
```
G-Ray, line 2: I think we can remove `additionalParams`, I don't see why `params` would not be enough

src/libs/listen.js
```diff
@@ -1,26 +0,0 @@
-/**
```
G-Ray, line 1: :+1:

src/libs/customEvents.js
bbecquet: Indeed. I just copied the previous code but this is worth simplifying. Thank you!

amatissart:
A minor conflict needs to be resolved on this PR.

PR #623, opened 2020-04-27
Update covid en strings

G-Ray:
containment -> lockdown
schedules -> opening hours
opened -> open

### DIFF
--- a/src/panel/poi/blocks/Covid19.jsx
+++ b/src/panel/poi/blocks/Covid19.jsx
@@ -49,7 +49,7 @@ const getContent = ({ status, opening_hours, note, contribute_url }) => {
       </div>}
       {!schedule &&
         <div className="covid19-changeWarning">
-          {_('Schedules subject to change', 'covid19')}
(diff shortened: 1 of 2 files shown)


src/panel/poi/blocks/Covid19.jsx
```diff
@@ -76,7 +76,7 @@ const LocalizedWarning = ({ countryCode }) => {
     'FR': (
       <div>
         <p>
-          {_('During the entire period of containment, movement to this location is only permitted with an overriding movement certificate.', 'covid19')}
+          {_('During the entire period of lockdown, movement to this location is only permitted with an overriding movement certificate.', 'covid19')}
```
bbecquet, line 79: Apparently, instead of `overriding`, the good term is `derogatory`.
I'm not sure about `movement` either…

G-Ray: Ok, I updated the string, can you review one more time please ?

PR #624, opened 2020-04-28
Fix suggest list reappearing after fast submit

amatissart:
## Description
* Show the suggestions list only if the input field is still focused
* Remove obsolete mechanism related to pending requests in `Suggest`, and other unused options for the `Autocomplete` object

## Why
When the user used the Enter key quickly after typing a query, the suggestions list could re-appear after a result had been selected and the search input had been blurred. 

First I thought this problem was related to a pending request not being correctly aborted on submit. But I realized that some recent refactoring of the Suggest made this mechanism obsolete in such a case: the `pending` flag is never reset to `false`.

### DIFF
--- a/src/vendors/autocomplete.js
+++ b/src/vendors/autocomplete.js
@@ -30,10 +30,6 @@ export default function autoComplete(options) {
     delay: 150,
     // Takes as arguments: items, search
     renderItems: function() {},
-    // Takes as arguments: e, term, item, items
-    onSelect: function() {},
-    // Takes as arguments: e, term, items
-    onUpdate: function() {},
   };
   for (const k in options) {
     if (options.hasOwnProperty(k)) {
--- a/src/vendors/autocomplete.js
+++ b/src/vendors/autocomplete.js
@@ -34,8 +34,6 @@ export default function autoComplete(options) {
     onSelect: function() {},
     // Takes as arguments: e, term, items
     onUpdate: function() {},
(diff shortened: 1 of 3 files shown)


src/vendors/autocomplete.js
```diff
@@ -91,7 +80,7 @@ export default function autoComplete(options) {
             that.sourcePending = o.source(val);
             that.sourcePending.then(source => {
               that.sourcePending = null;
-              if (source !== null) {
+              if (source !== null && document.activeElement === that) {
                 suggest(source);
```
amatissart, line 84: Here is the delicate change.  
Do you see any situation / platform where it would be expected to show the suggest list while the input field would not be the active element ?

bbecquet: Actually, this is something the design asked to do in future iterations on mobile… The virtual keyboard takes a lot of screen space when the field is focused, so they proposed to be able to scroll the list in full screen… We wanted to finish the component refacto first, before looking for an elegant solution to do that.

bbecquet: In any case, I think we'll implement that differently, as this autocomplete lib is supposed to disappear entirely. So any clean-up that can happen before is welcome :)

G-Ray: a similar fix is on line line 77 for firefox on android. I'm not aware of any situation where it should be an issue

PR #626, opened 2020-04-28
QMAPS-1469 don't list category POIs hidden by the Maps UI

xem:
## Description
Fetch POIs in a bbox that excludes the left panel and top header on desktop

## Screenshots
image

### DIFF
--- a/src/panel/category/CategoryPanel.jsx
+++ b/src/panel/category/CategoryPanel.jsx
@@ -11,7 +11,7 @@ import SearchInput from 'src/ui_components/search_input';
 import nconf from '@qwant/nconf-getter';
 import IdunnPoi from 'src/adapters/poi/idunn_poi';
 import CategoryService from 'src/adapters/category_service';
-import { getMapPaddings } from 'src/panel/layouts';
+import { getVisibleBbox } from 'src/panel/layouts';
 import { isMobileDevice } from '../../libs/device';
 
 const categoryConfig = nconf.get().category;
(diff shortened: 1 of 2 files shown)


src/panel/category/CategoryPanel.jsx
```diff
@@ -103,8 +105,23 @@ export default class CategoryPanel extends React.Component {
 
   fetchData = async () => {
     const { category, query } = this.props.poiFilters;
-
     const bbox = window.map.mb.getBounds();
+
+    // On desktop, fetch PoIs on a bbox that doesn't include the list panel's width and the header's height
+    if (!isMobileDevice()) {
```
bbecquet, line 111: Shouldn't we do the same on mobile by taking the top bar and panel into account?

src/panel/category/CategoryPanel.jsx
```diff
@@ -103,8 +105,23 @@ export default class CategoryPanel extends React.Component {
...
+      let ne = bbox.getNorthEast();
+      let sw = bbox.getSouthWest();
+      const ne_canvas = window.map.mb.project(ne);
```
bbecquet, line 114: It would be nice to put all these low level map-dependant operations somewhere else. What about a dedicated function in the `layouts` module, which would take the map instance as parameter so it doesn't create a hard dependency with `window`.

xem: Ok, I've put it in layouts. However, I don't think I can compute a reduced bbox without the Mapbox object (window.map.mb), so I had to keep it.

src/panel/layouts.js
```diff
@@ -25,6 +26,31 @@ export function getMapPaddings({ isMobile, isDirectionsActive }) {
...
+  const bbox = window.map.mb.getBounds();
+  let ne = bbox.getNorthEast();
+  let sw = bbox.getSouthWest();
+  const ne_canvas = window.map.mb.project(ne);
+  const sw_canvas = window.map.mb.project(sw);
+
+  if (isMobileDevice()) {
+    // On mobile, compute a bbox that excludes the resizable panel height and the header's height
+    ne_canvas.y += 65;
+    sw_canvas.y -= window.innerHeight / 2;
```
amatissart, line 40: Are we sure that the bbox should completely exclude the bottom half of the screen, including when the panel is reduced ?

xem: Well, it's debatable indeed: on principle yes, but it fits in a slightly larger edge case:

When we open the category, the panel is half-open and we want the request to target the upper half of the map.

If we move the panel to the bottom, what should happen? a new request with the updated bbox?

If we move the map with the panel folded at the bottom, should we recompute the list with the big bbox?

If we move the panel from bottom to the half of the screen, should we recompute the results with the small bbox?

If the answer to these questions is yes, then we can indeed take the panel folding into account in the bbox computation, but we also need to trigger new search requests when the panel is resized, which imo could look confusing, but we can try.

And if we decide to do it, I wonder what is the best way to "measure"  the panel height and send it to getVisibleBbox. maybe @bbecquet has an idea to do it "clean" ?

xem: We'll only exclude the header from the search bbox finally :)

src/panel/layouts.js
```diff
@@ -25,6 +26,30 @@ export function getMapPaddings({ isMobile, isDirectionsActive }) {
   return MOBILE_CARD;
 }
 
+export function getVisibleBbox() {
```
bbecquet, line 29: By passing the MapBoxGL map instance as parameter to this function, instead of using it multiple times as `window.map.mb`, you make this function decoupled from the `scene.js` implementation and from global variables, which is always a good thing. It's easier to maintain and to unit test in isolation.

PR #629, opened 2020-04-29
Refactor suggest

G-Ray:
## Description
Refactor `suggests.js` and `autocomplete.js`
convert autocomplete to a class

### DIFF
--- a/src/adapters/suggest.js
+++ b/src/adapters/suggest.js
@@ -1,96 +1,115 @@
 /* global _ */
 import React from 'react';
 import ReactDOM from 'react-dom';
-import Autocomplete from '../vendors/autocomplete';
 import nconf from '@qwant/nconf-getter';
-import PoiStore from './poi/poi_store';
 import NavigatorGeolocalisationPoi from 'src/adapters/poi/specials/navigator_geolocalisation_poi';
+
+import PoiStore from './poi/poi_store';
+import Autocomplete from '../vendors/autocomplete';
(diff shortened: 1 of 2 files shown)


src/vendors/autocomplete.js
```diff
@@ -1,140 +1,108 @@
...
+        .then((source) => {
+          if (source !== null && document.activeElement === this.input) {
+            this.suggest(source);
+          }
+        })
+        .catch((e) => {
+          console.warn(e); /* should be handled by a telemetry logger */
+        });
+
+      this.sourcePending = null;
```
amatissart, line 63: :warning: This is probably not equivalent to the previous behavior. 
`sourcePending` should not be set to `null` before the promise resolved (and the HTTP response has been received).

src/adapters/suggest.js
```diff
@@ -1,97 +1,116 @@
...
-          , reactElem
-        );
-      },
+      source: term => this.suggestResults(term),
+      renderItems: items => this.renderItems(items),
     });
 
-    this.searchInputDomHandler.addEventListener('blur', function handleSearchInputBlur() {
-      unmountReactSuggestDropdown();
+    this.searchInputDomHandler.addEventListener('blur', () => {
```
bbecquet, line 42: I just realized we never removed this listener… So we probably have always had memleaks when we open/close the direction panel, as we reinstantiate suggests multiple times. Hopefully the next refacto steps will make it easier to solve that.

PR #631, opened 2020-05-04
Suggest component

G-Ray:
* Implemented a `Suggest.jsx` component, which use `SuggestsDropdown`
* Replaced top search bar autocomplete logic with Suggest
* Replaced DirectionInput autocomplete logic with Suggest
* Remove good old suggest.js and autocomplete.js


src/components/ui/Suggest.jsx
```diff
@@ -0,0 +1,134 @@
...
+  useEffect(() => {
+    const handleSubmit = () => {
+      selectItem(items[0], true);
+    };
+
+    const handleKeyDown = async event => {
+      if (event.key === 'Esc' || event.key === 'Escape') { // esc
+        setIsOpen(false);
+        setIsHighlighted(false);
+      } else if (event.key === 'Enter' && !isHighlighted) {
```
bbecquet, line 81: `SuggestsDropdown` watches for the 'Enter' key too. There is no reason to have both. I think this should be done in this component only, and the dropdown should now just manage the arrow keys and mouse interaction.

Another way is to merge `Suggest` and `SuggestsDropdown` in a single component, as they can't exist separately anyway.

src/components/ui/Suggest.jsx
```diff
@@ -0,0 +1,134 @@
...
+        })
+        .catch(() => { /* Query aborted. Just ignore silently */ });
+    }, 100);
+
+    searchInputDomHandler.addEventListener('focus', handleFocus);
+    searchInputDomHandler.addEventListener('blur', handleBlur);
+    searchInputDomHandler.addEventListener('input', handleInput);
+
+    return () => {
+      searchInputDomHandler.removeEventListener('blur', handleFocus);
```
bbecquet, line 66: This should be `'focus'` here.

src/components/ui/Suggest.jsx
```diff
@@ -0,0 +1,134 @@
...
+      onHighlight={item => {
+        if (!item) {
+          setIsHighlighted(false);
+          searchInputDomHandler.value = lastQuery;
+        } else {
+          setIsHighlighted(true);
+          searchInputDomHandler.value = item.name;
+        }
+      }}
+      onSelect={item => {
```
bbecquet, line 124: You do a lot of things here that are done almost identically elsewhere. What do you think about defining a function like `onItemSelected(item)` and try to reuse it instead of doing everything here anonymously.

src/components/ui/Suggest.jsx
```diff
@@ -0,0 +1,134 @@
...
+const SUGGEST_MAX_ITEMS = geocoderConfig.maxItems;
+const SUGGEST_USE_FOCUS = geocoderConfig.useFocus;
+const SUGGEST_FOCUS_MIN_ZOOM = 11;
+
+const Suggest = ({ withCategories, withGeoloc }) => {
+  const [items, setItems] = useState([]);
+  const [isOpen, setIsOpen] = useState(false);
+  const [lastQuery, setLastQuery] = useState('');
+  const [isPending, setIsPending] = useState('unset');
+  const [isPendingSubmit, setIsPendingSubmit] = useState(false);
```
bbecquet, line 19: The point of `isPendingSubmit` in the lifecycle isn't clear to me. Maybe it's a naming problem. If you need to track a kind of status, I'd expect a simple `status` state, with possible values like 'fetching' and 'ready'.
If `isPendingSubmit` is to implement the behavior of submit validation while there is still a query pending (what was called 'preselect' before I think), it could probably be managed in a simpler way by having an async function providing the most up-to-date data, optionnaly waiting for the query result, called when submitting.

tests/integration/tests/autocomplete.js
```diff
@@ -228,20 +228,22 @@ test('submit key', async () => {
...
   await page.click('#clear_button');
 
   /* force specific query */
   responseHandler.addPreparedResponse(mockAutocompleteAllTypes, /autocomplete\?q=paris/);
-  await page.keyboard.type('paris');
+  await autocompleteHelper.typeAndWait('paris');
   await page.keyboard.press('Enter');
   await page.waitForSelector('.autocomplete_suggestions', { hidden: true });
 
+  await page.waitFor(300);
```
bbecquet, line 245: Do you have idea why it's needed in the new implementation? Does it mean the React component is less reactive to user inputs?

src/components/ui/Suggest.jsx
```diff
@@ -0,0 +1,139 @@
...
+        if (!item) {
+          setIsHighlighted(false);
+          searchInputDomHandler.value = lastQuery;
+        } else {
+          setIsHighlighted(true);
+          searchInputDomHandler.value = item.name;
+        }
+      }}
+      onSelect={item => {
+        window.__searchInput.searchInputHandle.value = item.name;
```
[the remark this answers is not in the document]
G-Ray: yes, thanks

tests/integration/tests/autocomplete.js
G-Ray: I have no idea, I don't understand this behavior meanwhile the first `expect` in this test pass.

src/components/ui/Suggest.jsx
G-Ray: Well, I would like but there are only two "selection" cases: With "onSelect" called by `SuggestDropdown` component, or a "selection" using the `Enter` key. Enter key is a special case because we have to wait until the query is resolved. I don't find a proper way to merge these two cases.

src/components/ui/Suggest.jsx
G-Ray: Great news! now that I re-enabled the submit button, this logic is not needed anymore, we just need to close the dropdown when Enter key is pressed.

src/panel/direction/DirectionInput.jsx
```diff
@@ -16,34 +17,36 @@ class DirectionInput extends React.Component {
...
   onChange = event => {
     const value = event.target.value;
     this.props.onChangePoint(value, null);
   }
 
-  onKeyPress = event => {
+  onKeyPress = async event => {
     if (event.key === 'Enter' && this.props.value !== '') {
-      this.suggest.onSubmit();
+      const items = await fetchSuggests(this.props.value);
```
bbecquet, line 38: I know it's related to how the Enter key behavior, but it doesn't seem right to have to do that here, whereas it's already done in the `Suggest` itself :/
And there is probably a bug also, as `fetchSuggests` include categories, which shouldn't be the case in the direction panel.

How can we solve that? Maybe managing the Enter key from outside wasn't the good way after all.

G-Ray: withCategories is now set to false by default, thanks.

About the Enter key, after discussion we opted to let this PR as this, and think about another solution to implement later. I'm open to any idea.

PR #634, opened 2020-05-06
QMAPS-1522 display city names with postal code when restoring POI (in itinerary)

xem:
## Description
display city names with postal code when restoring POI (in itinerary)

## Why
the postal codes disappear from the url and fields on refresh

## Screenshot
image

### DIFF
--- a/src/adapters/poi/idunn_poi.js
+++ b/src/adapters/poi/idunn_poi.js
@@ -46,6 +46,14 @@ export default class IdunnPoi extends Poi {
     case 'address':
     case 'street':
       return this.alternativeName;
+    case 'admin':
+      if (this.address && this.address.label) {
+        const split = this.address.label.split(',');
+        if (split[0]) {
+          return split[0];
+        }
+      }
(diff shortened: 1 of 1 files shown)


src/adapters/poi/idunn_poi.js
```diff
@@ -46,6 +46,14 @@ export default class IdunnPoi extends Poi {
     case 'address':
     case 'street':
       return this.alternativeName;
+    case 'admin':
+      if (this.address && this.address.label) {
```
bbecquet, line 50: Using the new optionnal fields syntax, you could replace all these lines with:
```
return this.address?.label?.split(',')[0] || this.name;
```

xem: Just because I could doesn't mean I should :D 
(but I guess it's better indeed)

bbecquet: Yep, up to you, that's why I validated the PR anyway :)

PR #642, opened 2020-05-12
[Fix] Disappearing pois

bbecquet:
## Description
Fixes the markers disappearing when the same category or full-text query was requested from the category panel view.
The reason was the event to remove markers was thrown, but as the props of CategoryPanel didn't change, it wasn't re-render and the markers were never re-added. Now we don't remove markers at all if we detect we come from the CategoryPanel to the CategoryPanel. Markers will be refreshed anyway if the query params have changed.

Also fixes another bug, somewhat related. The markers from a full-text query behaved differently from those of a category: they were removed if clicking on a single one. Now they will act the same: the clicked marker will be emphasized, and the other markers will be kept on the map.

### DIFF
--- a/src/libs/object.js
+++ b/src/libs/object.js
@@ -1,0 +1,1 @@
+export const isNullOrEmpty = obj => !obj || Object.keys(obj).length === 0;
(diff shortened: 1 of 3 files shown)


src/libs/object.js
```diff
@@ -0,0 +1 @@
+export const isNullOrEmpty = obj => !obj || Object.keys(obj).length === 0;
```
G-Ray, line 1: :+1

PR #643, opened 2020-05-13
QMAPS-1336: make the itinerary suggest fullscreen on mobile

xem:
## Description
make the itinerary suggest fullscreen on mobile

## Screenshots
image
image

### DIFF
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -305,7 +305,7 @@ input[type="search"] {
     top: 0;
     height: calc(100vh - 69px);
     max-height: none;
-    mrgin-top: -1px;
+    margin-top: -1px;
     overflow: auto;
   }
 
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -303,9 +303,10 @@ input[type="search"] {
   .autocomplete_suggestions {
     position: fixed;
     top: 0;
-    height: 100%;
+    height: calc(100vh - 69px);
     max-height: none;
-    mrgin: 0;
+    mrgin-top: -1px;
+    overflow: auto;
   }
 
   .search_form__logo {
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -302,8 +302,10 @@ input[type="search"] {
 
   .autocomplete_suggestions {
     position: fixed;
-    max-height: calc(100vh - 120px);
-    margin-top: -1px;
+    top: 0;
+    height: 100%;
+    max-height: none;
+    mrgin: 0;
   }
 
   .search_form__logo {
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -303,7 +303,7 @@ input[type="search"] {
   .autocomplete_suggestions {
     position: fixed;
     top: 0;
-    height: calc(100vh - 69px);
+    bottom: 0;
     max-height: none;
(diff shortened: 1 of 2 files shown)


src/scss/includes/search_form.scss
```diff
@@ -302,8 +302,10 @@ input[type="search"] {
 
   .autocomplete_suggestions {
     position: fixed;
-    max-height: calc(100vh - 120px);
-    margin-top: -1px;
+    top: 0;
+    height: 100%;
+    max-height: none;
+    mrgin: 0;
```
G-Ray, line 308: typo

G-Ray:
Icons are not centered in the input anymore. Screenshot from master as reference:

Screenshot from 2020-05-13 17-35-28

amatissart:
In this situation, it's now impossible to scroll or select the last items in the list without closing the keyboard. Is it expected ?

xem:
No, it's just that I couldn't test on a real phone yet :)

src/scss/includes/search_form.scss
```diff
@@ -302,8 +302,11 @@ input[type="search"] {
 
   .autocomplete_suggestions {
     position: fixed;
-    max-height: calc(100vh - 120px);
+    top: 0;
+    height: calc(100vh - 69px);
```
bbecquet, line 306: This `69px` seems out of nowhere (yes, 120px was too :)). And it doesn't seem right also, I see some pixels of space at the bottom:
Capture d’écran de 2020-05-15 15-21-49

bbecquet: Actually, you can just replace this rule by `bottom: 0;`, and you won't need any magic number :)

src/scss/includes/search_form.scss
```diff
@@ -302,8 +302,11 @@ input[type="search"] {
 
   .autocomplete_suggestions {
```
bbecquet, line 303: Using this selector, also affects the main search bar suggest. It may be done later, but I think that's not what we want in this PR and issue.

xem: it's ok, we decided to enable fullscreen suggests everywhere within this issue

src/scss/includes/search_form.scss
bbecquet: You resolved this comment without actually addressing it.

xem: sorry, my fix the push didn't work :( 
That and my laptop that crashed with a black screen of death earlier, no doubt, it's monday :D

src/scss/includes/search_form.scss
```diff
@@ -351,11 +358,16 @@ input[type="search"] {
...
-      left: 16px;
+      display: none;
+    }
+
+    .main_field_return {
+      display: block;
+      left: 10px;
+      top: 20px;
+      font-size: 25px;
+      color: #353C52;
```
bbecquet, line 369: Can you use the `$primary_text` variable here? It'll make it easier to keep things consistent during the redesign.
(and I'll probably make a PR to generalize the use of color vars instead of hard-coded value).

src/scss/includes/search_form.scss
```diff
@@ -314,6 +317,10 @@ input[type="search"] {
     background-size: 34px 34px;
   }
 
+  .main_field_return {
+    display: none;
```
bbecquet, line 321: This rule should be outside of the media query, as here it breaks the desktop design.
Capture d’écran de 2020-05-22 11-28-32

PR #644, opened 2020-05-13
Display suggestions in Direction panel

G-Ray:
- Display suggestions inside DirectionPanel
- Use props `inputNode` (input to fetch suggests from) and `ouputNode` to display suggestions results. If no outputNode is provided, an absolute position will be computed to keep the existing behavior (like in top search bar)
- outputNode will be rendered thanks to `ReactDOM.createPortal`

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -1,10 +1,11 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useContext } from 'react';
 import ReactDOM from 'react-dom';
 import debounce from 'lodash.debounce';
 import { bool, string, func, object } from 'prop-types';
 
 import SuggestsDropdown from 'src/components/ui/SuggestsDropdown';
 import { fetchSuggests, selectItem, modifyList } from 'src/libs/suggest';
+import { DeviceContext } from 'src/libs/device';
 
 const SUGGEST_DEBOUNCE_WAIT = 100;
 
@@ -19,6 +20,7 @@ const Suggest = ({
   const [items, setItems] = useState([]);
   const [isOpen, setIsOpen] = useState(false);
   const [lastQuery, setLastQuery] = useState('');
+  const isMobile = useContext(DeviceContext);
   let currentQuery = null;
 
   useEffect(() => {
@@ -28,6 +30,10 @@ const Suggest = ({
         fetchItems('');
       } else {
         inputNode.select();
+        if (isMobile) {
+          setIsOpen(true);
+          fetchItems(inputNode.value);
+        }
       }
     };
 
@@ -66,7 +72,7 @@ const Suggest = ({
 
     const handleKeyDown = async event => {
       if (event.key === 'Esc' || event.key === 'Escape') {
-        if (inputNode.value === '') {
+        if (inputNode.value === '' && !isMobile) {
           setIsOpen(false);
         } else {
           inputNode.value = '';
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -66,12 +66,9 @@ const Suggest = ({
 
     const handleKeyDown = async event => {
       if (event.key === 'Esc' || event.key === 'Escape') {
-        if (inputNode.value === '') {
-          setIsOpen(false);
-        } else {
-          inputNode.value = '';
-          fetchItems('');
-        }
+        inputNode.value = '';
+        fetchItems('');
+        setIsOpen(true);
       }
     };
 
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -60,6 +60,7 @@ const Suggest = ({
       const { value } = e.target;
       if (value !== lastQuery) {
         fetchItems(e.target.value);
+        setIsOpen(true);
       }
     };
 
@@ -71,8 +72,6 @@ const Suggest = ({
           inputNode.value = '';
           fetchItems('');
         }
-      } else {
-        setIsOpen(true);
       }
     };
 
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -15,6 +15,7 @@ const Suggest = ({
   withCategories,
   withGeoloc,
   onSelect = selectItem,
+  onClear,
   className,
 }) => {
   const [items, setItems] = useState([]);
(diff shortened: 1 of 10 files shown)


src/panel/direction/DirectionForm.jsx
```diff
@@ -56,7 +56,17 @@ export default class DirectionForm extends React.Component {
       destinationInputText,
     } = this.props;
 
+    const isMobile = isMobileDevice();
+
+    const VehicleSelectorComponent = props => <VehicleSelector
```
bbecquet, line 61: I don't see the point of declaring a function here. Why not using `VehicleSelector` directly?

src/panel/direction/DirectionForm.jsx
```diff
@@ -56,7 +56,17 @@ export default class DirectionForm extends React.Component {
...
+      vehicles={vehicles}
+      activeVehicle={activeVehicle}
+      onSelectVehicle={onSelectVehicle}
+      {...props}
+    />;
+
     return <div className="itinerary_form">
+      {!isMobile && <VehicleSelectorComponent />}
```
bbecquet, line 69: It it's only a display order which changes, you can keep exactly the same DOM and use CSS.
For example, `{display: flex; flex-direction: column-reverse}` on the container will do the trick.

src/panel/direction/DirectionInput.jsx
```diff
@@ -80,6 +81,7 @@ class DirectionInput extends React.Component {
   render() {
     const { pointType, inputRef, isLoading } = this.props;
     const { mounted, readOnly } = this.state;
+    const isMobile = isMobileDevice();
```
bbecquet, line 84: Same component as before. For this to be reactive to screen size changes, you have to use the context, not the function directly. An example, on another class component https://github.com/QwantResearch/erdapfel/blob/master/src/panel/poi/PoiPanel.jsx#L332

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -5,26 +5,32 @@ import { object, func, string, arrayOf } from 'prop-types';
 import SuggestItem from './SuggestItem';
 
 const SuggestsDropdown = ({
+  inputNode,
+  outputNode,
```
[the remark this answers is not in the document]
G-Ray: done :+1:

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -5,26 +5,32 @@ import { object, func, string, arrayOf } from 'prop-types';
...
+        width: boundingRect.width,
+      };
+    })();
 
   useEffect(() => {
     const keyDownHandler = e => {
       const { key } = e;
 
-      if (document.activeElement.getAttribute('id') !== inputId) {
+      if (document.activeElement.getAttribute('id') !== inputNode.id) {
```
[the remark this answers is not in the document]
G-Ray: Yeah, I think it was needed when the suggest was mounted on search bar with ReactDom

src/panel/direction/DirectionForm.jsx
G-Ray: yep, done

src/components/ui/Suggest.jsx
```diff
@@ -1,27 +1,40 @@
...
-      fetchItems(searchInputDomHandler.value);
+      if (inputNode.value === '') {
+        setIsOpen(true);
+        fetchItems('');
+      } else {
+        inputNode.select();
+        if (isMobile) {
+          setIsOpen(true);
+          fetchItems(inputNode.value);
+        }
```
amatissart, line 36: As a consequence, it's now impossible to see the suggestions list after the search field is blurred and no result has been selected. Is it really expected ?

src/components/ui/Suggest.jsx
```diff
@@ -50,57 +63,74 @@ const Suggest = ({
     }, SUGGEST_DEBOUNCE_WAIT);
 
     const handleInput = e => {
-      fetchItems(e.target.value);
+      const { value } = e.target;
+      if (value !== lastQuery) {
+        fetchItems(value);
+        setIsOpen(true);
+      }
```
amatissart, line 70: The value of `lastQuery` is not updated here during typing: as far as I understand the component is not re-rendered and the updated state value is not accessible.

Actually I am not sure this check is useful. `fetchItems` is already responsible for caching results, so maybe we could remove this check completely.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -332,6 +332,7 @@ export default class DirectionPanel extends React.Component {
           className="direction_panel"
         >
           {form}
+          <div id="itinerary_autocomplete_suggestions" />
```
amatissart, line 335: The suggestions list remain visible after the last field (that triggers the directions search) is submitted with Enter key:

src/components/ui/Suggest.jsx
G-Ray: I think you're right about removing this check. `lastQuery` was updated in `fetchItems`. I moved this update in this `handleInput` function, thanks. `lastQuery` is notably used to restore what the user has previously typed when cycling over suggests results with arrow keys. (after one complete cycle, the typed value is restored).

src/components/ui/Suggest.jsx
G-Ray: you mean on desktop right ? This is the current design choice, we should contribute your comment to the discussion about design, what do you think ?

src/panel/direction/DirectionPanel.jsx
amatissart: This bug turns out to be specific to Firefox.

G-Ray: thanks, it should be fixed now

src/panel/direction/DirectionInput.jsx
```diff
@@ -23,11 +24,21 @@ class DirectionInput extends React.Component {
   }
 
   componentDidMount() {
+    if (this.props.isLoading) {
+      this.props.inputRef.current.blur();
```
G-Ray, line 28: for instance, `isLoading` is true when the app is restoring origin and destination points from the url after a reload[...]

src/panel/direction/DirectionInput.jsx
```diff
@@ -23,11 +24,21 @@ class DirectionInput extends React.Component {
...
+    }
+
     this.setState({
       mounted: true,
     });
   }
 
+  componentDidUpdate(prevProps) {
+    if (!prevProps.isLoading && this.props.isLoading) {
+      this.props.inputRef.current.blur();
```
G-Ray, line 38: [...] Or when selecting a poi. This will implicitly close the dropdown.

PR #648, opened 2020-05-19
Recycling POI block

bbecquet:
## Description
Implement a "Recycling" POI block, to display details about recycling containers.

**Note**
This PR is quite big for such content as I used it as an excuse to introduce things that I think were currently missing and can ease development for all parts of the app in the future:
 - I introduced a generic `<Flex>` component. It's very basic for now, to be completed when needed and/or replaced by a similar component from the Qwant Design System if we find a way to use it one day… The goal is to promote the use of the `flex` layout in a standard way, instead of floats or absolute positionings that are still used in many parts of the app with a lot of unnecessary CSS and many inconsistencies. 
 - I introduced a generic `<Text>` component, to standardize the typographic classes beyond just CSS and do common patterns in a consistent way (like displaying an icon before the text). As for `<Flex>`, it's meant to be re-used anywhere possible, adapted to our needs and replaced by a Design System component when the time is right.
 - I developed the `<Meter>` component as a shared one, so it can be re-used easily in other parts if needed. BTW, it's a choice not to use the native `meter` HTML element, as it's a real hell to style properly and would introduce broken designs on older browsers.
 - I introduced a way to export individual generic components from a centralized file, so they can be imported more easily. Once again, the idea is to encourage their use. So, instead of:
```js
import Flex from 'src/components/ui/Flex';
import Text from 'src/components/ui/Text';
import Meter from 'src/components/ui/Meter';
```
It's now possible to do:
```js
import { Flex, Text, Meter } from 'src/components/ui';
```
  
## Screenshots
localhost_3000_place_osm_node_721429822@

### DIFF
--- a/src/components/ui/Flex.jsx
+++ b/src/components/ui/Flex.jsx
@@ -1,0 +1,9 @@
+import React from 'react';
+import classnames from 'classnames';
+
+const Flex = ({ children, spaceBetween, className }) =>
+  <div className={classnames('flex', { 'flex--spaceBetween': spaceBetween }, className)}>
+    {children}
+  </div>;
+
+export default Flex;
--- a/src/components/ui/Flex.jsx
+++ b/src/components/ui/Flex.jsx
@@ -1,8 +1,12 @@
 import React from 'react';
 import classnames from 'classnames';
 
-const Flex = ({ children, spaceBetween, className }) =>
-  <div className={classnames('flex', { 'flex--spaceBetween': spaceBetween }, className)}>
+const Flex = ({ children, inline, spaceBetween, alignCenter, className }) =>
+  <div className={classnames('flex', {
+    'flex--spaceBetween': spaceBetween,
+    'flex--alignCenter': alignCenter,
+    'flex--inline': inline,
+  }, className)}>
     {children}
   </div>;
 
--- a/src/components/ui/Flex.jsx
+++ b/src/components/ui/Flex.jsx
@@ -1,13 +1,22 @@
 import React from 'react';
 import classnames from 'classnames';
+import PropTypes from 'prop-types';
 
-const Flex = ({ children, inline, spaceBetween, alignCenter, className }) =>
-  <div className={classnames('flex', {
-    'flex--spaceBetween': spaceBetween,
-    'flex--alignCenter': alignCenter,
-    'flex--inline': inline,
-  }, className)}>
+const Flex = ({ children, inline, className,
+  justifyContent,
+  alignItems = 'center',
+}) => {
+  const style = { justifyContent, alignItems };
+  return <div
+    style={style}
+    className={classnames('flex', { 'flex--inline': inline }, className)}>
     {children}
   </div>;
+};
+
+Flex.propTypes = {
+  justifyContent: PropTypes.oneOf(['space-between']),
+  alignItems: PropTypes.oneOf(['center']),
+};
 
 export default Flex;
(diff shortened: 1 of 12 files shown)


src/components/ui/Flex.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import classnames from 'classnames';
+
+const Flex = ({ children, spaceBetween, className }) =>
+  <div className={classnames('flex', { 'flex--spaceBetween': spaceBetween }, className)}>
```
G-Ray, line 5: :+1:

src/components/ui/index.js
```diff
@@ -0,0 +1,26 @@
+import Button from './Button';
```
G-Ray, line 1: :+1:

src/panel/poi/blocks/Recycling.jsx
```diff
@@ -0,0 +1,57 @@
...
+const Container = ({ type, filling_level, updated_at }) =>
+  <div className="recycling-container">
+    <Text level="caption" icon="icon_clock" className="u-mb-8">
+      {_(
+        'Updated {datetime}',
+        'recycling', {
+          datetime: Intl.DateTimeFormat(window.getLang(), {
+            year: 'numeric', month: 'numeric', day: 'numeric',
+            hour: 'numeric', minute: 'numeric', second: 'numeric',
+            hour12: false,
```
G-Ray, line 26: Do we really want to force a 24-hour time for countries where a 12-hour time is used ?

src/components/ui/Flex.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import classnames from 'classnames';
+
+const Flex = ({ children, inline, spaceBetween, alignCenter, className }) =>
+  <div className={classnames('flex', {
+    'flex--spaceBetween': spaceBetween,
+    'flex--alignCenter': alignCenter,
```
G-Ray, line 7: What about using props named `align` and `justify` for instance ? This way user could set `align='center/start/end'` etc

bbecquet: I wanted something more declarative and easier for calling code, with just bool props and no 1-to-1 mapping to real flex rules, otherwise there is not much added value compared to just using classes or inline styles.
This is also the approach taken by the design system team (but I can't link to the code because… you know :roll_eyes:). But the prop names are not the same, I should have taken them.
So I guess we can discuss the best action:
 - keeping this approach, but changing the props to match the DS component
 - use your approach, with classic key=value props

G-Ray: If we keep the current approach, the thing is nothing prevent the user to set contradictory props like settings both `alignCenter` and `alignStart` for instance.
By offering an API which restrict possibilities with prop types, we can ensure a predictable behavior and keep a minimal amount of props.

Reference:
https://rebassjs.org/reflexbox/#flexbox-props

As for current Design System implementation, it seems they chose to go with a higher level api, like props `center` setting both `align-items: center` and justify-content: `center` I guess. I think this could be limiting for some case, but could be shorthands for other props.

src/components/ui/index.js
G-Ray: I mean, the index.js is great, not the button import specifically :smile:

src/panel/poi/blocks/Recycling.jsx
bbecquet: Thank you, I removed it. I also fixed a locale problem and removed the display of seconds (I don't think they are critical here :P)

src/components/ui/Flex.jsx
bbecquet: I used props matching flex rules like you proposed, limiting to the ones we really need for now.

PR #649, opened 2020-05-20
Update PoiPanel actions design

G-Ray:
## Description
- Updated design for PoiPanel actions
- Phone button still has an `href` attribute with `tel:` protocol handler
- Phone number visibility is no longer managed by `PoiPanel.jsx`, it's now `ActionsButtons.jsx` responsibility.
- Telemetry `phone` is now sent on a `onClickPhoneNumber` event, meanwhile it was sent only when requesting to show a number before.
- Added a phone block to keep existing behavior (on desktop and mobile, like others blocks for now)

## Screenshots
mobile
### Without phone number
no_phone

## Phone block
phone-block

## PoiCard
poi-card

### DIFF
--- a/src/panel/poi/ActionButtons.jsx
+++ b/src/panel/poi/ActionButtons.jsx
@@ -7,7 +7,6 @@ import Button from 'src/components/ui/Button';
 export default class ActionButtons extends React.Component {
   static propTypes = {
     poi: PropTypes.object.isRequired,
-    isMobile: PropTypes.bool.isRequired,
     isDirectionActive: PropTypes.bool,
     openDirection: PropTypes.func,
     openShare: PropTypes.func,
@@ -17,16 +16,14 @@ export default class ActionButtons extends React.Component {
   }
 
   renderPhone() {
-    const { onClickPhoneNumber, poi, isMobile } = this.props;
+    const { onClickPhoneNumber, poi } = this.props;
     return <Button
       className="poi_panel__action__button poi_panel__action__phone"
       onClick={onClickPhoneNumber}
       icon="icon_phone"
       href={poi.blocksByType.phone.url}
       rel="noopener noreferrer external"
-    >
-      {!isMobile && poi.blocksByType.phone.local_format}
-    </Button>;
+    />;
   }
 
   render() {
--- a/src/panel/poi/ActionButtons.jsx
+++ b/src/panel/poi/ActionButtons.jsx
@@ -1,7 +1,8 @@
 /* globals _ */
 import React from 'react';
 import PropTypes from 'prop-types';
-import classnames from 'classnames';
+
+import Button from 'src/components/ui/Button';
 
 export default class ActionButtons extends React.Component {
   static propTypes = {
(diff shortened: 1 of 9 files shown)


src/scss/includes/component.scss
```diff
@@ -10,6 +10,9 @@ $secondary_text : #5c6f84;
 $secondary_text_clear: #697496;
 $secondary_variant_text : #018475; /* link */
 
+/* actions */
+$action-blue-base: #1a6aff;
```
G-Ray, line 14: name is directly taken from mockups

src/panel/poi/PoiBlockContainer.jsx
```diff
@@ -33,6 +35,7 @@ export default class PoiBlockContainer extends React.Component {
       {hourBlock && <HourBlock block={hourBlock} covid19enabled={!!displayCovidInfo} />}
       {informationBlock && <InformationBlock block={informationBlock} />}
       {websiteBlock && <WebsiteBlock block={websiteBlock} poi={this.props.poi} />}
+      {phoneBlock && <PhoneBlock block={phoneBlock} />}
```
amatissart, line 38: The phone block should be put before the website block, according to the mockup.

src/panel/poi/blocks/Phone.jsx
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  render() {
+    const { block } = this.props;
+
+    return <div className="poi_panel__info__section poi_panel__info__section--phone">
+      <div className="poi_panel__info__section__description">
+        <div className="icon-icon_phone poi_panel__block__symbol"></div>
+        <div className="poi_panel__block__content">
+          {block.local_format}
```
amatissart, line 16: It would be helpful to use a `tel:` link here too, don't you think ?

src/panel/poi/ActionButtons.jsx
```diff
@@ -1,62 +1,58 @@
...
+        className="poi_panel__action__button poi_panel__action__favorite"
+        onClick={this.props.toggleStorePoi}
+        icon={this.props.isPoiInFavorite ? 'icon_star-filled' : 'star'}
+      />
+
+      <Button
+        className="poi_panel__action__button poi_panel__action__share"
+        onClick={this.props.openShare}
+        icon="share-2"
+      />
```
amatissart, line 55: minor issue: a rectangular border is visible on click on these buttons, on Firefox:
image

bbecquet: It's possible to remove it with a special rule for Firefox:
```css
button::-moz-focus-inner {
  border: 0;
}
```

But I think there are other a11y/interactivity problems with the buttons here, like no hover / or active state.

This can be addressed separately I guess, probably by copying the button styles from the design system.

src/panel/poi/blocks/Phone.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class PhoneBlock extends React.Component {
```
bbecquet, line 4: Don't you want to use functional components for new stuff, especially when it's so elementary?

src/panel/poi/PoiBlockContainer.jsx
G-Ray: ok done

src/panel/poi/blocks/Phone.jsx
G-Ray: Of course, it's done

src/panel/poi/blocks/Phone.jsx
G-Ray: :+1: done

src/panel/poi/ActionButtons.jsx
G-Ray: Ok, I added your proposed fix @bbecquet

bbecquet: So I guess we'll address the question of hover/active states separately :)

PR #651, opened 2020-05-20
Direction: check if origin exists before focus

G-Ray:
Fix Focus being stolen when trying to editing the origin while destination was empty.

From now, focus will be given to a field where a POI has not been selected. However, this currently only works when selecting a POI with Enter key.

### DIFF
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -30,10 +30,12 @@ export default class DirectionForm extends React.Component {
     if (isMobileDevice() || this.props.isInitializing) {
       return;
     }
-    const { origin, destination, originInputText, destinationInputText } = this.props;
-    if (!originInputText && (prevProps.destination !== destination || prevProps.isInitializing)) {
+
+    const { origin, destination } = this.props;
+
+    if (!origin && destination && (prevProps.destination !== destination || prevProps.isInitializing)) {
       this.originRef.current.focus();
-    } else if (!destinationInputText && prevProps.origin !== origin) {
+    } else if (origin && !destination && prevProps.origin !== origin) {
       this.destinationRef.current.focus();
     }
   }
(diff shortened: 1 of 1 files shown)


src/panel/direction/DirectionForm.jsx
```diff
@@ -30,10 +30,12 @@ export default class DirectionForm extends React.Component {
...
     }
+
     const { origin, destination, originInputText, destinationInputText } = this.props;
+
     if (!originInputText && (prevProps.destination !== destination || prevProps.isInitializing)) {
       this.originRef.current.focus();
-    } else if (!destinationInputText && prevProps.origin !== origin) {
+    } else if (!destinationInputText && prevProps.origin !== origin && origin) {
       this.destinationRef.current.focus();
     }
```
amatissart, line 40: I have a question, probably unrelated.
  
In these 2 conditions, it's unclear for me why we would need to check for `!originInputText` and `!destinationInputText`, instead of `!origin` and `!destination`. Especially when the input text may be populated asynchronously for latlon objects that require an API call to fetch the address.  

Am I missing something ? And maybe some comments would help to make this clearer.

G-Ray: I agree, I remove destinationInputText and originInputText comparisons :+1:

src/panel/direction/DirectionForm.jsx
```diff
@@ -30,10 +30,15 @@ export default class DirectionForm extends React.Component {
...
-    const { origin, destination, originInputText, destinationInputText } = this.props;
-    if (!originInputText && (prevProps.destination !== destination || prevProps.isInitializing)) {
+
+    const { origin, destination } = this.props;
+
+    if (!origin && destination &&
+        (prevProps.destination !== destination || prevProps.isInitializing)) {
+      // a destination is set, origin is empty, so let's focus on origin
       this.originRef.current.focus();
```
amatissart, line 39: Thanks for the comments !  
But we still want to focus the origin field when both fields are empty, right ?

G-Ray: This is still the case using `prevProps.isInitializing`, if page is "reloaded".

PR #654, opened 2020-05-25
Popup: close on move

G-Ray:
## Description
close popup on map `move`. An option named `closeOnMove` can be passed to `Popup()`, but we would miss the chance to clear our `popupHandle` in this case.

## Why
When hovering a poi, a popup appears.
However, if we dezoom the map, the poi will eventually disappear meanwhile the popup will stay.

### DIFF
--- a/src/adapters/poi_popup.js
+++ b/src/adapters/poi_popup.js
@@ -47,6 +47,10 @@ PoiPopup.prototype.addListener = function(layer) {
     }, WAIT_BEFORE_DISPLAY);
   });
 
+  this.map.on('move', () => {
+    this.close();
+  });
+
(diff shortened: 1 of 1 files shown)


src/adapters/poi_popup.js
```diff
@@ -72,7 +80,6 @@ PoiPopup.prototype.showPopup = function(poi, event) {
   const popupOptions = {
     className: 'poi_popup__container',
     closeButton: false,
-    closeOnClick: true,
```
G-Ray, line 75: If we used `closeOnClick` or `closeOnMove`, we wouldn't have a chance to call `this.close()`, which would results in `this.popupHandle` not being cleared from memory.

PR #655, opened 2020-05-25
[Fix] Recycling POIs in directions

bbecquet:
## Description
Fixes two bugs related to recycling POIs used as origin or destination in the Direction panel:
 1. when clicking on a recycling POI, then choosing the Direction button, an error occured when generating the URL part describing the POI
  *Fix: don't serialize the `name` part in the URL if non-existent* 
 2. when already in the Direction panel and clicking on a recycling POI to fill one of the direction points, the field renamed empty
  *Fix: display the address/position of the POI (like for `latlon` POI) if name is non-existent*

## Why
The recycling POI are special because for now they are the only kind of interactive POIs which don't have a name. This caused problems for some operations on direction origin/destination points, which assumed the POIs were either raw locations (`latlon`, already managed as a special case) or interactive POIs with a mandatory name.

### DIFF
--- a/src/libs/pois.js
+++ b/src/libs/pois.js
@@ -9,7 +9,7 @@ export function toUrl(poi) {
   if (poi.id === 'geolocalisation' || poi.type === 'latlon') {
     return `latlon:${poi.latLon.lat.toFixed(5)}:${poi.latLon.lng.toFixed(5)}`;
   }
-  return `${poi.id}@${ExtendedString.slug(poi.name)}`;
+  return poi.name ? `${poi.id}@${ExtendedString.slug(poi.name)}` : poi.id;
 }
 
 export function toAbsoluteUrl(poi) {
--- a/src/libs/pois.js
+++ b/src/libs/pois.js
@@ -12,6 +12,18 @@ export function toUrl(poi) {
   return poi.name ? `${poi.id}@${ExtendedString.slug(poi.name)}` : poi.id;
 }
 
+export async function getInputValue(poi) {
+  if (poi.type === 'latlon' || !poi.name) {
+    return await reverseGeocode(poi);
+  }
+  return poi.getInputValue();
+}
+
+async function reverseGeocode(poi) {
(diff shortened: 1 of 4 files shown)


src/libs/pois.js
```diff
@@ -9,7 +9,19 @@ export function toUrl(poi) {
...
+}
+
+export async function getInputValue(poi) {
+  if (poi.type === 'latlon' || !poi.name) {
+    return await reverseGeocode(poi);
+  }
+  return poi.getInputValue();
+}
+
+async function reverseGeocode(poi) {
```
amatissart, line 22: `reverseGeocode` may be a confusing name for this function. In many cases, an address exists in the POI from the database, and no actual "reverse" is required.
Maybe `fetchAddressLabel` would be a better name ?

bbecquet: You're right, I'll use your proposition.

src/libs/pois.js
```diff
@@ -9,7 +9,19 @@ export function toUrl(poi) {
   if (poi.id === 'geolocalisation' || poi.type === 'latlon') {
     return `latlon:${poi.latLon.lat.toFixed(5)}:${poi.latLon.lng.toFixed(5)}`;
   }
-  return `${poi.id}@${ExtendedString.slug(poi.name)}`;
+  return poi.name ? `${poi.id}@${ExtendedString.slug(poi.name)}` : poi.id;
+}
+
+export async function getInputValue(poi) {
```
G-Ray, line 15: Can we define these functions as methods on parent class `Poi` ?

G-Ray: this way, any inherited classes like `Bragi_poi` would benefit from the default implementation, and could override it

bbecquet: Nope, the overall plan is to get rid of the classes completely, not complexify them :) Most of the functions in this file are already factorizations of behaviors that were scattered in all different POI classes, which is complex and increases special cases and such.
At term I'd like to manipulate simple flat structures with pure functions only.

PR #657, opened 2020-05-26
Improve standard button style and use it for CRO contributing

bbecquet:
## Description
Put the colors and `primary` CTA variant directly in the `<Button>` component style, as well as some tweaks according to the design.
As a result, remove some style over-specifity where `<Button>` was used.
Then change the style of the button to contribute to CaResteOuvert in the POI panel. 

## Why
Consistency \o/

## Screenshots
|Before|After|
|---|---|
|Capture d’écran 2020-05-26 à 14 56 30|Capture d’écran 2020-05-26 à 14 56 02|

### DIFF
--- a/src/components/ui/Button.jsx
+++ b/src/components/ui/Button.jsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import classnames from 'classnames';
+import PropTypes from 'prop-types';
 
 const Button = ({
-  children, icon, variant, className = '', type = 'button',
-  href, onClick, ...rest
+  children, icon, variant = 'outline', type = 'button',
+  href, onClick, className, ...rest
 }) => {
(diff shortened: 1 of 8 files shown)


G-Ray:
Great :clap: 

What about:
- Set default variant props to `primary`
- Set an `outline` variant, used for covid button for instance.

bbecquet:
> Great clap
> 
> What about:
> 
>     * Set default variant props as `primary`
> 
>     * Set an `outline` variant, used for covid button for instance.

According to the designs, the "outline" style is the default, i.e. the style you see repeated multiple times in a design, in contrary to the primary CTA which can be present only one.
For this reason, I don't think `primary` should be a default, and the most common case should be the easier to use, requiring no extra prop.

PR #659, opened 2020-05-28
Redesign poi category items

G-Ray:
## Description
- Implement new design for poi results from categories.
- Removed Phone number button on desktop (Please note the telemetry call has been removed)
- Opening Hour is now displayed below reviews
- :warning: Color variables changes. Please note this can have repercussions accross the app.
- Red border when Hovering is now displayed with a before pseudo-element instead, this way it does not take extra space event when not hovering.

## Screenshots
mobile

desktop


src/panel/category/PoiItem.jsx
```diff
@@ -0,0 +1,52 @@
...
+import PoiTitleImage from 'src/panel/poi/PoiTitleImage';
+
+const PoiItem = ({ poi }) => {
+  const reviews = poi.blocksByType.grades;
+  const address = poi.address || {};
+
+  const Subclass = () =>
+    poi.subClassName
+      ? <p className="u-text--subtitle u-firstCap">{poiSubClass(poi.subClassName)}</p>
+      : <br/>
```
bbecquet, line 15: I don't understand why you need `<br />`. Let's aim for the minimal DOM and return nothing if we can. Or do we really want weird white spaces?

src/components/OpeningHour.jsx
```diff
@@ -8,33 +8,40 @@ const getStatusMessage = status => {
...
       {' '}
       <div className="openingHour-circle" style={{ background: color }} />
     </div>;
   }
 
-  return <div className="openingHour">
+  return <div className="openingHour u-text--label">
     {label}
-    {nextTransition &&
+    {!showNextOpenOnly && nextTransition &&
```
bbecquet, line 40: Looks like it can be simplified, with a single condition for this block and the one below.

src/scss/includes/component.scss
```diff
@@ -3,10 +3,10 @@
 $panel_width : 400px;
 
 /* text */
-$primary_text: #353c52;
+$primary_text: #0c0c0e;
```
bbecquet, line 6: This is a change for the whole app, and quite big for this primary color going from gray to black. I don't think this is a problem, but it probably shouldn't have been in a PR related to a single feature.

src/scss/includes/components/poiItem.scss
```diff
@@ -0,0 +1,34 @@
+.PoiItem {
```
bbecquet, line 1: I had all the css classes start with a lower case letter, like their file.

src/scss/includes/components/poiItem.scss
```diff
@@ -0,0 +1,34 @@
...
+  padding: 16px 30px;
+  cursor: pointer;
+  display: flex;
+
+  @media (max-width: 640px) {
+    padding: 12px 16px;
+    display: flex;
```
bbecquet, line 8: Useless if it's already in the general case.

src/scss/includes/components/poiItem.scss
```diff
@@ -0,0 +1,34 @@
...
+    overflow: hidden;
+    margin-bottom: 4px;
+  }
+
+  &-reviews {
+    display: inline-flex;
+    margin-bottom: 4px;
+  }
+
+  &-right {
```
bbecquet, line 29: Not sure I understand the rules on this element.
Do we want to center vertically the image? What does the `margin-left: auto` do?
BTW, on the screenshots, the icons of POI without an image isn't aligned with the image of the other POIs.

src/scss/includes/reviewScore.scss
```diff
@@ -1,10 +1,11 @@
 .reviewScore {
```
bbecquet, line 1: The `<ReviewScore>` component is also used in popups for PJ POIs. Have you checked this doesn't change the look too much?

src/scss/includes/reviewScore.scss
```diff
@@ -1,10 +1,11 @@
...
-  display: inline-block;
-  color: #5c6f84;
-  margin: 2px 10px 2px 0;
+  color: $secondary_text;
+  display: inline-flex;
+  justify-content: center;
```
bbecquet, line 4: This rule probably doesn't have any effect as it's an inline element without a specified width.

src/panel/category/PoiItem.jsx
G-Ray: IIRC, I found it a bit weird to get the address at the same position, with the same font as the `subClassName` (type of poi). So I prefered to add a `<br />` tag to keep the address at the usual position. I removed these `br` tags, let's see if it's ok for all.

src/scss/includes/component.scss
G-Ray: Yes, I missed to note this change in the PR's description. Do you think this could cause inconsistencies between expected design and the app ?

Should we Revert this change and open another PR focused on color changes ?

src/scss/includes/components/poiItem.scss
G-Ray: The name `right` might be confusing ? It's the right section of the item (the card). Hence, `margin-left: auto` will "push" the element to the right.

src/scss/includes/reviewScore.scss
G-Ray: Yes, only `align-items` is necessary.

src/scss/includes/reviewScore.scss
G-Ray: Yes it seems ok to me

G-Ray: image

src/scss/includes/component.scss
bbecquet: I don't think it's a big deal, as long as it evolved toward the target design. But let's split such changes next time :)

src/scss/includes/reviewScore.scss
bbecquet: I meant the map popups on, not the cards.
Capture d’écran de 2020-06-01 18-14-15

But it looks ok anyway :+1:

src/components/OpeningHour.jsx
```diff
@@ -8,33 +8,37 @@ const getStatusMessage = status => {
...
     </div>;
   }
 
-  return <div className="openingHour">
+  return <div className="openingHour u-text--label">
     {label}
-    {nextTransition &&
+    {((showNextOpenOnly && status === 'closed') ||
+      (!showNextOpenOnly && nextTransition)) &&
       ` - ${_('until {nextTransitionTime}', 'hour panel', { nextTransitionTime: nextTransition })}`
```
bbecquet, line 42: Apparently `nextTransitionTime` isn't always defined.
Capture d’écran de 2020-06-01 18-17-58

src/scss/includes/components/poiItem.scss
bbecquet: Thanks for the explanation. That's the kind of case where I prefer using `space-between` for pushing to the right, and the `margin-left` to control the mininum margin with the element on the left.

src/panel/category/PoiItem.jsx
bbecquet: Looks ok to me.

src/scss/includes/components/poiItem.scss
```diff
@@ -0,0 +1,33 @@
+.poiItem {
+  padding: 16px 30px;
```
bbecquet, line 2: According to the designs, `30px` is too much. You can see the content should start and stop at the same level as the separator bars. `18px` seems to be ok.

src/scss/includes/components/poiItem.scss
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  .poiTitleImage {
+    width: 65px;
+    height: 65px;
+  }
+
+  &-address {
+    text-overflow: ellipsis;
+    max-width: 210px;
```
bbecquet, line 17: Trick: You can avoid this fixed value and have something really responsive by giving the rule `min-width: 0;` to the container `div` (unnamed for now, but could be `.poiItem-left`)

src/components/OpeningHour.jsx
G-Ray: ok, I added a fix, thanks

src/scss/includes/components/poiItem.scss
G-Ray: Thanks for the tip, I set space-between and added a margin-left of 16px on the "right" class. Is the name ".right" ok for you ? May I rename it rightSection or something else ?

src/scss/includes/components/poiItem.scss
G-Ray: I removes the extra space taken by the "red border" when hovering on the left. 4px was takend even when not hovering. It seems ok to me, what do you think ?

src/scss/includes/components/poiItem.scss
G-Ray: Ok, I added a `left` class. Not sure about the benefits of setting `min-width: 0` through

bbecquet: It's required for the ellipsis to work without the hard-coded size. As I understand it, the idea is this rule forces the browser to compute an explicit size for the container during its layout phase.
I'll test it on multiple browsers to see if it does the trick everywhere.

src/scss/includes/components/poiItem.scss
bbecquet: Yep, thank you. Now content and separator are correctly aligned again. (I still think personnaly we lose too much space with this large padding, but this is consistent with the design so :man_shrugging:)

src/scss/includes/components/poiItem.scss
bbecquet: Thanks for the change. And `.poiItem-right` is ok.

bbecquet: The alignement problem of icons vs. images 
Capture d’écran de 2020-06-04 12-03-44

src/scss/includes/components/poiItem.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  .poiTitleImage {
+    width: 65px;
+    height: 65px;
+  }
+
+  &-left {
+    min-width: 0;
+
+    &-address {
```
bbecquet, line 19: You have to put `.&-address` and `.&-reviews` outside of `.&-left`, otherwise the `&` will refer to `.poiItem-left`, not `.poiItem`.

G-Ray: my bad yes, I fixed it

PR #660, opened 2020-06-01
Add a few tooltips

G-Ray:
### Description
- Add a few tooltips on actions across the app
- Add `mapbox/locale.js` to patch mapbox strings with our i18n process

I got an overwhelming diff when generating i18n files, so I did not add new string to po files. Should I do something ?

I know scoping has no effect on our i18 strings for now, but I let some "generic" reusable strings like close or menu without scoping.

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -18,6 +18,7 @@ import { toUrl, getBestZoom } from 'src/libs/pois';
 import Error from 'src/adapters/error';
 import { fire, listen } from 'src/libs/customEvents';
 import { isNullOrEmpty } from 'src/libs/object';
+import locale from '../mapbox/locale';
 
 const baseUrl = nconf.get().system.baseUrl;
 
@@ -68,6 +69,7 @@ Scene.prototype.initMapBox = function() {
     center: this.center,
     hash: false,
     maxZoom: 20,
+    locale,
   });
 
   this.popup.init(this.mb);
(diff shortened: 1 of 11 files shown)


src/mapbox/extended_nav_control.js
```diff
@@ -10,11 +10,19 @@ export default class ExtendedControl {
     this.bottomButtonGroup = document.createElement('div');
 
     const buttonClass = 'icon-plus map_control_group__button__zoom';
-    this._zoomInButton = this._createButton(buttonClass, 'Zoom In', () => this._map.zoomIn());
+
+    this._zoomInButton = this._createButton(
+      buttonClass,
+      'Zoom In',
+      () => this._map.zoomIn(),
+      'Zoom +'
```
bbecquet, line 18: There is no added value here to use a different text in `title` and `aria-label`, you can just re-use the same string for both.

src/adapters/scene.js
```diff
@@ -68,6 +69,7 @@ Scene.prototype.initMapBox = function() {
     center: this.center,
     hash: false,
     maxZoom: 20,
+    locale,
```
bbecquet, line 72: :+1:

src/panel/direction/VehicleSelector.jsx
```diff
@@ -15,6 +23,7 @@ const VehicleSelector = ({ vehicles, activeVehicle, onSelectVehicle }) =>
       )}
       onClick={() => onSelectVehicle(vehicle)}
       aria-label={vehicle}
```
bbecquet, line 25: You can apply the same localization to aria-label

src/mapbox/extended_nav_control.js
G-Ray: Ok, title is set according to ariaLabel now

PR #661, opened 2020-06-01
[Fix] Recycling POIs in directions (not broken this time)

bbecquet:
## Description
Restore and fix this PR https://github.com/QwantResearch/erdapfel/pull/655, which had been reverted in #658 because it introduced bugs.

Also add an integration test case for the behavior of direction inputs, to limit such problems in the future.

### DIFF
--- a/src/libs/pois.js
+++ b/src/libs/pois.js
@@ -9,7 +9,19 @@ export function toUrl(poi) {
   if (poi.id === 'geolocalisation' || poi.type === 'latlon') {
     return `latlon:${poi.latLon.lat.toFixed(5)}:${poi.latLon.lng.toFixed(5)}`;
   }
-  return `${poi.id}@${ExtendedString.slug(poi.name)}`;
+  return poi.name ? `${poi.id}@${ExtendedString.slug(poi.name)}` : poi.id;
+}
+
+export async function getInputValue(poi) {
+  if (poi.type === 'latlon' || !poi.name) {
+    return await fetchAddressLabel(poi);
+  }
+  return poi.getInputValue();
+}
+
+async function fetchAddressLabel(poi) {
+  const address = await IdunnPoi.poiApiLoad(poi);
+  return address.alternativeName || address.name;
 }
 
 export function toAbsoluteUrl(poi) {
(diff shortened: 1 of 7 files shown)


src/panel/direction/DirectionInput.jsx
```diff
@@ -70,12 +71,12 @@ class DirectionInput extends React.Component {
       }
 
       if (selectedPoi.status === navigatorGeolocationStatus.FOUND) {
-        this.props.onChangePoint(selectedPoi.getInputValue(), selectedPoi);
+        this.props.onChangePoint(await getInputValue(selectedPoi), selectedPoi);
```
bbecquet, line 74: The bug on direction inputs was here, and the same below: I forgot the `await`, so the field displayed `[object Promise]` instead of the POI label.

src/libs/pois.js
```diff
@@ -9,7 +9,19 @@ export function toUrl(poi) {
...
     return `latlon:${poi.latLon.lat.toFixed(5)}:${poi.latLon.lng.toFixed(5)}`;
   }
-  return `${poi.id}@${ExtendedString.slug(poi.name)}`;
+  return poi.name ? `${poi.id}@${ExtendedString.slug(poi.name)}` : poi.id;
+}
+
+export async function getInputValue(poi) {
+  if (poi.type === 'latlon' || !poi.name) {
+    return await fetchAddressLabel(poi);
+  }
```
amatissart, line 18: Minor: this may trigger a redundant request to the server.
For example when `poi` has no name but comes from the PoiPanel, data have already been fetched from the server, and the address label is already available in `poi.alternativeName` (still a bad naming...).

bbecquet: Good catch, thank you :+1:  Still the mess about POI mutation between all classes, and the bad naming made me miss that…
As we all agree `alternativeName` is bad, I propose to change it for something better, and try to clean all from there. Would it be ok for you if I do that in another PR?

tests/integration/tools.js
```diff
@@ -38,3 +39,16 @@ export const clearStore = async function(page) {
     localStorage.clear()
   );
 };
+
+export async function simulateClickOnMap(page, latLng) {
```
G-Ray, line 43: :+1: If anyone find a way to avoid duplicating the code here

src/libs/pois.js
amatissart: ok :+1:

PR #663, opened 2020-06-02
Qmaps 1553 - create a share menu and use it in favorites/poi panel/routes

xem:
## Description
Create ShareMenu component (dropdown)
Modify favorites list to include share + delete icons
Cleanup old share components

## Screenshots

drop


G-Ray:
Share button is no longer circular, it's now `38 x 36` instead of `36 x 36`
image

src/components/ui/ShareMenu.jsx
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const twitterShareUrl = location => {
+  return `https://twitter.com/intent/tweet?url=${ encodeURIComponent(location) }`;
+};
+
+const menu_height = 3 * 32;
+
+export default class ShareMenu extends React.Component {
+  _isMounted = false;
```
G-Ray, line 17: You can get rid of this class property, you can already use this.state.open to know if the menu has to be close

G-Ray: this way we'll not even need to use component lifecycle methods

xem: I didn't write this behavior, took it from the previous ContextMenu component.
Are you sure we can do without it? For example, when the poi panel is unloaded, shouldn't we tell the currently open ShareMenu to close itself?

G-Ray: I'm not sure to understand your point. The behavior should be the same if you replace this check : https://github.com/QwantResearch/erdapfel/pull/663/files#diff-9f0fe226d7a5706cacfdfed978e5c974R56 with 
```js
if (this.state.open) {
  this.setState({ open: false });
}
```

Can you confirm ?

src/scss/includes/components/shareMenu.scss
```diff
@@ -1,37 +1,33 @@
...
+    width: 150px;
   }
 
   &-menuItem {
-    padding: 4px 6px;
+    padding: 7px 12px;
+    height: 32px;
     cursor: pointer;
     
     &:hover {
```
G-Ray, line 33: A new color `#e5eeff` for `hover` state should be introduced according to the mockups

G-Ray: Also, text should adopt `grey-black` color on hover.
I think we should define color variables with names taken from mockups, like `grey-black`.

src/scss/includes/panels/favorite_panel.scss
```diff
@@ -53,13 +53,21 @@ $MASQ_BANNER_HEIGHT: 93px;
...
   white-space: nowrap;
   color: $primary_text;
   font-weight: bold;
 }
 
+.favorite_panel__item__share,
+.favorite_panel__item__delete {
+  width: 30px;
+  font-size: 16px;
+  color: #59595F;
```
G-Ray, line 68: Can we define and reuse `grey-semi-darkness` in component.scss ?

src/components/ui/ShareMenu.jsx
```diff
@@ -0,0 +1,115 @@
...
+    open: false,
+    top: 0,
+    left: 0,
+  }
+
+  componentDidMount() {
+    this._isMounted = true;
+  }
+
+  componentWillUnmount() {
```
G-Ray, line 34: We need to call `this.close()` here to prevent the following error when navigating in the panel while a menu is still opened: `react_devtools_backend.js:6 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

src/panel/poi/ActionButtons.jsx
```diff
@@ -51,12 +52,11 @@ export default class ActionButtons extends React.Component {
...
 
-      <Button
-        className="poi_panel__action__button poi_panel__action__share"
-        onClick={this.props.openShare}
-        icon="share-2"
-        title={_('Share', 'poi panel')}
-      />
+      <Button className="poi_panel__action__button poi_panel__action__share"
+        title={_('Share', 'poi panel')}>
+        <ShareMenu url={window.location.toString()} scrollableParent=".panel-content"/>
```
G-Ray, line 57: This leads us to a invalid dom tree where a button is inside another button. Style of the button is corrupted and a tiny area is not clickable.

I think we should let the button triggering the action of displaying the menu as before, and add the `SareMenu` component only once a click has been triggered. `<SareMenu` can take an element/ref as props in order to compute its display position.

This way, `ShareMenu` is only responsible of displaying the menu, meanwhile any buttons could trigger it. That would lead us to a better "separation of concerns".

What do you think @xem @bbecquet ?

xem: I want to do that at the beginning but never found a clean solution to measure where a given DOM node (for example the share button defined in another component) is positioned on the screen. That's why I ended up including the button in ShareMenu :/ 
React really isn't dev-friendly for this kind of stuff.

src/components/ui/ShareMenu.jsx
xem: yep, works great, thanks

src/panel/poi/ActionButtons.jsx
G-Ray: Once you have the reference to the element your should be able to compute its position with `node.getBoundingClientRect()`, is there any constraints ?

src/scss/includes/panels/favorite_panel.scss
xem: sure

src/panel/poi/ActionButtons.jsx
xem: We had this discussion yesterday with @bbecquet about the scrollableParent prop and finally decided that the least dirty solution was to pass it as a classname string to ShareMenu and let ShareMenu do things like:
`document.querySelector(this.props.scrollableParent).addEventListener('scroll', this.close);`

If you think the ref approach is cleaner, we can use it for the buttons and the scrollable parent. Let's discuss that with Benjamin tomorrow or monday :)

bbecquet: @G-Ray The reason why I suggested @xem to avoid ref is because you would have to pass it down the tree at every level, from the Panel to the DropDown itself, which introduces a lot of potential dependency for a component that should be generic, relatively self-contained and can be placed anywhere.
The use of a selector seemed like the simple and pragmatic choice, instead of aiming for a 100% pure component without any concrete benefit.

G-Ray: Ok @bbecquet for the scrollableParent, however I'm still thinking about a a code like this to pass a node/ref of the button which is triggerring the `ShareMenu` :

```js
<Button
  ref={this.myRef}
  className="poi_panel__action__button poi_panel__action__share"
  title={_('Share', 'poi panel')}>
  onClick={() => this.setState({ openShare: !this.state.openShare })
/>

{this.state.openShare && (
  <ShareMenu
    url={window.location.toString()}
    scrollableParent=".panel-content"
    outputNode={this.myRef}
  />
)}
```

This way we still have the existing `scrollableParent` class to register scroll events, meanwhile we have a new prop `outputNode` which whille let the dropdown know where it should be rendered in absolute positioning.

bbecquet: @G-Ray I like this solution because it doesn't hide the triggering element in the ShareMenu itself, so it's easier to define the style you want.
But I'd use another name than `outputNode`, which suggests that we will inject something in this node, and that's not the case. What about `triggerRef`?
Following this path, I would prefer to manage the `onClick` inside `<ShareMenu>` itself. That way, we don't need special state in the container. The ShareMenu would just render nothing until it has been triggered.
```jsx
<Button
  ref={this.shareButtonRef}
  className="poi_panel__action__button poi_panel__action__share"
  title={_('Share', 'poi panel')}>
/>
<ShareMenu
  url={window.location.toString()}
  scrollableParent=".panel-content"
  triggerRef={this.shareButtonRef}
/>
```

bbecquet: Oops, I forgot function components can't be given refs, so this can't work.
So there is your proposition with the state. Or I can try something with a render function :)

G-Ray: Still possible with React.forwardRef ;)
https://reactjs.org/docs/forwarding-refs.html

bbecquet: This would be the resulting JSX. No required ref nor state, and complete control over the trigger element. Seems to me like the best solution, while keeping in line with React principle of considering composition before other solutions.
```jsx
<ShareMenu
  url={window.location.toString()}
  scrollableParent=".panel-content"
>
  {openMenu => <Button className="…"
    title={_('Share', 'poi panel')}
    icon="icon-share-2"
    onClick={openMenu}
   />}
</ShareMenu>
```

src/modals/ShareModal.jsx
```diff
@@ -1,97 +0,0 @@
...
-      </div>
-    </Modal>;
-  }
-}
-
-function close() {
-  ReactDOM.unmountComponentAtNode(document.querySelector('.react_modal__container'));
-}
-
-export function openShareModal(url) {
```
bbecquet, line 83: We have to put that elsewhere, that would be a pity to remove this feature you just implemented ;)

src/components/ui/ShareMenu.jsx
```diff
@@ -0,0 +1,124 @@
...
+  render() {
+    const { url, children } = this.props;
+
+    return <Fragment>
+      {children(this.open)}
+      {this.state.open && <div className="shareMenu-menu"
+        style={{ left: this.state.left + 'px', top: this.state.top + 'px' }}>
+
+        <div className="shareMenu-menuItem shareMenu-menuItem--copy" onClick={e => {
+          e.preventDefault();
```
bbecquet, line 93: If these two lines are useless, let's just remove them.

src/components/ui/ShareMenu.jsx
```diff
@@ -0,0 +1,124 @@
...
+          {
+            this.state.copied
+              ? <span className="shareMenu-menuItem--copied">
+                { _('Copied!') }
+              </span>
+              : _('Copy link', 'share')
+          }
+        </div>
+
+        <div className="shareMenu-menuItem shareMenu-menuItem--facebook" onClick={e => {
```
bbecquet, line 108: I guess you'll have a linter error about `e` not being used.

PR #665, opened 2020-06-04
Make panel resizeable for default size

G-Ray:


src/components/ui/Panel.jsx
```diff
@@ -188,7 +184,10 @@ export default class Panel extends React.Component {
       >
         {resizable && size === 'minimized' && minimizedTitle ? minimizedTitle : title}
       </div>
-      <div className="panel-content">
+      <div className="panel-content"
+        style={(size === 'default' ? { overflow: 'hidden' } : {})}
```
bbecquet, line 188: `size` is set as a class on panel, so you should be able to set this rule in the CSS with `.panel.default`. (BTW, I should have written these clases as `panel--{size}`, but that's another problem)

src/components/ui/Panel.jsx
```diff
@@ -104,8 +104,6 @@ export default class Panel extends React.Component {
   }
 
   holdResizer = event => {
-    event.preventDefault();
```
bbecquet, line 107: Can you explain why this is needed?

G-Ray: I restored the old code as we want to prevent mouse events to be raised when touch event has occured.

src/components/ui/Panel.jsx
G-Ray: done, thanks

amatissart:
In that situation:
image

the user needs to scroll to the top in order to reduce the panel size, and I think it's expected.
But the "handle" remains visible and I think it's confusing as it's impossible to interact with it.

G-Ray:
> In that situation:
> image
> 
> the user needs to scroll to the top in order to reduce the panel size, and I think it's expected.
> But the "handle" remains visible and I find this confusing as it's impossible to interact with it.

So I guess we should let the "handle" resize the panel even if content is scrolled right ?

bbecquet:
> So I guess we should let the "handle" resizes the panel even if content is scrolled right ?

Would be the logical thing to do for me. That and also handling the click to go back to the medium size.

G-Ray:
> > So I guess we should let the "handle" resizes the panel even if content is scrolled right ?
> 
> Would be the logical thing to do for me. That and also handling the click to go back to the medium size.

the click should work, can you confirm @bbecquet ?

G-Ray:
The panel's handler should now force a resize @amatissart @bbecquet

bbecquet:
> the click should work, can you confirm @bbecquet ?

Yes, it works, thanks :) And I tested on Firefox and Chrome for Android, everything works according to what we want IMO.

I'm still not convinced that's it an UX improvement rather than a regression, not being able to scroll while in medium size. I tested on the direction result list for example, and it's really annoying not being able to see the third route alternative details at the same time than its representation on the map…

G-Ray:
I tested with Gnome Web (webkit), I needed to remove existing listeners as soon a a new holding occurs to prevent a bug where holding the panel until the top of the screen "where our search bar is" would not occurs a "mouseup" event.

G-Ray:
> GG for being able to make it work through all these special cases.
> 
> However I'm very worried about the complexity introduced for future maintenance. And as I explained I'm not sure it's an improvement…

I think code is still manageable if we don't add more complexity anytime soon :crossed_fingers: . As for the user experience, I personally find it more pleasant to use the panel as this, however we might add an (animated) icon to let the user understand that he needs to open the panel up in order to see all items, would it cover the annoyances you raised @bbecquet ? Also, we should discuss with designers by letting them try this new behavior to get their opinion

PR #667, opened 2020-06-05
i18n: Add occurrences location in .po and sync messages with Transifex

amatissart:
## Description
* Extract translation keys directly from `src` directory instead of using .js bundles
* Add occurrences details in .po (filename and line number)
* Sort .po content by file location 
* Sync messages with Transifex

## Why
This change should make future integration of Transifex messages easier to review, now that .po files will be generated more consistently.

### DIFF
--- a/i18n.sh
+++ b/i18n.sh
@@ -3,8 +3,9 @@
 set -ev
 find config "views/" -name "*.ejs"| xargs xgettext --language=PHP -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -
 find config -name "*.yml"| xargs xgettext --language=Python -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -j -
-find bin "public/build/javascript/" -name "*.js"| xargs xgettext -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -j -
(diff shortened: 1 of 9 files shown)


i18n.sh
```diff
@@ -3,8 +3,9 @@
 set -ev
 find config "views/" -name "*.ejs"| xargs xgettext --language=PHP -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -
 find config -name "*.yml"| xargs xgettext --language=Python -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -j -
-find bin "public/build/javascript/" -name "*.js"| xargs xgettext -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -j -
+find bin src -name "*.js" | xargs xgettext -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -j -
+find src -name "*.jsx" | xargs xgettext --language=Python -k_n:1,2 --force-po -o tt.pot --from-code=UTF-8 -j -
```
amatissart, line 7: Hackish: `xgettext` can't parse .jsx syntax, but it seems that the Python parser is permissive enough to extract translation keys from .jsx files.

amatissart:
Parsing .jsx files with the latest version of gettext, or another parser would be more robust. Let's forget this for now.

New i18n messages have been included in #670

PR #675, opened 2020-06-12
Use Idunn extend_bbox feature on category panel

bbecquet:
## Description
Allow the Idunn API call to take an `extend_bbox` parameter to use the auto BBOX extension feature introduced in https://github.com/QwantResearch/idunn/pull/146.

I refactored the CategoryPanel before that (notably turning it into a functional component with hooks) to have a clearer lifecycle and separate more cleanly the map management, data loading, and rendering.
I thought (now I'm not so sure -_-) it was necessary to avoid falling into dead locks with auto bbox and such, and the resulting hooks are clear as long as you pay attention to when they are actually called.
As a reminder:
```js
  useEffect(() => {
    // ...
  }, [category, query]); // i.e. this hook will be called only when `category` or `query` has changed
```

We use the bbox extension feature, only when loading the first results of a category/query, not on subsequent map pans and zooms. This is to avoid changing the view when the user is interacting with the map. I re-used the `initialLoading` state as an indication of that.

## Why
 - Be more helpful to the user when no data is returned in the first place
 - Try to simplify the CategoryPanel component lifecycle

## Screenshots
Peek 22-06-2020 10-46

### DIFF
--- a/src/adapters/poi/idunn_poi.js
+++ b/src/adapters/poi/idunn_poi.js
@@ -53,9 +53,9 @@ export default class IdunnPoi extends Poi {
     }
   }
   /* ?bbox={bbox}&category=<category-name>&size={size}&verbosity=long/ */
-  static async poiCategoryLoad(bbox, size, category, query) {
+  static async poiCategoryLoad(bbox, size, category, query, extendBbox) {
     const url = `${serviceConfig.idunn.url}/v1/places`;
-    const requestParams = { bbox, size };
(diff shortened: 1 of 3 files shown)


bbecquet:
In hindsight, the refacto to a functional component isn't crucial to the feature itself and makes the review process far more difficult than needed.
I'm closing this one in favor of https://github.com/QwantResearch/erdapfel/pull/686.

PR #676, opened 2020-06-12
pois: define address object and build full address

G-Ray:
Format addresses to`[address.name], [address.city], [address.country]` in suggestions and directions inputs.

:warning: 
Saved favorites do not contains fields needed from geocoder to display the full address. In the following screenshot, `CFA [...]` and `Villla [...]` are favorites already registered in browser before this patch :
favorites

Please also note that directions now display full address, so we do not display "Musée d'Orsay" for instance.

direction

Addresses in lists and on a poi Header are still displayed like before, with zip code.
poi list

### DIFF
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -62,14 +62,4 @@ export default class BragiPoi extends Poi {
       .values(geocoding.administrative_regions)
       .find(a => a.zone_type === name);
   }
-
-  getInputValue() {
-    switch (this.type) {
-    case 'house':
-    case 'street':
-      return this.value;
-    default:
-      return this.name;
-    }
-  }
 }
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -1,5 +1,5 @@
 import Poi from './poi';
-import { normalizeAddress } from '../../libs/address';
+import { normalize as normalizeAddress } from '../../libs/address';
 
 export default class BragiPoi extends Poi {
   constructor(feature, queryContext) {
@@ -51,6 +51,6 @@ export default class BragiPoi extends Poi {
     this.value = label;
     this.queryContext = queryContext;
 
-    this.address = normalizeAddress('bragi', feature.properties.geocoding);
+    this.address = normalizeAddress('bragi', feature.properties);
   }
 }
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -3,7 +3,7 @@ import Poi from './poi';
 export default class BragiPoi extends Poi {
   constructor(feature, queryContext) {
     const geocodingProps = feature.properties.geocoding;
-    const { id, type, label, address, city, administrative_regions } = geocodingProps;
+    const { id, type, label } = geocodingProps;
 
     let poiClassText = '';
     let poiSubclassText = '';
@@ -22,24 +22,14 @@ export default class BragiPoi extends Poi {
 
     /* generate name corresponding to poi type */
     let name = '';
-    let alternativeName = '';
-
-    const postcode = geocodingProps.postcode && geocodingProps.postcode.split(';')[0];
-    const cityObj = administrative_regions.find(region => region.zone_type === 'city');
-    const country = administrative_regions.find(region => region.zone_type === 'country');
-    const countryName = country && country.name;
 
     switch (type) {
     case 'poi':
       name = geocodingProps.name;
-      alternativeName = (address && address.label)
-        || (cityObj && cityObj.label)
-        || [postcode, city, countryName].filter(zone => zone).join(', ');
       break;
     case 'house':
     case 'street':
       name = geocodingProps.name;
-      alternativeName = [postcode, city, countryName].filter(zone => zone).join(', ');
       break;
     default: {
       /* admin */
@@ -48,19 +38,29 @@ export default class BragiPoi extends Poi {
         name = label;
       } else {
         name = label.slice(0, splitPosition);
-        alternativeName = label.slice(splitPosition + 1);
       }
     }
     }
 
-    super(id, name, alternativeName, type, {
+    super(id, name, type, {
       lat: feature.geometry.coordinates[1],
       lng: feature.geometry.coordinates[0],
     }, poiClassText, poiSubclassText, geocodingProps.bbox);
 
     this.value = label;
-
     this.queryContext = queryContext;
+    this.address = {
+      name: geocodingProps.address?.name || geocodingProps?.name,
+      city: this._findAdmin(geocodingProps, 'city')?.name,
+      country: this._findAdmin(geocodingProps, 'country')?.name,
+      label: geocodingProps.label,
+    };
+  }
+
+  _findAdmin(geocoding, name) {
+    return Object
+      .values(geocoding.administrative_regions)
+      .find(a => a.zone_type === name);
   }
 
   getInputValue() {
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -49,8 +49,16 @@ export default class BragiPoi extends Poi {
 
     this.value = label;
     this.queryContext = queryContext;
+
+    let street = geocodingProps?.address?.name;
+    if (geocodingProps.type === 'house') {
+      // Street address is received in the name field
+      street = geocodingProps?.name;
+    }
+
     this.address = {
-      name: geocodingProps.address?.name || geocodingProps?.name,
+      name: geocodingProps?.name,
+      street,
       city: this._findAdmin(geocodingProps, 'city')?.name,
       country: this._findAdmin(geocodingProps, 'country')?.name,
       label: geocodingProps.label,
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -57,7 +57,6 @@ export default class BragiPoi extends Poi {
     }
 
     this.address = {
-      name: geocodingProps?.name,
       street,
       city: this._findAdmin(geocodingProps, 'city')?.name,
       country: this._findAdmin(geocodingProps, 'country')?.name,
--- a/src/adapters/poi/bragi_poi.js
+++ b/src/adapters/poi/bragi_poi.js
@@ -1,4 +1,5 @@
 import Poi from './poi';
+import { normalizeAddress } from '../../libs/address';
 
 export default class BragiPoi extends Poi {
   constructor(feature, queryContext) {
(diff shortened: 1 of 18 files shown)


tests/integration/tests/menu.js
```diff
@@ -40,7 +40,7 @@ test('menu open favorite', async () => {
   page.click('.menu__panel__action:nth-child(3)');
 
   expect(await exists(page, '.favorite_panel')).toBeTruthy();
-});
+}, 10000);
```
G-Ray, line 43: This test often fail on my computer due to a too short duration

bbecquet: Maybe we can augment the timeout for every test in the global config. That would be better than having this 10s here in a test case that doesn't seem so costly. 
I'm curious, how long does it take to run?

src/adapters/poi/bragi_poi.js
```diff
@@ -48,19 +38,29 @@ export default class BragiPoi extends Poi {
...
+    super(id, name, type, {
       lat: feature.geometry.coordinates[1],
       lng: feature.geometry.coordinates[0],
     }, poiClassText, poiSubclassText, geocodingProps.bbox);
 
     this.value = label;
-
     this.queryContext = queryContext;
+    this.address = {
+      name: geocodingProps.address?.name || geocodingProps?.name,
```
bbecquet, line 53: What is an address "name", and what is the difference with a "label"?
This change is the occasion to use more explicitly named properties, maybe we can find something better?

src/adapters/poi/bragi_poi.js
```diff
@@ -48,19 +38,29 @@ export default class BragiPoi extends Poi {
...
+      city: this._findAdmin(geocodingProps, 'city')?.name,
+      country: this._findAdmin(geocodingProps, 'country')?.name,
+      label: geocodingProps.label,
```
bbecquet, line 56: I may be wrong, but it looks like you removed every use of an address label. Is it still needed?

src/adapters/poi/poi.js
```diff
@@ -5,10 +5,9 @@
 export const POI_TYPE = 'poi';
 
 export default class Poi {
-  constructor(id, name, alternativeName, type, latLon, className, subClassName, bbox) {
```
bbecquet, line 8: Good riddance :+1:

src/adapters/poi/bragi_poi.js
G-Ray: address.label is used in `src/components/PoiItem`. This way the behavior is changed only for suggestions.
A label contain the zip code, it can be `7 r Jean Lantier, 75001 Paris` for instance.
I'm not sure we can reconstruct this label from name + zip + city, as order might depend from countries ?

src/adapters/poi/poi.js
G-Ray: yeah, We only get rid of alternativeName for now

src/adapters/poi/bragi_poi.js
G-Ray: Ok let's find a proper name, for now we have:
name: "X Rue Lacépède"
city: Paris
country: France

And the function `getFullAddress` in `pois.js` returns the string "X Rue Lacépède, Paris, France".

We can try to format `X Ru lacépède` from street info, but the order might depends on localization.

G-Ray: differences between a "label" and "getFullAddress" may not be explicit.

tests/integration/tests/menu.js
G-Ray: I only know it's between 5000 and 10000ms :turtle:

src/libs/pois.js
```diff
@@ -13,15 +13,23 @@ export function toUrl(poi) {
 }
 
 export async function getInputValue(poi) {
-  if (poi.type === 'latlon' || !poi.name) {
+  if (poi.type === 'latlon' || !poi.address) {
     return await fetchAddressLabel(poi);
   }
-  return poi.getInputValue();
```
bbecquet, line 19: This is what cause the regression on direction input fields, which is quite big IMO compared to the one on favorites.
I'm totally ok to eventually get rid of the `getInputValue` method on POI classes (I even think it can be removed in after this change), but always returning the address isn't the best solution. The address should be a fallback, but if the POI has a real name it should be displayed in priority.

G-Ray: I don't said it's a regression, I though it was the expected behavior, to get only full addresses in inputs according to the ticket.
Should we display `Musée d'Orsay, Paris, France` in this case ?

G-Ray: This is resolved in the latest commit (https://github.com/QwantResearch/erdapfel/pull/676/commits/c3ea2e8ccf1998bef5b6b9c95c57a1db2359065c) Is it ok ?

G-Ray:
I introduced two functions:
getNameAddress, to get formatted address with poi's name (used for direction inputs),
getStreetAddress, to get formatted address of the poi (used for autocomplete suggestions)

src/libs/pois.js
bbecquet: > I don't said it's a regression

But it is :) "1 rue de la légion d'Honneur, Paris" makes no sense to much people, whereas "Musée d'Orsay" does.
I don't know which ticket you're referring to, there is no mention of directions in 1531, it's only about the 2nd line on suggestions.
I'd say for we just display `Musée d'Orsay` in this case, i.e. the POI name if it's available, like currently.

tests/integration/tests/menu.js
bbecquet: Tests on your computer seem to behave really strangely… Are you mining bitcoins aside? :smile:
Anyway, there is no reason *this* case in particular should take longer. I'm pretty sure if we change the order or add some other test case it will be another one which fails, so there is no point to keep this `10000` here.

src/libs/pois.js
G-Ray: Okay, it makes sense. I updated the PR, it should be more convenient to read and user these `getStreetAddress` and `getNameAddress` now :). Plus, we don't need `getInputValue` anymore!

src/libs/pois.js
```diff
@@ -12,16 +12,32 @@ export function toUrl(poi) {
...
+
+export async function getStreetAddress(poi) {
+  poi = await fetchAddress(poi);
+  return formatAddress(poi.address?.name, poi.address?.city, poi.address?.country);
 }
 
-async function fetchAddressLabel(poi) {
-  const address = await IdunnPoi.poiApiLoad(poi);
-  return address.alternativeName || address.name;
+async function fetchAddress(poi) {
```
bbecquet, line 37: It's a bit strange to return a poi in a function named `fetchAddress`. What do you think about returning the `address` only here?
And maybe if an API call is needed, store the address in the `poi` argument in the function, for caching… But maybe it's a bit too side-effect, I don't know.

src/adapters/poi/bragi_poi.js
bbecquet: Sorry, I'm still not sure what is this address 'name'…
Looking at exemples it looks like it's actually the street part of the address (number + street name).
If that's the case, it probably:
 - should be renamed to something more explicit, like simplify `street` or smth,
 - should be allowed to be empty, for exemple for cities and such,
 - should be distinguished from the name of the POI, when it exists (in idunn_poi the POI name is used as a fallback for it, it may induce confusion.)

src/adapters/poi/bragi_poi.js
bbecquet: Ok for now, let's keep it until we see clearer in all this formatting, there are already a lot of improvements.

src/components/ui/SuggestItem.jsx
```diff
@@ -48,16 +49,24 @@ const CategoryItem = ({ category }) => {
...
 
 const PoiItem = ({ poi }) => {
-  const { name, className, subClassName, type, alternativeName } = poi;
+  const { name, className, subClassName, type } = poi;
   const icon = IconManager.get({ className, subClassName, type });
+  const [streetAddress, setStreetAddress] = React.useState(null);
+
+  useLayoutEffect(() => {
+    (async () => {
+      const streetAddress = await getStreetAddress(poi);
```
amatissart, line 58: This function may trigger a request to the server for items with no `address` object, such as POIs stored in local storage with an older format. In that case, a request is triggered for each item and each keystroke to fetch the address (actually twice per keystroke, as the render seems to be executed twice).
This also causes flickering of the list.

Maybe we would need a mechanism (in `PoiStore` ?) to update the local storage with the new format when a POI is stored with an obsolete format.

src/libs/pois.js
G-Ray: You right, I update and now only return the address.
Do you think we can safely assign the address to the poi for caching purpose ?  I'm not sure either

src/adapters/poi/bragi_poi.js
G-Ray: I remove the name field. We now have :
`street`, `city`, `country`, `label` (used by PoiPanel and PoiPopup)

src/components/ui/SuggestItem.jsx
G-Ray: We now display `alternativeName` when it exists. Is it okay for you ?

amatissart: Yes fine :+1: 
Migrrating the data present in local storage is definitely out of the scope of this PR.

tests/integration/tests/autocomplete.js
```diff
@@ -264,30 +264,20 @@ test('check template', async () => {
...
-  );
+  expect(lines[2][0]).toEqualCaseInsensitive('maga');
+  expect(lines[2][1]).toEqualCaseInsensitive('maga, Handlová, Slovensko');
 
   /* admin */
-  const labelFragments = mockAutocompleteAllTypes.features[3].properties.geocoding.label.split(',');
-  expect(lines[3][0]).toEqualCaseInsensitive(labelFragments[0]);
-  expect(lines[3][1]).toEqualCaseInsensitive(labelFragments.slice(1).join(',').trim());
+  expect(lines[3][0]).toEqualCaseInsensitive('Le Cannet (06110)');
+  expect(lines[3][1]).toEqualCaseInsensitive('Le Cannet, France');
```
amatissart, line 280: Have this change been validated by the Product / Design ?  
In this case, it looks weird to me to repeat the name on the second line, whereas the details about the regions where the city is located are lost. This may be specially confusing when multiples cities have the same name in the same country.

Or maybe we could consider this PR as a good step to make this formatting logic easier to maintain, and polish this behavior separately ?

tests/integration/tests/poi.js
```diff
@@ -259,6 +260,8 @@ test('add a poi as favorite and find it back in the favorite menu', async () =>
   // we then reopen the poi panel and 'unstar' the poi.
   await page.click('.favorite_panel__item');
   expect(await exists(page, '.poiTitle')).toBeTruthy();
+  expect(await exists(page, '.poiTitle')).toBeTruthy();
```
bbecquet, line 263: This line is duplicated

src/libs/address.js
```diff
@@ -0,0 +1,79 @@
+import IdunnPoi from '../adapters/poi/idunn_poi';
+
+/**
+ * Format an address given an address object (name, city, and country)
+ * @param {*} street
```
bbecquet, line 5: The @param comments need to be updated with the new API.

PR #679, opened 2020-06-15
Qmaps 1559 - fix desktop and mobile service panels and itinerary button

xem:
## Description
fix desktop and mobile service panels and itinerary button (colors, fonts, lowercase, margins, ...)

## Screenshots
image
image

### DIFF
--- a/src/panel/service/ServicePanelDesktop.jsx
+++ b/src/panel/service/ServicePanelDesktop.jsx
@@ -1,5 +1,5 @@
 /* global _ */
-import React, { Fragment } from 'react';
+import React from 'react';
 import Panel from 'src/components/ui/Panel';
 import CategoryList from 'src/components/CategoryList';
 import Action from 'src/components/ui/MainActionButton';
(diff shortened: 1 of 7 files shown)


src/panel/service/ServicePanelDesktop.jsx
```diff
@@ -1,24 +1,26 @@
 /* global _ */
-import React from 'react';
+import React, { Fragment } from 'react';
 import Panel from 'src/components/ui/Panel';
 import CategoryList from 'src/components/CategoryList';
 import Action from 'src/components/ui/MainActionButton';
 import nconf from '@qwant/nconf-getter';
 
+const directionConf = nconf.get().direction;
```
G-Ray, line 8: directionConf and Fragment are not used

src/panel/service/ServicePanelDesktop.jsx
```diff
@@ -1,24 +1,26 @@
...
     >
-      <CategoryList className="service_panel__categories" />
+      <div className="service_panel__actions">
 
-      <hr/>
+        <h3 className="u-text--smallTitle u-center">{_('Qwant Maps Services', 'service panel')}</h3>
+
+        <CategoryList className="service_panel__categories" />
+
+        <hr/>
```
G-Ray, line 22: can we use utils class like `.u-mr-8` for instance instead of hr ?

src/scss/includes/panels/service_panel.scss
```diff
@@ -52,6 +51,18 @@
     h3 {
       margin-bottom: 6px;
     }
+
+    .panel-header {
+      padding: 7px 0 0;
```
G-Ray, line 56: Shouldn't we use 4, 8, 16, 24 px only ?

xem: I didn't know we had this kind of rule, but sure, 8px works too

src/panel/service/ServicePanelDesktop.jsx
xem: this class only adds a 8px margin bottom, while hr includes a horizontal line

xem: but I added a missing container around the action buttons so the artificial hr's margin is no longer necessary ;)

src/scss/includes/components/panel.scss
```diff
@@ -36,7 +36,7 @@
   }
 
   &-content {
-    max-height: calc(100vh - 160px);
+    max-height: calc(100vh - 130px);
```
G-Ray, line 39: Would it be possible to compute this size dynamically ?

xem: I think not, except by measuring things in JS.
We'll see that during the incoming homepage redesign

PR #680, opened 2020-06-16
OpeningHour: hide close status

G-Ray:
- Hide Close status, display only `reopening at` (nextTransition)
- Capitalize with css and add `u-ml-4` class
- Improved readability

image

### DIFF
--- a/tests/integration/tests/poi.js
+++ b/tests/integration/tests/poi.js
@@ -77,16 +77,15 @@ test('load a poi from url on mobile', async () => {
   });
   await page.goto(`${APP_URL}/place/osm:way:63178753@Musée_dOrsay#map=17.49/2.3261037/48.8605833`);
   await page.waitForSelector('.poiTitle');
-  const { title, address, hours } = await page.evaluate(() => {
+  const { title, address } = await page.evaluate(() => {
     return {
       title: document.querySelector('.poi_card .poiTitle').innerText,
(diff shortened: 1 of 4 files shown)


tests/integration/tests/poi.js
```diff
@@ -77,16 +77,15 @@ test('load a poi from url on mobile', async () => {
...
     return {
       title: document.querySelector('.poi_card .poiTitle').innerText,
       address: document.querySelector('.poi_card .poiItem-address').innerText,
-      hours: document.querySelector('.poi_card .openingHour').innerText,
     };
   });
   expect(title).toMatch(/Musée d'Orsay/);
   expect(address).toMatch(/1 Rue de la Légion d'Honneur \(Paris\)/);
-  expect(hours).toMatch(/Fermé/);
+  await page.waitForSelector('.openingHour--closed');
```
bbecquet, line 88: This doesn't change much but I would scope the selector to `.poi_card` like the others, for consistency and extra-safety.

And another point: to have real failed assertion messages instead of a dirty Jest errors, I recently changed these kind of tests to: `expect(await exists(page, '.openingHour--closed)).toBeTruthy();`
See https://github.com/QwantResearch/erdapfel/pull/669 for details.
Or if you don't want to bother with that, you can extract the `classname` in the `evaluate` and use `expect.toMatch`.

PR #681, opened 2020-06-17
Search input mobile: focus only if active

G-Ray:
## Description
On mobile, clearing the input with the "close" icon currently triggers a focus on the input.

This PR checks if the input is already focused (suggestions are displayed), and triggers a focus only if it's the case. If the input is not focused, The input is only cleared.
Desktop's behavior is unchanged.

Peek 2020-06-17 12-30

### DIFF
--- a/src/ui_components/search_input.js
+++ b/src/ui_components/search_input.js
@@ -1,5 +1,6 @@
 
 import { selectItem, fetchSuggests } from 'src/libs/suggest';
+import { isMobileDevice } from 'src/libs/device';
 
 const MAPBOX_RESERVED_KEYS = [
   'ArrowLeft', // ←
@@ -11,6 +12,8 @@ const MAPBOX_RESERVED_KEYS = [
   '=', // =
 ];
 
+const SEARCH_INPUT_ID = 'search';
+
 export default class SearchInput {
 
   constructor(tagSelector) {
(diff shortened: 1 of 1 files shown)


src/ui_components/search_input.js
```diff
@@ -21,24 +24,29 @@ export default class SearchInput {
...
+    window.__searchInput = new SearchInput(tagSelector);
 
-      window.clearSearch = () => {
-        window.__searchInput.searchInputHandle.value = '';
-        window.app.navigateTo('/');
+    window.clearSearch = () => {
+      const isMobile = isMobileDevice();
+      window.__searchInput.searchInputHandle.value = '';
+      window.app.navigateTo('/');
+      if (!isMobile || isMobile && document.activeElement.id === SEARCH_INPUT_ID) {
```
bbecquet, line 37: You don't need the `SEARCH_INPUT_ID` constant, the `tagSelector` parameter is already what you need.

src/ui_components/search_input.js
```diff
@@ -74,7 +82,7 @@ export default class SearchInput {
         if (document.activeElement
           && document.activeElement.tagName !== 'INPUT'
           && window.__searchInput.isEnabled) {
-          document.getElementById('search').focus();
```
bbecquet, line 77: Looks like this can be written as `this.searchInputHandle.focus()`.

PR #682, opened 2020-06-17
QMAPS-1454 log nlu suggest clicks and if they correspond to a category

xem:
## Description
log NLU suggest clicks and if they correspond to a category

## Screenshots
hotel nice + clic sur "Hotel à proximité":
image

### DIFF
--- a/src/libs/suggest.js
+++ b/src/libs/suggest.js
@@ -7,6 +7,7 @@ import PoiStore from 'src/adapters/poi/poi_store';
 import Category from 'src/adapters/category';
 import Intention from 'src/adapters/intention';
 import { toUrl } from 'src/libs/pois';
+import Telemetry from 'src/libs/telemetry';
 import { suggestResults } from 'src/adapters/suggest_sources';
 
 const geocoderConfig = nconf.get().services.geocoder;
@@ -24,6 +25,15 @@ export const selectItem = (selectedItem, replaceUrl = false) => {
     window.app.navigateTo(`/places/?type=${selectedItem.name}`,
       {}, { replace: replaceUrl });
   } else if (selectedItem instanceof Intention) {
+    Telemetry.add(
+      Telemetry.SUGGEST_CLICK,
+      null,
+      null,
+      {
+        useNlu: geocoderConfig.useNlu,
+        isCategory: !!selectedItem.category,
+      }
+    );
     window.app.navigateTo(`/places/${selectedItem.toQueryString()}`,
       {}, { replace: replaceUrl });
   }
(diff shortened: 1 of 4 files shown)


src/ui_components/search_input.js
```diff
@@ -85,7 +85,7 @@ export default class SearchInput {
     const results = await fetchSuggests(query, { withCategories: true });
     if (results && results.length > 0) {
       const firstResult = results[0];
-      selectItem(firstResult, true);
+      selectItem(firstResult, true, true);
```
G-Ray, line 88: the third param seems extra ?

src/libs/suggest.js
```diff
@@ -24,6 +25,15 @@ export const selectItem = (selectedItem, replaceUrl = false) => {
     window.app.navigateTo(`/places/?type=${selectedItem.name}`,
       {}, { replace: replaceUrl });
   } else if (selectedItem instanceof Intention) {
+    Telemetry.add(
+      Telemetry.SUGGEST_CLICK,
+      null,
+      null,
+      {
+        useNlu: geocoderConfig.useNlu,
+        isCategory: !!selectedItem.category,
```
bbecquet, line 34: @amatissart proposes to send the category itself, instead of just its presence, to allow for more analysis. I agree this would be useful.

amatissart: Yes the name of the category would be useful.  
And maybe some more details about the presence of a query or a place, such as 
```
hasFullTextQuery: !!selectedItem.fullTextQuery,
hasPlace: !!selectedItem.place
```

src/libs/suggest.js
```diff
@@ -24,6 +25,17 @@ export const selectItem = (selectedItem, replaceUrl = false) => {
...
   } else if (selectedItem instanceof Intention) {
+    Telemetry.add(
+      Telemetry.SUGGEST_CLICK,
+      null,
+      null,
+      {
+        useNlu: geocoderConfig.useNlu,
+        category: selectedItem.category ? selectedItem.category.name : null,
+        hasFullTextQuery: !!selectedItem.fullTextQuery,
+        hasPlace: !!selectedItem.place,
```
bbecquet, line 36: Just sending the `place` and `fullTextQuery` would be more flexible for the future, and more consistent with what we do with the `category`. What do you think @amatissart ?

amatissart: I suggested to use these flags, as they are precise enough to build the metrics that we plan to follow. The raw queries are already processed on the back-end.  
Moreover, `selectedItem.place` is an object here, so it would require some logic to be validated on the server. I am ok to keep it simple for now.

PR #686, opened 2020-06-22
Use Idunn extend_bbox feature on category panel

bbecquet:
*Alternative to https://github.com/QwantResearch/erdapfel/pull/675 with minimal changes, for easier review. I mixed far too many refacto with features in the other one…*

## Description

Allow the Idunn API call to take an `extend_bbox` parameter to use the auto BBOX extension feature introduced in https://github.com/QwantResearch/idunn/pull/146.
We use the bbox extension feature only when loading the first results of a category/query, not on subsequent map pans and zooms. This is to avoid changing the view when the user is interacting with the map. I re-used the `initialLoading` state as an indication of that.

## Why
Be more helpful to the user when no data is returned in the first place

## Screenshots
Peek 22-06-2020 10-46

### DIFF
--- a/src/adapters/poi/idunn_poi.js
+++ b/src/adapters/poi/idunn_poi.js
@@ -36,9 +36,9 @@ export default class IdunnPoi extends Poi {
   }
 
   /* ?bbox={bbox}&category=<category-name>&size={size}&verbosity=long/ */
-  static async poiCategoryLoad(bbox, size, category, query, extendBBox = false) {
+  static async poiCategoryLoad(bbox, size, category, query, extendBbox = false) {
     const url = `${serviceConfig.idunn.url}/v1/places`;
-    const requestParams = { bbox, size, extend_bbox: extendBBox };
+    const requestParams = { bbox, size, extend_bbox: extendBbox };
     if (category) {
       requestParams['category'] = category;
     }
--- a/src/adapters/poi/idunn_poi.js
+++ b/src/adapters/poi/idunn_poi.js
@@ -36,9 +36,9 @@ export default class IdunnPoi extends Poi {
   }
 
   /* ?bbox={bbox}&category=<category-name>&size={size}&verbosity=long/ */
-  static async poiCategoryLoad(bbox, size, category, query) {
+  static async poiCategoryLoad(bbox, size, category, query, extendBBox) {
     const url = `${serviceConfig.idunn.url}/v1/places`;
-    const requestParams = { bbox, size };
+    const requestParams = { bbox, size, extend_bbox: !!extendBBox };
     if (category) {
       requestParams['category'] = category;
     }
--- a/src/adapters/poi/idunn_poi.js
+++ b/src/adapters/poi/idunn_poi.js
@@ -36,9 +36,9 @@ export default class IdunnPoi extends Poi {
   }
 
   /* ?bbox={bbox}&category=<category-name>&size={size}&verbosity=long/ */
-  static async poiCategoryLoad(bbox, size, category, query, extendBBox) {
+  static async poiCategoryLoad(bbox, size, category, query, extendBBox = false) {
     const url = `${serviceConfig.idunn.url}/v1/places`;
-    const requestParams = { bbox, size, extend_bbox: !!extendBBox };
+    const requestParams = { bbox, size, extend_bbox: extendBBox };
     if (category) {
       requestParams['category'] = category;
     }
--- a/src/libs/bounds.js
+++ b/src/libs/bounds.js
@@ -6,6 +6,6 @@ export const parseBboxString = bboxString =>
 
 export const boundsToString = llBounds =>
   llBounds.toArray()
-    .flat()
+    .reduce((flatArray, current) => flatArray.concat(current), []) // flatten
     .map(coord => coord.toFixed(7))
     .join(',');
--- a/src/libs/bounds.js
+++ b/src/libs/bounds.js
@@ -1,0 +1,11 @@
+export const boundsFromFlatArray = coords =>
+  [[coords[0], coords[1]], [coords[2], coords[3]]];
+
+export const parseBboxString = bboxString =>
+  boundsFromFlatArray(bboxString.split(',').map(coord => Number(coord)));
+
+export const boundsToString = llBounds =>
+  llBounds.toArray()
+    .flat()
+    .map(coord => coord.toFixed(7))
+    .join(',');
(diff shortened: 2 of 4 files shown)


src/adapters/poi/idunn_poi.js
```diff
@@ -53,9 +53,9 @@ export default class IdunnPoi extends Poi {
     }
   }
   /* ?bbox={bbox}&category=<category-name>&size={size}&verbosity=long/ */
-  static async poiCategoryLoad(bbox, size, category, query) {
+  static async poiCategoryLoad(bbox, size, category, query, extendBBox) {
     const url = `${serviceConfig.idunn.url}/v1/places`;
-    const requestParams = { bbox, size };
+    const requestParams = { bbox, size, extend_bbox: !!extendBBox };
```
G-Ray, line 58: Can we take a boolean as parameter instead of casting in the function ?

src/adapters/poi/idunn_poi.js
```diff
@@ -53,9 +53,9 @@ export default class IdunnPoi extends Poi {
     }
   }
   /* ?bbox={bbox}&category=<category-name>&size={size}&verbosity=long/ */
-  static async poiCategoryLoad(bbox, size, category, query) {
+  static async poiCategoryLoad(bbox, size, category, query, extendBBox = false) {
     const url = `${serviceConfig.idunn.url}/v1/places`;
-    const requestParams = { bbox, size };
+    const requestParams = { bbox, size, extend_bbox: extendBBox };
```
G-Ray, line 58: Really minor thing: should we prefer `extendBBox` or `extendBbox` with a lowercase 'b' ? Like `parseBboxString`

src/libs/bounds.js
```diff
@@ -0,0 +1,11 @@
+export const boundsFromFlatArray = coords =>
```
G-Ray, line 1: :+1 for file

src/panel/category/CategoryPanel.jsx
```diff
@@ -37,28 +38,22 @@ export default class CategoryPanel extends React.Component {
...
-    for (const key in poiFilters) {
-      if (poiFilters[key] !== prevProps.poiFilters[key]) {
-        this.fetchData();
-        break;
-      }
-    }
-
-    if (bbox && bbox !== prevProps.bbox) {
-      window.execOnMapLoaded(() => { this.fitMapAndFetch(); });
+    if (JSON.stringify(prevProps.poiFilters) !== JSON.stringify(this.props.poiFilters)
```
G-Ray, line 52: Nice trick, `_.isEqual` from lodash, (or other deep equal function), is probably more performant for bigger objects though

src/adapters/poi/idunn_poi.js
bbecquet: Looks like we have more `Bbox` than `BBox` already, which is consistent with the frequent use of the standalone `bbox`. So I'll standardize these cases to `Bbox`.
Thanks for making this point :)

src/panel/category/CategoryPanel.jsx
```diff
@@ -130,6 +121,12 @@ export default class CategoryPanel extends React.Component {
       initialLoading: false,
     });
 
+    if (bbox_extended) {
+      // The returned bbox is sure to contain at least one POI.
+      // Extend the current one to include it.
+      window.map.mb.fitBounds(currentBounds.extend(boundsFromFlatArray(contentBbox)));
```
amatissart, line 127: Don't we need to take margins into account when using `fitBounds` ?  
(here and also on line 78)

amatissart: Related to this question: is it expected that a query that returned an extended bbox triggers a second request to the API ? Here `fitBounds` will cause a `map_moveend` event, which fetch data again.  
It's probably not an issue in most cases (although the second query is probably redundant), but this may cause flickering if some results are close to the edge of the screen, and margins are not handled properly here.

bbecquet: > is it expected that a query that returned an extended bbox triggers a second request to the API ? Here fitBounds will cause a map_moveend event, which fetch data again.

Yes I noticed that too. Do you see a way to prevent this if we keep this behavior of joining the current bounds and the extended ones?

bbecquet: @amatissart I adapted and used an existing call that supports padding to do the "fit bounds" operations. https://github.com/QwantResearch/erdapfel/pull/686/commits/09eb1120ee13d87fa7a28e80cee159cc62bff91d

amatissart: It looks good to me.  
You're right that the second query is necessary at least in some cases, the current behavior is probably acceptable :+1:

src/panel/category/CategoryPanel.jsx
```diff
@@ -102,26 +96,23 @@ export default class CategoryPanel extends React.Component {
       window.map.mb.flyTo({ center: [2.35, 48.85], zoom: 12 });
     } else if (currentZoom < 12) { // Zoom < 12: zoom up to zoom 12
       window.map.mb.flyTo({ zoom: 12 });
-    } else if (currentZoom > 16) { // Zoom > 16: dezoom to zoom 16
-      window.map.mb.flyTo({ zoom: 16 });
```
amatissart, line 106: This rule could actually still be useful to avoid very large zoom out (maybe for z>18 only ?)


I understand that we want to avoid multiple zoom changes, but they can still happen in some cases with the previous rule (z=10 :arrow_right: z=12 :arrow_right: extended bbox to z=11)

bbecquet: I restored the rule but changed it to z>18 as you proposed.

src/panel/category/CategoryPanel.jsx
```diff
@@ -78,11 +73,10 @@ export default class CategoryPanel extends React.Component {
...
   }
 
-  fitMapAndFetch() {
-    const rawBbox = (this.props.bbox || '').split(',');
-    const bbox = rawBbox.length === 4 && [[rawBbox[0], rawBbox[1]], [rawBbox[2], rawBbox[3]]];
-    if (bbox) {
-      window.map.mb.fitBounds(bbox, { animate: false });
+  fitMap() {
+    if (this.props.bbox) {
+      fire('fit_map', parseBboxString(this.props.bbox), true);
```
amatissart, line 78: As a consequence of `forceAnimate: true` any Intention with bbox with trigger an animated "flyTo" to fit the map. 
I find this acceptable, but want to be sure that we are aware of this.

bbecquet: Good catch, I removed this one and kept the animation on the bbox extension.

PR #687, opened 2020-06-22
PoiCard: move mobile button on resize

G-Ray:
Geolocate control was previously not resized when changing window's size. This caused some UX issues where the button would be under the poi's card.

Peek 2020-06-22 18-06


G-Ray:
PoiCard has been removed

PR #689, opened 2020-06-23
QMAPS-1232: error message in the search/itinerary suggests list if it's empty

xem:
## Description
Use the message: "No result found" / "Aucun résultat trouvé" when the suggest list is empty.
(in Directions panel, empty means there's only 1 item: "Your position", that is always visible)

## Screenshots
image

image

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -21,7 +21,7 @@ const Suggest = ({
   const [items, setItems] = useState([]);
   const [isOpen, setIsOpen] = useState(false);
   const [lastQuery, setLastQuery] = useState('');
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
   const isMobile = useContext(DeviceContext);
   let currentQuery = null;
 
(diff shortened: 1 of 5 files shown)


src/components/ui/Suggest.jsx
```diff
@@ -22,6 +22,7 @@ const Suggest = ({
   const [isOpen, setIsOpen] = useState(false);
   const [lastQuery, setLastQuery] = useState('');
   const isMobile = useContext(DeviceContext);
+  let isLoading = false;
```
bbecquet, line 25: `isLoading` should be managed as a state value of the component.
Sorry for the bad example of `currentQuery`, this is a special case where we need to keep a kind of singleton independantly from the render cycle. It could probably be done differently, but anyway…
For your need, `useState` is the way to go :)

tests/integration/tests/autocomplete.js
```diff
@@ -45,6 +45,13 @@ test('search and clear', async () => {
   expect(searchValueAfterClear).toEqual('');
 });
 
+test('search with no suggest', async () => {
+  responseHandler.addPreparedResponse(mockAutocompleteEmpty, /autocomplete\?q=Goodbye/);
+  await page.goto(APP_URL);
+  await autocompleteHelper.typeAndWait('Goodbye');
+  await page.waitForSelector('.autocomplete_suggestion--no-result');
```
bbecquet, line 52: Tests should end with an `expect`, otherwise when they fail they exit with a dirty, hard-to-read error (`Target closed`, etc.), instead of clear message explaining the difference between what we want and what we got.
For existence tests, you can use
```js
expect(await exists(page, '.autocomplete_suggestion--no-result')).toBeTruthy();`
```

PR #690, opened 2020-06-23
Update categories colors, with background

G-Ray:


src/scss/main.scss
```diff
@@ -1,4 +1,5 @@
 @import "includes/font";
+@import "includes/colors";
```
bbecquet, line 2: :+1: thank you!

src/scss/includes/colors.scss
```diff
@@ -0,0 +1,49 @@
...
+/* separator */
+$separator_color : #e0e1e6;
+
+/* error  */
+$error_background : #f8d7da;
+$error_border : #f5c6cb;
+$error_text : #721c24;
+
+:root {
+  --orange-base: #ff9b3c;
```
bbecquet, line 31: Sorry, I love real CSS vars too, but as long as we support IE11 we can't use them and must keep to Sass vars.

src/components/CategoryList.jsx
```diff
@@ -11,7 +11,10 @@ const CategoryList = ({ className }) =>
         variant="category"
         label={category.label}
         icon={category.iconName}
-        iconStyle={{ backgroundColor: category.backgroundColor }}
+        iconStyle={{
+          color: `var(--${category.color})`,
+          backgroundColor: `var(--${category.backgroundColor})`,
```
bbecquet, line 16: I notice you finally decided to declare both colors instead of generating one as a variant of the other. I'm curious, why? Actually I thought in this PR you would rename `backgroundColor` to `color` and just deduce the new background from it.

G-Ray: Yeah this is what I wanted to achieve too. However, I couldn't find a good way to import the (s)css variables in JS with our webpack setup. Do you have any idea ?

bbecquet: > Do you have any idea ?

Actually I don't see where these values would be useful in CSS, so I don't get the need for a change. 
IMO the current setup with the values defined in the YAML file is good enough for the quantity of data and the frequency of changes.

src/scss/includes/colors.scss
G-Ray: What about adding a ponyfill for ie support, like https://github.com/jhildenbiddle/css-vars-ponyfill ?

bbecquet: I refer to my previous comment https://github.com/QwantResearch/erdapfel/pull/690#discussion_r446197048: I don't see the need of these vars on the CSS side (and any dependency we don't need is a win).

package.json
```diff
@@ -88,6 +88,7 @@
     "classnames": "^2.2.6",
     "color": "^3.1.2",
     "compression": "^1.7.3",
+    "dotenv": "^8.2.0",
```
bbecquet, line 91: This comes from the dotenv PR :)

PR #691, opened 2020-06-23
Qmaps 1507 search bar redesign

bbecquet:
## Description
Implement the new simplistic design for the application top bar.

Cleaned-up many things in the process, in particular complex CSS for the current bar style.

Compared to the designs linked in the ticket, here are some compromises we agreed on:
 - Suggestions are still displayed in full screen on mobile, and hide completely the current panel on desktop
 - The "directions" button which appear in the designs on the right of the search bar will be moved there when the service panel is redesigned
 - On "directions" mode, the top bar simply disappears

## Screenshots
|Before|After|
|---|---|
|localhost_3000_ (4)|localhost_3000_place_latlon_48 87270_2 28112|
|Capture d’écran de 2020-06-30 17-04-42|Capture d’écran de 2020-07-06 12-38-13|
|localhost_3000_ (5)|localhost_3000_place_latlon_48 87270_2 28112 (1)|
|localhost_3000_(Pixel 2)|localhost_3000_place_latlon_48 87270_2 28112(Pixel 2)|
|Capture d’écran de 2020-06-30 17-05-43|Capture d’écran de 2020-07-06 12-36-24|


src/components/ui/Suggest.jsx
```diff
@@ -30,6 +30,14 @@ const Suggest = ({
     setItems([]);
   };
 
+  useEffect(() => {
+    if (isOpen) {
+      document.body.classList.add('suggestions-open');
```
G-Ray, line 35: Why is it needed ?

src/panel/PanelManager.jsx
```diff
@@ -127,20 +116,6 @@ export default class PanelManager extends React.Component {
     router.routeUrl(getCurrentUrl());
   }
 
-  toggleMinify = () => {
```
G-Ray, line 130: :open_mouth:

src/components/ui/Suggest.jsx
bbecquet: To hide the panel, here https://github.com/QwantResearch/erdapfel/pull/691/files#diff-46b992b598e848147f038499405bff57R1

G-Ray:
Could we remove the padding around the input field ? This way, the focus could be more accessible

xem:
Hi!

The code seems super nice and clean!

My (visual) feedback after testing locally:

1) The panel's border radius (12px) seems exaggerated, especially when a scrollbar is present: the bottom right of the scrollbar gets cropped with the important radius. 
Firefox:
image
Chrome:
image
Can we negociate a smaller (ex: 5px) radius with the designers?

2) Is it normal that I can never see the "large" version of the search input on desktop? Only the small one remains at all times? Feels a bit disturbing at first :)
image

3) The gradient border radius that appears on focus is not displayed properly on my laptop (W10, Chrome & Fx), the bottom is thinner than the top. In mobile simulation, it's ok.
and 4) The line-height inside the field is too short in height, when I click in the grey area outside of the blue zone shown below, I lose focus
image

5) Maybe not your fault, but these icons are not centered vertically
image

src/components/ui/Suggest.jsx
G-Ray: Ok, could we come up with a solution which would not involve adding a class in the suggest component ? Maybe using an event (Suggest.js should not know any panel's implementation), a "global store" or simply using a "panel" lib which would expose a "toggle" function ?

bbecquet: > I think Suggest.jsx should not know nothing about panel's implementation.

Technically it knows nothing here, it just sets a state on the document in a clearly separated side-effect, wuthout having to now what's made of it. Actually using an imperative panel toggling lib would be more coupled than that :)

> Maybe using an event, a "global store" or simply using a "panel" lib which would expose a "toggle" function ?

I chose this class solution because this is the easier one to toggle the visibility in CSS only. I don't want to toggle panels with conditional rendering using React, as it won't be as fast as needed and some panels will request some info when remounting.

Sure I could use an event, but the listener code would probably be the same. Would you like it better?

bbecquet:
@G-Ray 

I changed the input field padding, it's now easier to use. Thanks for the great suggestion! :+1: (and thanks flexbox for making it easy :))

|Before|After|
|---|---|
|Capture d’écran de 2020-07-02 14-33-58|Capture d’écran de 2020-07-02 14-33-26|

src/components/ui/Suggest.jsx
G-Ray: The only thing bothering me is that I would prefer a (mini) layer of abstraction, I would prefer not having to know a "suggestions-open" class has to be set/unset inside `Suggest.jsx` directly, because if we change the way we hide the panel one day, we would need to edit `Suggest.jsx` instead of a file like `lib/panel.js` for instance.

bbecquet: I understand. So, what do you think about firing an event like `'toggle-suggestions'` in `Suggest.jsx` and listening to it in `PanelManager.jsx`? I can't see a better place to "manage panels" :smiley:

G-Ray: I think we can avoid events, what about creating a `lib/panel.js` with two exported functions: `open` `close` (or hide).
Each one of these function would set/unset the class `suggestions-open`. This way, any part or our app wanting to open/hide the panel would call `panel.close()` or `panel.open()`.

bbecquet:
@xem Thanks for the review, I'm answering point by point:

1. I agree the new border radius seems a bit out of place right now on panels, especially wrt. scroll bars and the small color bar at the top of service and direction panels. But this is confirmed to be part of the design of the new top bar, and I didn't want to merge something that appeared too inconsistent. As the very next step will be to simplify (the service panel, I intend to address this kind of issues with the next ticket. Maybe when we have every pieces in place we can discuss with the design team to convince them these are too big? (although I'm personnaly growing fond of them :))

2. Yep, no more "large" version at all. I guess the point is to have a design that appears less unbalanced and give more room to the map itself.

3. Thanks for noticing, I was almost sure something like that would happen with this hack on some configs :/ Do you think it's blocking? If so, we can go with a more simple approach and use a border until we have a more reliable solution for the gradient.

4. Thanks, I addressed that https://github.com/QwantResearch/erdapfel/pull/691#issuecomment-652981531 :+1: 

5. Not related, and these two buttons will disappear very soon, so :man_shrugging:

src/scss/includes/search_form.scss
```diff
@@ -227,162 +233,18 @@ input[type="search"] {
...
-}
-
-@media (max-width: 380px) {
-  .search_form__input {
-    letter-spacing: -.5px;
-  }
-
-  &::placeholder {
-    font-size: 14px;
   }
```
amatissart, line 248: As a result, the placeholder is truncated on some mobiles:
image

(Here on Iphone 6/7/8 simulated by Chromium)

bbecquet: What do you think we should do? Having the same font-size for the placeholder and the content is an improvement to me. We can reduce the padding of the magnifiying glass button, but it would reduce its accessibilty. And the problem may be even bigger for other languages, I doubt we can have a solution working for 100% of the cases…
But I agree we should improve this use case which is probably one of the most frequent.

bbecquet:
@amatissart Thanks for the review!

 - For the panel height, it looks like the bug was introduced before, as it exists in prod.
I'll try to fix it here if it's trivial, otherwise I'll make a separate PR.

 - I fixed the icon issue of category suggestion items:
Capture d’écran de 2020-07-03 17-54-31
I forgot to test with `useNlu=false`, I was already in the future :)

src/scss/includes/search_form.scss
amatissart: If we no longer want to adapt the font-size or letter-spacing, the simplest solution would be to hide the placeholder on small screens.
This is already done for `max-width: 340px` (just a few lines below). We could update this threshold to around 400-450px to be safe.

Technically, it would be acceptable to me. It would make the brand "Qwant Maps" less visible though.

src/libs/panel.js
```diff
@@ -0,0 +1,8 @@
+
+export function togglePanelVisibility(visible) {
```
G-Ray, line 2: :+1:

src/scss/includes/search_form.scss
bbecquet: I adjusted some paddings to win the missing pixels, I think it's better than deciding to hide the string on so many devices. Here is the result on the IPhone 6/7/8 simulator in Chromium:
localhost_3000_(iPhone 6_7_8) (4)

And the before/after on my personal phone (360/640):
|Before|After|
|---|---|
|Screenshot_20200706-101309_Chrome|Screenshot_20200706-101023_Chrome|

src/scss/includes/top_bar.scss
```diff
@@ -48,12 +21,9 @@
     top: 0;
     left: 0;
     width: 100%;
-    background: url('../images/backgrounds/top_mobile.svg') no-repeat;
```
G-Ray, line 51: Can we delete this file too :smiley:

bbecquet: We can't, it's still use for the mobile direction panel, which hasn't been redesigned yet.

G-Ray: ah ok, let's be patient then

PR #694, opened 2020-06-26
Load .env with dotenv

G-Ray:
## Description
Load environment file `.env` with `dotenv` (https://github.com/motdotla/dotenv#readme).

## Why
Simplify starting the app (`npm start` script) without having to export environment variables manually.
Simplify .env using syntax like
```sh
TILEVIEW_mapStyle_poiMapUrl=["https://tileserver/maps/tiles/ozpoi/{z}/{x}/{y}.pbf"]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ bench/report.json
 bench/results.har
 
 stats.json
+
+.env
+.env.*
--- a/README.md
+++ b/README.md
@@ -27,7 +27,9 @@ system:
   timeout: 5
 ```
 
-is overriden by `TILEVIEW_system_timeout=3`
+**Loading .env file**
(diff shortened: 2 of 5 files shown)


amatissart:
Thanks for this :+1:

I think that `/.env` should be added to `.gitignore` to avoid commiting some private configuration by mistake.  
And maybe this new possibility should be described in the README file ?

bbecquet:
Good idea.
Would it have implications for the build/deploy process? I guess if the `.env` doesn't exist, nothing happens and the real env vars are the ones taken, but just being sure.

In any case, I would add `.env` and `.env.*` to `.gitignore`, so you can do whatever you want with files on your local folder.

PR #695, opened 2020-06-29
Qmaps 1599 - Fix markers

G-Ray:
## Description
* Add a precondition to avoid setting same poi marker multiple times
* When closing DirectionPanel, go to `/` route, ignoring a previously selected poi.
* Reload categories pois and poiFilters from `history.state` when refeshing app.

When closing directionPanel, expected behavior is now to go back to the default panel.

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -17,7 +17,6 @@ import { parseMapHash, getMapHash } from 'src/libs/url_utils';
 import { toUrl, getBestZoom } from 'src/libs/pois';
 import Error from 'src/adapters/error';
 import { fire, listen } from 'src/libs/customEvents';
-import { isNullOrEmpty } from 'src/libs/object';
 import locale from '../mapbox/locale';
 
 const baseUrl = nconf.get().system.baseUrl;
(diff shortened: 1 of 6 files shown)


G-Ray:
@bbecquet I removed the last commit about removing markers when unmounting `CategoryPanel`. We can review the remaining commits to confirms the fixed behavior.

src/adapters/scene.js
```diff
@@ -333,6 +328,11 @@ Scene.prototype.ensureMarkerIsVisible = function(poi, options) {
 };
 
 Scene.prototype.addMarker = function(poi) {
+  if (this.currentMarker?.latLon === poi.latLon) {
+    // Prevent setting the same poi multiple times
+    return this.currentMarker;
+  }
+
```
amatissart, line 335: This doesn't seem to have any effect, as `.latLon` is not a property of mapbox-gl `Marker`.
And this approach inspecting `this.currentMarker` does not look correct to me. What we need to check is that the map still contains a list of markers (stored by `SceneCategory`). They are distinct from `this.currentMarker`.

Here is a gif showing a problem with the proposed implementation:
Peek 01-07-2020 18-23

G-Ray: You right, I updated the relevant code

src/adapters/scene.js
```diff
@@ -357,6 +361,7 @@ Scene.prototype.addMarker = function(poi) {
 Scene.prototype.cleanMarker = async function() {
   if (this.currentMarker !== null) {
     this.currentMarker.remove();
+    this.currentMarker = null;
```
G-Ray, line 364: This cleanup was missing

src/panel/PanelManager.jsx
```diff
@@ -139,7 +139,7 @@ export default class PanelManager extends React.Component {
     });
 
     // Route the initial URL
-    return router.routeUrl(getCurrentUrl());
+    return router.routeUrl(getCurrentUrl(), window.history.state || {});
```
bbecquet, line 142: :+1:

PR #697, opened 2020-06-30
Telemetry: fix "itinerary_share" sent from route results and favorites

amatissart:
## Description
After recent refactoring in #663, the "itinerary_share" event was sent from the ShareMenu component.

This PR defines a wrapper on the click handler to trigger a specific telemetry event in each context where the share menu is used.


src/panel/direction/RouteSummary.jsx
```diff
@@ -29,6 +31,11 @@ export default class RouteSummary extends React.Component {
     this.props.openPreview(this.props.id);
   }
 
+  onShareClick = handler => e => {
```
bbecquet, line 34: :+1:

amatissart:
After discussion with @G-Ray I replaced the double arrow functions with more explicit functions for readability.  
I also defined a new `poi_share` event called by the action button, similar to `itinerary_share` and `favorite_share`.

PR #699, opened 2020-07-01
QMAPS-1505 center map on user region on their first visit

xem:
## Description
Done:
- Create a middleware "initial_position.js"
- Load this middleware in "app.js", and make the `initialBbox` array global.
- Store this array in "window" in index.ejs
- Read the value of this array in "scene.js" and use it instead of the value stored in "constants.yml" as the default position in the first visit (ie. if no hash and no localStorage is present).

Todo:

- [x] Headers `x-geoip-country-code` & `x-geoip-region-code` are used to determine the user region
- [x] `config/regions.yaml` contains all bboxes of all the regions (or countries), and send the right bbox to the user, instead of the default position
- [x] In all cases, we need to choose a maximum zoom level (I suggest 8) to be used when the map is fitted to a position or region.


amatissart:
I updated this PR to include a file `regions.yml` that defines a bbox for each country and region codes.

If a code is found in the headers, its bbox is used to set the initial bounds of the map, with a maxZoom value as @xem suggested.

src/adapters/scene.js
```diff
@@ -46,6 +47,8 @@ Scene.prototype.setupInitialPosition = async function(locationHash) {
     if (lastLocation) {
       this.center = [lastLocation.lng, lastLocation.lat];
       this.zoom = lastLocation.zoom;
+    } else if (window.initialBbox) {
```
bbecquet, line 50: What do you think about rewriting this to a function returning directly an object to be merged with the Map options object., instead of storing stuff in `this` (which seem useless after initialization)? 
It would be called directly by `initMapBox`. It would make the whole init process easier to read and test.

amatissart:
@xem GitHub does not let me request your review as your opened this PR originally, but your comments are welcome on the latest changes.

xem:
Hey, of course!
I din't really speak python, though the PR code is clean, congrats! I think all our ideas for this feature are finally gathered in a nuce PR \o/

src/adapters/scene.js
```diff
@@ -26,30 +26,39 @@ const LONG_TOUCH_DELAY_MS = 500;
...
+    return {
+      bounds: window.initialBbox,
+      fitBoundsOptions: {
+        padding: this.getCurrentPaddings(),
+        maxZoom: 9,
+      },
+    };
+  }
+  return {
+    zoom: map.zoom,
```
bbecquet, line 56: Cool simplification of the map init code, it's simpler to follow this way. The only thing I had to trace back in the code was this `map` constant, which is ambiguously named in this context. What do you think about renaming it to `mapConfig` on import while we are changing this code?

amatissart: Sure, it's done.

bbecquet (approved):

PR #704, opened 2020-07-06
Redesign service panel

bbecquet:
## Description
Implement the new simplified design of the service panel:
 - style consistent with the new top bar
 - new order for featured categories
 - new font sizes and alignments
 - new hover style for categories
 - remove the Favorites button and put a redesigned direction button in the new top bar (also changes the one on mobile to be consistent)
 - fixes the panel max-height bug.
*Note: This PR doesn't implement the collapsable list of categories, from 4 to the full list. This will be done in a separate PR*

## Screenshots
|Before|After|
|---|---|
|Capture d’écran - 2020-07-06 à 16 58 19|Capture d’écran - 2020-07-06 à 16 58 32|
|Capture d’écran - 2020-07-06 à 16 59 26|Capture d’écran - 2020-07-06 à 16 58 44|
|Capture d’écran - 2020-07-06 à 16 57 42|Capture d’écran - 2020-07-06 à 16 57 20|


xem:
Petit souci de height sur les poi lists depuis ce redesign: ils dépassent de l'écran au lieu d'avoir une max-height qui les fait se terminer un peu avant
image

bbecquet:
@xem about the panel height bug, do you remember how you got that result?
I explicitely fixed an existing bug (in prod) about that in the fifth commit https://github.com/QwantResearch/erdapfel/pull/704/commits/56eacdf5676450acf8f00922242eae8587ff6e58, but I may have overlooked some cases.

xem:
Sorry, I was still checking on the master branch, on 1588 it is fixed indeed... :)
the truncated scrollbar remains a ux problem for me (but we have time to discuss that with design at the end)
image

tests/integration/favorites_tools.js
```diff
@@ -1,5 +1,7 @@
+/* global PANEL_ANIMATION_DELAY_MS */
```
G-Ray, line 1: I'm not sure this is the best way to store our variables, as the number could grow over the time.
Should'nt we share a module and import it into the relevant tests files ?

bbecquet: Same problem you met lately: this is something defined only in CSS, not in JS.
Maybe there are alternatives and I'm open to it, as long as it doesn't get overkill :) (here it's just for the tests)

For example, we can wait for the transition without knowing its actual duration, but it needs some work :)
https://stackoverflow.com/questions/51547994/wait-for-transition-to-end-in-puppeteer
WDYT?

src/scss/includes/component.scss
```diff
@@ -2,6 +2,7 @@
 
 $panel_width : 400px;
 $top_bar_height: 64px;
+$ui_margin: 12px;
```
G-Ray, line 5: Shouldn't this be `u-m-12` in `utils.scss` ?

Not necessarily in the context of this PR, but I think we should define spaces of the design system.

bbecquet: This is different, for two reasons:
 - this is an individual value, not a class like `.u-mb-12` and such ;
 - more importantly, it has a semantic (it's the main external margin of the whole UI), it's not just some generic util.

But I agree we should have a discussion about vars and utils in our own small version of the design system. For example, I noticed the Zeplin designs use named size values, like `-xl`, `-xs`, etc. Maybe we should move to that eventually. We could then have these as vars, and utils using them. For example:
```scss
$size-s: 8px;
$size-l: 16px;
// ...
.u-mb-s {
   margin-bottom: $size-s;
}
.u-mb-l {
   margin-bottom: $size-l;
}
.u-mr-s {
   margin-right: $size-s;
}
.u-mr-l {
   margin-right: $size-l;
}
// ...
```
Or just see how it's done in the "real" DS and just copy/paste it.
Let's organize something to talk about it:)

tests/integration/favorites_tools.js
G-Ray: waiting for the transition to complete seems to be the more generic solution. Can we give it a try using the transition event ? And maybe define an util function in our helpers ?

views/index.ejs
```diff
@@ -49,6 +49,7 @@
           <input type="submit" value="" class="search_form__action" onclick="submitSearch()" title="<%= _('Search') %>">
         </div>
         <div id="react_menu__container"></div>
+        <button class="search_form__direction_shortcut" title="<%= _('Directions', 'top bar') %>"></button>
```
amatissart, line 52: What if Directions are disabled ?  
Should this button be disabled ? or hidden ?

bbecquet: I think the best solution is to hide it. I'll do it, thanks!

tests/integration/favorites_tools.js
bbecquet: Ok, did it. At first I thought it didn't work, but it does :) You just have to listen to the right event, in this case it wasn't `transitionend` but `animationend`, because of the way this animation is declared in CSS.

views/index.ejs
bbecquet: Done!

|Enabled|Disabled|
|---|---|
|Capture d’écran de 2020-07-09 14-46-57|Capture d’écran de 2020-07-09 14-46-01|

G-Ray:
Should we add the close button ?
image

G-Ray:
The final render is neat :+1:

bbecquet:
> Should we add the close button ?

Yet another recent change, the previous close button was inside the input :(…
I'll do it later, let's keep this PR focused on the panel, the direction shortcut is here only because we removed the one from the ServicePanel.

G-Ray:
Screenshot from 2020-07-09 17-56-40
Can we increase the burger menu button zone to improve accessibility on mobile ?

bbecquet:
@G-Ray 

> Can we increase the width of burger menu button to improve accessibility on mobile ?

Done. I used the same trick as the direction shortcut.
Capture d’écran de 2020-07-09 18-15-03

tests/integration/tools.js
```diff
@@ -81,3 +81,14 @@ export async function isHidden(page, selector) {
     return false;
   }
 }
+
+export async function waitForAnimationEnd(page, selector) {
```
G-Ray, line 85: :+1:

PR #707, opened 2020-07-07
Qmaps 1232 no suggest error message (fix)

xem:
## Description
- fix error message style (introduce an ErrorLabel suggest item type)
- ~~fix line-height in general for the "normal" suggest items (letters can get truncated sometimes, esp. on W10 + System zoom >= 125%)~~ (Edit by @bbecquet: design team is investigating how to fix that in a generic way, as it seems the problem is with SegoeUI font everywhere. Let's fix that separately)

## Screenshots
image

### DIFF
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -254,6 +254,12 @@ input[type="search"] {
   padding: 9px 18px;
 }
 
+.autocomplete_error {
+  padding: 15px 10px 5px;
+  color: $primary_text;
+  cursor: default;
+}
+
 @media (max-width: 640px) {
   .search_form__wrapper {
     padding: 0 9px 0 12px;
--- a/src/components/ui/SuggestItem.jsx
+++ b/src/components/ui/SuggestItem.jsx
@@ -71,9 +71,18 @@ const SeparatorLabel = ({ label }) =>
     {label}
   </h3>;
 
+const ErrorLabel = ({ label }) =>
+  <div className="autocomplete_error">
+    {label}
+  </div>;
+
 const SuggestItem = ({ item }) => {
   if (item.simpleLabel) {
     return <SeparatorLabel label={item.simpleLabel} />;
+  }
+
+  if (item.errorLabel) {
+    return <ErrorLabel label={item.errorLabel} />;
   }
 
   if (item instanceof NavigatorGeolocalisationPoi) {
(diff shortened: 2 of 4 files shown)


src/components/ui/Suggest.jsx
```diff
@@ -28,8 +28,8 @@ const Suggest = ({
   let currentQuery = null;
 
   const close = () => {
-    setIsOpen(false);
-    setItems([]);
+    //setIsOpen(false);
```
bbecquet, line 31: This should be removed from the PR.

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -1,3 +1,4 @@
+/* global _ */
```
bbecquet, line 1: All changes in this file should be reverted, as the error message condition is controlled from outside now, this dropdown is just rendering stuff without knowing much about them.

src/scss/includes/search_form.scss
```diff
@@ -254,6 +256,12 @@ input[type="search"] {
   padding: 9px 18px;
 }
 
+.autocomplete_error {
+  padding: 15px 10px 5px;
+  font-size: 16px;
+  color: $primary_text;
+}
```
bbecquet, line 263: A `cursor: default` rule is probably required like for the separator, so the user doesn't think this is an active element.

src/scss/includes/search_form.scss
```diff
@@ -254,6 +256,13 @@ input[type="search"] {
   padding: 9px 18px;
 }
 
+.autocomplete_error {
+  padding: 15px 10px 5px;
+  font-size: 16px;
```
bbecquet, line 261: I don't think it's useful to redefine this `font-size`, as it's already `16px` on the parent element.

src/scss/includes/search_form.scss
```diff
@@ -233,11 +233,13 @@ input[type="search"] {
 
     &.autocomplete_suggestion__first_line {
       color: $primary_text;
+      line-height: 21px;
     }
 
     &.autocomplete_suggestion__second_line {
       color: $secondary_text;
       font-size: 14px;
+      line-height: 18px;
```
bbecquet, line 242: Not a big deal, but has this been approved by the design team yet?
In the meantime, we can remove this change from this PR if we want to merge.

PR #708, opened 2020-07-08
Alert component

G-Ray:
# Mobile
mobile
# Desktop
desktop


src/components/ui/MasqAlert.jsx
```diff
@@ -0,0 +1,41 @@
+/* global _ */
+import React, { useState } from 'react';
+import nconf from '@qwant/nconf-getter';
+import Alert from 'src/components/ui/Alert';
+
+const masqAlertEnabled = nconf.get().masq.alertEnabled;
+const masqLink = nconf.get().masq.link;
+const masqAlertDate = (new Date(nconf.get().masq.alertDate)).toDateString();
```
amatissart, line 8: `toDateString()` does not use the current language.  
A more user-readable format could be used with `Intl.DateTimeFormat`.

src/components/ui/MasqAlert.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const MasqAlert = () => {
+  const masqDismissed = window.localStorage.getItem('masq_alert_dismissed');
```
amatissart, line 11: Is there not a risk that `localStorage.getItem()`  raises an error if `localStorage` is not available ?  

It would also be safer to not use the localStorage at all if the alert is disabled by the configuration.

src/components/ui/MasqAlert.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const [masqAlertDate, setMasqAlertDate] = useState('');
+  const [isVisible, setIsVisible] = useState(masqAlertEnabled && masqDismissed !== 'true');
+  if (!isVisible) {
+    return null;
+  }
+
+  useEffect(() => {
+    setMasqAlertDate(Intl
+      .DateTimeFormat(window.getLang().locale.replace('_', '-'))
```
amatissart, line 29: The default format uses only numeric values, and it may be unclear for users whether the GB or US format is used here.

```suggestion
      .DateTimeFormat(window.getLang().locale.replace('_', '-'), {day: "numeric", month: "long", year: "numeric"})
```

src/components/ui/MasqAlert.jsx
```diff
@@ -0,0 +1,62 @@
...
+        'masq',
+        { masqAlertDate }
+      )}
+      <a href={ masqLink } rel="noopener noreferrer">
+        {` ${_('Learn more', 'masq')}`}
+      </a>
+      </span>
+    }
+    type="info"
+    onClose={dismiss}
```
amatissart, line 57: A React error is visible in the console when the dismiss button is clicked:

```
Uncaught Invariant Violation: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

src/components/ui/MasqAlert.jsx
```diff
@@ -0,0 +1,62 @@
...
+    title={_('Masq will be disabled', 'masq')}
+    description={
+      <span>{_(
+        'Masq by Qwant will be disabled starting from {masqAlertDate}.',
+        'masq',
+        { masqAlertDate }
+      )}
+      <a href={ masqLink } rel="noopener noreferrer">
+        {` ${_('Learn more', 'masq')}`}
+      </a>
```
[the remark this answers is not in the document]
G-Ray: thanks ! The color seems to already be used in the masq app, I can submit it to the design team to get their opinion

PR #709, opened 2020-07-10
Service panel redesign: collapsible category list on desktop

bbecquet:
## Description
Implement the new default view of the service panel displaying only the 4 first categories and a button to show/hide the remaining ones.

*For now on desktop only, as it's still unclear on which screen should have that on mobile.*

## Screenshots
|Collapsed|Open|
|---|---|
|localhost_3000_ (9)|localhost_3000_ (10)|

### DIFF
--- a/src/scss/includes/components/button.scss
+++ b/src/scss/includes/components/button.scss
@@ -14,6 +14,12 @@
     color: $action-blue-base;
   }
 
+  &--shadow {
+    background-color: white;
+    color: $action-blue-base;
+    box-shadow: 0 1px 4px 0 rgba(12, 12, 14, 0.2), 0 0 2px 0 rgba(12, 12, 14, 0.12);
+  }
+
   &--primary {
     background-color: $action-blue-base;
     border: 1px solid $action-blue-base;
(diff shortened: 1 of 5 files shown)


src/scss/includes/components/button.scss
```diff
@@ -14,6 +14,12 @@
     color: $action-blue-base;
   }
 
+  &--shadow {
```
G-Ray, line 17: Do you think we could set a shadow independently of the variant ?
I mean, this props could be composed : we could have a variant with, or without a shadow props maybe ?

bbecquet: It makes sense, but here this would require defining first a variant without a border, then having a prop and CSS class for the shadow. If you think it's worth it I'm ok with it.

src/scss/includes/components/button.scss
```diff
@@ -37,13 +42,16 @@
...
-    margin-left: 6px;
+    margin-left: 4px;
   }
 
   &::-moz-focus-inner {
     border: 0;
   }
 
+  &:not(.button--primary):hover {
+    background: $action-lighter;
```
G-Ray, line 53: On touch devices we can add `:active` to avoid the hover state getting stucked once clicked

PR #713, opened 2020-07-16
suggest: Do not return errors if no query

G-Ray:
## Description
Fix "no result" label being displayed when there is no query typed and there is no favorites stored.

### DIFF
--- a/src/libs/suggest.js
+++ b/src/libs/suggest.js
@@ -68,7 +68,7 @@ export const fetchSuggests = (query, options = {}) =>
     maxFavorites: options.maxFavorites || !query ? 5 : 2,
   });
 
-export const modifyList = (items, withGeoloc) => {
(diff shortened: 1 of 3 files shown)


src/libs/suggest.js
```diff
@@ -79,7 +79,7 @@ export const modifyList = (items, withGeoloc) => {
     items.splice(0, 0, NavigatorGeolocalisationPoi.getInstance());
   }
 
-  if (items.length === 0 || items.length === 1 && withGeoloc) {
+  if (query.length > 0 && items.length === 0 || items.length === 1 && withGeoloc) {
```
amatissart, line 82: Similarly, the error should not be returned on empty input with geoloc:
```suggestion
  if (query.length > 0 && (items.length === 0 || items.length === 1 && withGeoloc)) {
```

PR #715, opened 2020-07-17
Use admins hierarchy as address for places and regions where "street" is not defined

amatissart:
## Description
This PR suggests to add a new field `parentAdmins` in the address object, to be used as an additional context for the object, especially for places where the exact address or street name is not defined (e.g cities, districts, regions, etc.).

## Why
In many cases, using the city and country names is not enough to disambiguate the different search results. 
Using the regions where a city is located is more relevant for the user, compared to the current situation where the place name is repeated on the second line.

## Screenshots
Before|After
--|--
image|image

### DIFF
--- a/src/libs/address.js
+++ b/src/libs/address.js
@@ -1,12 +1,15 @@
 import IdunnPoi from '../adapters/poi/idunn_poi';
 
 /**
- * Format an address given an address object (name, city, country, and label)
+ * Format an address given an address object (name, city, country, parentAdmins and label)
  * @param {*} address - an address object
  */
 export function format(address) {
   if (!address) {return '';}
-  const { street, city, country } = address;
+  const { street, city, country, parentAdmins } = address;
+  if (!street && parentAdmins) {
+    return parentAdmins;
+  }
   return [street, city, country]
     .filter(i => i) // Filter out any undefined value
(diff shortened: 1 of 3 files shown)


src/libs/address.js
```diff
@@ -75,3 +80,14 @@ function findAdminIdunn(raw, name) {
     .values(raw.address?.admins || {})
     .find(a => a.class_name === name);
 }
+
+/**
+ * Format administrative hierarchy where a poi (from Idunn or Bragi) is located
+ * @param {*} raw - the raw poi object
+ */
+function extractParentAdmins(raw) {
```
G-Ray, line 88: Instead of returning a preformatted string, can we extract values independently ?
The city and country should already be in an address object, so we should only extract new information, like region/state.

This could be done in the `normalize` function, with helpers like `findAdminBragi`.

G-Ray: This way, the formatting would still occurs at one place : in the `format` function

amatissart: I was a bit reluctant at first to deal with the complexity of all administrative regions in these functions.
But you're right, it's probably more future-proof to format the regions explicitly. 

I updated the PR to add 5 new fields in the address object, in order to cover all types of regions.

PR #717, opened 2020-07-22
QMAPS-1624 multiline addresses in poi panel

xem:
## Description
Display multiline addresses for PoIs, locations, latlons

## Screenshots
POI
image

Address from geocoder
image

Street name from geocoder
image

Area
image

Latlon near an address
image

Pages Jaunes list
image

Pages jaunes item
image

(addresses in suggest stay unchanged)

PS: After doing F5 on a poi panel, the address appears without postal code, then <1s later, it's redrawn with the postal code.
I don't understand where that behavior comes from and why we can't have the postal code from the beginning...help?
wut

PS 2: I don't understand why the tests don't pass, more exactly why `address: document.querySelector('.poiItem-address').innerText` is "" when a PoI (musée d'orsay) is loaded. In my console it's not empty but "1 Rue de la Légion d'Honneur\n75007 Paris\nFrance"

### DIFF
--- a/src/components/PoiTitle.jsx
+++ b/src/components/PoiTitle.jsx
@@ -39,7 +39,7 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
   return <div className="poiTitle">
     <h2 className="poiTitle-main u-text--smallTitle">{title || subclass}</h2>
     {alternative && <div className="poiTitle-alternative u-text--subtitle u-italic">
-      <MultilineAddress address={address}/>
+      {alternative}
     </div>}
     {title && <div className="poiTitle-subclass u-text--subtitle">{subclass}</div>}
   </div>;
--- a/src/components/PoiTitle.jsx
+++ b/src/components/PoiTitle.jsx
@@ -2,6 +2,7 @@
 import React from 'react';
 import poiSubClass from 'src/mapbox/poi_subclass';
 import { capitalizeFirst } from 'src/libs/string';
+import { filter } from 'src/libs/address';
 
 const PoiTitle = ({ poi, withAlternativeName }) => {
   const { name, localName, subClassName, address } = poi;
@@ -11,10 +12,12 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
     const latLon = name;
 
     // Close to (address) + GPS coordinates
-    if (address?.label) {
+    if (address) {
       return <div className="poiTitle">
         <div className="u-text--subtitle u-italic u-mb-4">{ _('Close to', 'poi')}</div>
-        <h2 className="poiTitle-main u-text--smallTitle u-mb-4">{address?.label}</h2>
+        <h2 className="poiTitle-main u-text--smallTitle u-mb-4">{
+          filter(address).map(item => <div key={item}>{item}</div>)
+        }</h2>
         <div className="poiTitle-position">{latLon}</div>
       </div>;
     }
@@ -32,10 +35,11 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
   const alternative = (withAlternativeName && name && localName && localName !== name) && localName;
   const subclass = capitalizeFirst(poiSubClass(subClassName));
 
+  // Location / address
   return <div className="poiTitle">
     <h2 className="poiTitle-main u-text--smallTitle">{title || subclass}</h2>
     {alternative && <div className="poiTitle-alternative u-text--subtitle u-italic">
-      {alternative}
+      { filter(address).map(item => <div key={item}>{item}</div>) }
     </div>}
     {title && <div className="poiTitle-subclass u-text--subtitle">{subclass}</div>}
   </div>;
--- a/src/components/PoiTitle.jsx
+++ b/src/components/PoiTitle.jsx
@@ -16,7 +16,7 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
       return <div className="poiTitle">
         <div className="u-text--subtitle u-italic u-mb-4">{ _('Close to', 'poi')}</div>
         <h2 className="poiTitle-main u-text--smallTitle u-mb-4">{
-          filter(address).map(item => <div key={item}>{item}</div>)
+          filter(address).map((item, index) => <div key={index}>{item}</div>)
         }</h2>
         <div className="poiTitle-position">{latLon}</div>
       </div>;
@@ -39,7 +39,7 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
   return <div className="poiTitle">
     <h2 className="poiTitle-main u-text--smallTitle">{title || subclass}</h2>
     {alternative && <div className="poiTitle-alternative u-text--subtitle u-italic">
-      { filter(address).map(item => <div key={item}>{item}</div>) }
+      { filter(address).map((item, index) => <div key={index}>{item}</div>) }
     </div>}
     {title && <div className="poiTitle-subclass u-text--subtitle">{subclass}</div>}
   </div>;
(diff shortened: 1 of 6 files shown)


src/libs/address.js
```diff
@@ -1,30 +1,41 @@
 import IdunnPoi from '../adapters/poi/idunn_poi';
 
 /**
- * Format an address given an address object (name, city, country, label, and other regions)
+ * Filter an address and return an array with the relevant items
  * @param {*} address - an address object
  */
-export function format(address) {
-  if (!address) {return '';}
+export function filter(address) {
```
G-Ray, line 7: I'm not sure about the name, what about something like "extractAddressFields" or 'extractAddressComponents` ?

src/libs/address.js
```diff
@@ -1,30 +1,41 @@
...
+
+  if (!address) {return [];}
+
+  const city = address.postcode && address.city
```
G-Ray, line 11: Will this be correct accross countries ?

src/libs/address.js
G-Ray: Do not forget to add a new unit test :) It should be straightforward

src/components/PoiTitle.jsx
```diff
@@ -11,10 +12,12 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
     const latLon = name;
 
     // Close to (address) + GPS coordinates
-    if (address?.label) {
+    if (address) {
       return <div className="poiTitle">
         <div className="u-text--subtitle u-italic u-mb-4">{ _('Close to', 'poi')}</div>
-        <h2 className="poiTitle-main u-text--smallTitle u-mb-4">{address?.label}</h2>
+        <h2 className="poiTitle-main u-text--smallTitle u-mb-4">{
+          filter(address).map(item => <div key={item}>{item}</div>)
```
amatissart, line 19: the same here

src/components/PoiTitle.jsx
```diff
@@ -32,10 +35,11 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
   const alternative = (withAlternativeName && name && localName && localName !== name) && localName;
   const subclass = capitalizeFirst(poiSubClass(subClassName));
 
+  // Location / address
   return <div className="poiTitle">
     <h2 className="poiTitle-main u-text--smallTitle">{title || subclass}</h2>
     {alternative && <div className="poiTitle-alternative u-text--subtitle u-italic">
-      {alternative}
+      { filter(address).map(item => <div key={item}>{item}</div>) }
```
amatissart, line 42: the same here. Hum maybe this logic could be moved to its own component :)

src/libs/address.js
amatissart: I agree about the name. `toArray` or maybe `formatMultiline` ?

src/libs/address.js
amatissart: Probably not. But this is probably good enough for now, as we are not sure yet about how this formatted address should be used in the application. In the longer term the API could be responsible for providing the appropriate format depending on the country.

amatissart: And I would name this `cityAndPostcode` to be more explicit

src/libs/address.js
```diff
@@ -1,30 +1,41 @@
...
+    ? address.postcode + ' ' + address.city
+    : address.city;
 
   if (!address.street) {
     return [
       address.suburb,
       address.cityDistrict,
-      address.city,
+      city,
```
amatissart, line 19: I would not use the postcode here, as this format is used to provide a context about where a point or zone is located if no postal address is appropriate.

src/libs/address.js
```diff
@@ -62,12 +73,13 @@ export function normalize(type, raw) {
     };
   }
 
-  if (type === 'idunn') {
+  if (type === 'idunn' || type === 'latlon') {
```
amatissart, line 76: Here `type` describes the API where the `raw` object comes from : Idunn (the main API for all places), or Bragi (the geocoder used for search). So this extra condition should not be necessary.

src/components/PoiTitle.jsx
bbecquet: Actually, as we discussed multiple times, there should be an `<Address>` component responsible for all this.

src/libs/address.js
xem: ok for toArray, as this function is not only used for multiline but also in format(address, ",") !

src/components/PoiTitle.jsx
bbecquet: How I see it: a "dumb" component rendering a `<div>` (neutral markup, can be wrapped in something else if needed) with at least an `address` prop to pass the object.
Other props that will prove quickly useful include an optional `className` and probably props to change the format (`inline` true/false, `withCountry` true/false, etc.). We don't need to include everything in the beginning, just what is useful right now, but at least we will have introduced consistency and have a well-defined place to do everything related to address format.

xem: of course, I'm on it!
I didn't mark this discussion as solved yet, I just wanted to push a first fix this morning to see if the tests behave like I want with the poi.json modification (it does, I know how to fix the tests now :) )

src/libs/address.js
```diff
@@ -18,13 +23,19 @@ export function format(address) {
...
-    .join(', ');
+  return [address.street, cityAndPostcode, address.country]
+    .filter(i => i); // Filter out any undefined value
+}
+/**
+ * Format an address given an address object (name, city, country, label, and other regions)
+ * @param {*} address - an address object
+ * @param string separator - how items are joined
+ */
+export function format(address, separator = ', ') {
```
bbecquet, line 37: I think this function isn't used anymore in the code base.

src/libs/address.js
```diff
@@ -1,11 +1,16 @@
...
+ * Filter an address and return an array with the relevant items
  * @param {*} address - an address object
  */
-export function format(address) {
-  if (!address) {return '';}
+export function toArray(address) {
+
+  if (!address) {return [];}
+
+  const cityAndPostcode = address.postcode && address.city
```
bbecquet, line 11: This concerns only the last case so would be better placed below the next `if`

src/libs/address.js
```diff
@@ -1,11 +1,16 @@
...
+    ? address.postcode + ' ' + address.city
+    : address.city;
 
   if (!address.street) {
     return [
```
bbecquet, line 16: So for such addresses we can display 7 lines if everything is filled and different? This seems a bit much. I wonder if some filtering decision should be added…

src/libs/address.js
xem: actually it is, but renamed (I'm not the author, is that really necessary?)
image

src/libs/address.js
xem: Indeed, let's @amatissart as he wrote this list

xem: so: in practice it never happens, and we can keep it like that for now, until a more sophisticated server-side formatting is introduced

src/libs/address.js
bbecquet: Good catch! Sorry I searched for `format(` calls and forgot the function could be renamed on import :)
But this raises another question: you modified this function behavior in `toArray`, does it still match what we expect in `SuggestItem`?

xem: The edits I made in address.js only introduce a new intermediary toArray function and adds the postcode in some cases.
According to my tests, the suggest items haven't been altered by these edits:
If I search an address, or a city, or a PoI, the addresses inside the suggestion items won't contain postcodes, so nothing changed compared to before imo.

PR #718, opened 2020-07-23
Remove PoiCard and display full poi details

G-Ray:
## Description
Remove the `PoiCard` component and its associated `css` code. Display full poi's details instead. This behavior will change as soon as we introduce a re-sizable panel for a poi, which will be done in another PR.

The "header" of panel has been refactored. This might introduce visual glitches I did not noticed yet.

## Screenshots
Peek 2020-07-29 15-47


src/panel/poi/PoiPanel.jsx
```diff
@@ -268,28 +252,22 @@ export default class PoiPanel extends React.Component {
...
-          <span className="poi_panel__back_text">{backAction.text}</span>
-        </div>
-      }
-      {isFromPagesJaunes(poi) && <img className="poi_panel__pj_logo"
-        src="./statics/images/pagesjaunes.svg"
-        alt="PagesJaunes" />
-      }
-    </div>;
+    const header = backAction &&
+      <div className={backAction.className} onClick={backAction.callback}>
```
bbecquet, line 258: Maybe it's possible to use the `<Flex>` component here to get rid of even more CSS.

G-Ray: yep, it's done, thanks

src/components/ui/Flex.jsx
```diff
@@ -1,19 +1,29 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 
-const Flex = ({ children, inline, className,
+const Flex = forwardRef(({
```
G-Ray, line 5: I used forwardRef and added rest props

src/panel/poi/PoiPanel.jsx
```diff
@@ -157,20 +155,6 @@ export default class PoiPanel extends React.Component {
     window.app.navigateTo('/favs');
   }
 
-  showDetails = () => {
-    const poi = this.getBestPoi();
-    Telemetry.add(Telemetry.POI_SEE_MORE, null, null,
```
bbecquet, line 162: The corresponding constant (`poi_see_more`) in `local_modules/telemetry/index.js` can be removed.

src/panel/poi/PoiPanel.jsx
```diff
@@ -268,28 +253,21 @@ export default class PoiPanel extends React.Component {
         text: _('Back to list'),
         className: 'poi_panel__back_to_list',
```
bbecquet, line 254: Nit: There is only one version left of this, so this could be put inline in the `Flex` element.

PR #719, opened 2020-07-24
Center actions buttons in service panel on mobile

amatissart:
## Screenshots
Before|After
--|--
image|image


src/scss/includes/panels/service_panel.scss
```diff
@@ -66,6 +66,7 @@
   .service_panel__actions {
     margin: 0 auto;
     display: block;
+    text-align: center;
```
G-Ray, line 69: easy :+1:

G-Ray: KISS

PR #720, opened 2020-07-24
Fix favorite toggle for POI selected from a list: fullPoi was not fetched

amatissart:
## Description

The favorite toggle expects that `state.fullPoi` is populated with the data from the API. However, when a user clicked on a POI item from a list, `loadPoi()` was never called and the state was never updated.

Using `loadPoi()` revealed another bug in this context: highlighting the marker related to the selected POI depended on the property `marker_id` being set in the object. This mechanism breaks when this POI object had been de-serialized in the PoiPanel.

### DIFF
--- a/src/adapters/scene_category.js
+++ b/src/adapters/scene_category.js
@@ -5,6 +5,10 @@ import Telemetry from 'src/libs/telemetry';
 import { toUrl } from 'src/libs/pois';
 import { fire, listen } from 'src/libs/customEvents';
 
+function getMarkerId(poi) {
+  return `marker_${poi.id}`;
+}
+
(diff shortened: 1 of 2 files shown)


amatissart:
Bonus question: is using `Poi.deserialize` still necessary, now that `Poi` is basically a data class without any method ?

bbecquet:
> Bonus question: is using `Poi.deserialize` still necessary, now that `Poi` is basically a data class without any method ?

I agree we should be at (or near) the point where we'll be able to remove it, but as there are still two occurences of `instanceof Poi` in the code, it may require some extra steps and checks. Let's address it seperately.

PR #721, opened 2020-07-29
QMAPS-1481 

xem:
## Description

On mobile: 
- blur the current field (to hide the mobile keyboard) when the user is scrolling the suggests list
- close the suggest list when clicking on the "<-" or "x" icon or one of the suggest items.
- apply this on the main search field and the two Directions panel fields

On desktop: change nothing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5125,11 +5125,6 @@
       "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.2.0.tgz",
       "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw=="
     },
-    "are-passive-events-supported": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/are-passive-events-supported/-/are-passive-events-supported-1.1.1.tgz",
-      "integrity": "sha512-5wnvlvB/dTbfrCvJ027Y4L4gW/6Mwoy1uFSavney0YO++GU+0e/flnjiBBwH+1kh7xNCgCOGvmJC3s32joYbww=="
-    },
     "are-we-there-yet": {
       "version": "1.1.5",
       "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -5125,6 +5125,11 @@
       "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.2.0.tgz",
```
G-Ray, line 5125: package-lock shouldn't need any update I think ?

src/components/ui/Suggest.jsx
```diff
@@ -159,3 +157,12 @@ Suggest.propTypes = {
 };
 
 export default Suggest;
+
+// Blur the focused field and close the suggest list when clicking outside of a field
+if (!window.onclick) {
+  window.onclick = e => {
```
G-Ray, line 163: We should register the event with `addEventListeners` in order to allow multiple listeners

src/components/ui/Suggest.jsx
```diff
@@ -159,3 +157,12 @@ Suggest.propTypes = {
...
+    if (e.target.tagName !== 'INPUT') {
+      fire('close_suggest');
```
G-Ray, line 165: we might need to use the locally defined `close` function which will clear items ? I'm not sure

xem: yes, the event listener calls close()

src/components/ui/Suggest.jsx
xem: well, I tried that at first, but the opposite happened: multiple event listeners were added, while this "onclick" function is always unique.

G-Ray: yes, because the event listener should be registered once in a `useEffect`, (do not forget te cleanup function)

src/components/ui/Suggest.jsx
G-Ray: do we really need to register a global `close_suggest` event here ? I think all the logic can be kept inside `Suggest.jsx`

src/components/ui/Suggest.jsx
```diff
@@ -159,3 +162,7 @@ Suggest.propTypes = {
 };
 
 export default Suggest;
+
+// Blur the focused field and close the suggest list when clicking outside of a field
+window.onclick = null;
```
G-Ray, line 167: is this needed ?

src/components/ui/SuggestsDropdown.jsx
```diff
@@ -3,6 +3,13 @@ import classnames from 'classnames';
 import { object, func, string, arrayOf, bool } from 'prop-types';
 
 import SuggestItem from './SuggestItem';
+import { isMobileDevice } from '../../libs/device';
+
+const blurField = () => {
+  if (isMobileDevice()) {
```
G-Ray, line 9: Can you use the `DeviceContext` instead ?

src/components/ui/Suggest.jsx
xem: ah, no, not anymore

src/components/ui/SuggestsDropdown.jsx
xem: yes, but shouldn't we remove isMobileDevice everywhere?

PR #724, opened 2020-07-30
Update poi panel design

G-Ray:
## Description
Update Poi panel design according to mockups

## Screenshots
mobile
desktop


src/components/ui/Block.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const Block = ({ icon, title, children, className }) =>
```
bbecquet, line 4: Not a bit deal, but I'm not sure this component should be stored in this folder, as it's only for POI block for now. Having a name as generic as "Block" can be misleading.

src/panel/poi/blocks/Covid19.jsx
```diff
@@ -2,6 +2,7 @@
 import React, { Fragment } from 'react';
 import nconf from '@qwant/nconf-getter';
 import TimeTable from './TimeTable';
+
```
bbecquet, line 5: Why all these line jumps in imports? :cry:

src/panel/poi/blocks/Images.jsx
```diff
@@ -12,10 +12,9 @@ const ImagesBlock = ({
   }
 
   return <div className="poi_panel__pictures">
-    <h5 className="poi_panel__pictures_title">
-      <i className="icon-icon_camera"></i>
+    <div className="u-text--smallTitle" style={{ marginBottom: 20 }}>
```
bbecquet, line 15: Time for a `.u-mb-20` it seems :)

src/panel/poi/blocks/Images.jsx
```diff
@@ -12,10 +12,9 @@ const ImagesBlock = ({
   }
 
   return <div className="poi_panel__pictures">
-    <h5 className="poi_panel__pictures_title">
```
bbecquet, line 15: `h5` was clearly not the good choice anymore, but it would be nice to keep some kind of semantic hierarchy. As the POI title is `h2`, logically this should be a `h3`. What do you think? Or we decide we address all this globally on the app later?

src/panel/poi/blocks/Wiki.jsx
```diff
@@ -10,14 +11,16 @@ const WikiBlock = ({
         { block.description }
       </p> }
     <br/>
-    { block.url &&
+    { block.url && <Flex inline justifyContent="center" style={{ width: '100%' }}>
```
bbecquet, line 14: If you remove the `inline` prop, you don't need the explicit style.

src/scss/includes/components/divider.scss
```diff
@@ -0,0 +1,6 @@
+.divider {
+  &-line {
+    height: 1px;
+    background-color: $grey-light;
+  }
```
bbecquet, line 5: Just put the `width: 100%` here (see previous comment).

src/scss/includes/panels/poi_panel.scss
```diff
@@ -197,109 +161,46 @@ $BLOCK_ICON_FONT_SIZE: 16px;
...
   color: $secondary_text;
+  margin-left: auto;
 }
 
 .poi_panel__block__collapse--reversed {
   transform: rotate(180deg);
 }
 
-
 .poi_panel__sub__sub_block__title {
```
bbecquet, line 183: We should get rid of that class and use one of the standard fonts.

src/components/ReviewScore.jsx
```diff
@@ -30,7 +30,7 @@ const ReviewScore = ({ poi, reviews: { global_grade, total_grades_count, url },
       )}
     </span>
     <span className="u-text--caption reviewScore-count">
-      {total_grades_count} {_n('review', 'reviews', total_grades_count)}
+      ({total_grades_count}) {_('With Page Jaunes', 'reviews')}
```
amatissart, line 33: typo
```suggestion
      ({total_grades_count}) {_('With PagesJaunes', 'reviews')}
```

bbecquet, line 33: Also should be "on PagesJaunes", not "with".

src/panel/poi/blocks/Images.jsx
G-Ray: I think it's better to define our typography once and for all across the app :)

src/components/ui/Block.jsx
G-Ray: I moved it to `src/poi/blocks/Block.jsx`

src/components/ReviewScore.jsx
```diff
@@ -30,7 +30,7 @@ const ReviewScore = ({ poi, reviews: { global_grade, total_grades_count, url },
       )}
     </span>
     <span className="u-text--caption reviewScore-count">
-      {total_grades_count} {_n('review', 'reviews', total_grades_count)}
+      ({total_grades_count}) {_('on PageJaunes', 'reviews')}
```
amatissart, line 33: still a typo :)
```suggestion
      ({total_grades_count}) {_('on PagesJaunes', 'reviews')}
```

src/panel/poi/blocks/Hour.jsx
```diff
@@ -16,16 +18,11 @@ export default class HourBlock extends React.Component {
...
-        <div className="icon-icon_clock poi_panel__block__symbol"></div>
-        <div className="poi_panel__block__content">
-          <TimeTable
-            schedule={schedule}
-            title={this.props.covid19enabled && _('See the usual opening hours', 'covid19')}
-          />
-        </div>
-      </div>
-    </div>;
+    return <Block icon="icon_clock" title={_('Opening Hours')}>
```
bbecquet, line 21: No need for a capital on "Opening **H**ours", and there'ss already an i18n string `'Opening hours'`.

src/panel/poi/PoiBlockContainer.jsx
```diff
@@ -29,18 +33,38 @@ export default class PoiBlockContainer extends React.Component {
...
       {phoneBlock && <PhoneBlock block={phoneBlock} />}
+      {hourBlock && <HourBlock block={hourBlock} covid19enabled={!!displayCovidInfo} />}
       {recyclingBlock && <RecyclingBlock block={recyclingBlock} />}
-      {websiteBlock && <WebsiteBlock block={websiteBlock} poi={this.props.poi} />}
       {contactBlock && <ContactBlock block={contactBlock} />}
-      {imagesBlock && <ImagesBlock block={imagesBlock} poi={this.props.poi} />}
+      {this.props.poi.address.label &&
+        <Block
+          icon="map-pin"
+          title={_('Address')}
```
bbecquet, line 57: You should use the `'address'` string, without the first cap, and capitalize using `u-firstCap` or `capitalizeFirst`.

G-Ray: Ok, I switched all block titles to lowercase

PR #725, opened 2020-07-31
[POI panel] Dynamic height based on content

bbecquet:
## Description
Alternative to https://github.com/QwantResearch/erdapfel/pull/722, based on DOM changes instead of explicit JS size computation.

### Principle
If provided with a function as `children` prop, the `<Panel>` component can transmit its size (`default`, `minimized` or `maximized`) so the calling component can decide what to render depending on the panel size. This allows to hide the content that doesn't need to be displayed in `default` or `minimized` sizes.
The POI panel is also given `auto` height instead of fixed height. That way, the height adapts automatically to the content, even if it changes asynchronously (ressource loading, etc.).

### TODO

 - [x] **Fix a bug**: https://github.com/QwantResearch/erdapfel/pull/718 apparently introduced a bug on panel resize (all of them, not just POI).
When coming from 'minized' state, a mouse/touch drag to the left or right starts a resize action but stops right there, leaving the panel in an intermediate state. This conflicts with the current PR behavior, maybe hiding other problems with it, so this should be fixed first.
**=> simple fix: https://github.com/QwantResearch/erdapfel/pull/741**

 - [ ] **Display full content early**: currently, when going from `default` size to `maximized`, the panel displays extra POI info only when the panel is released, leaving a misleading white space during the resize. We could compute the target size of the Panel (the one "towards" which it's currently being resized) during the resize and give it to the children so we could fill the panel with content early.
**=> changing the panel content dynamically causes the bug mentioned above. Let's deal with it separately.**

 - [ ] **Restore CSS transitions**: using `height: auto;` for `default` and `minized` sizes makes CSS height transitions from and to them impossible, as they need numerical values. We can keep the computed size as start value, but there may be no easy way for target value.
**=> there is no major feature break with the loss of these transitions. Let's deal with it separately**

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -252,7 +252,7 @@ export default class Panel extends React.Component {
               ref={this.panelContentRef}
               {...(isMobile && resizeHandlers)}
             >
-              {typeof children === 'function' ? children(size) : children}
+              {typeof children === 'function' ? children(size, isMobile) : children}
             </div>
           </div>
         }
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -38,7 +38,7 @@ function getTargetSize(previousSize, moveDuration, startHeight, endHeight, maxSi
 
 export default class Panel extends React.Component {
   static propTypes = {
-    children: PropTypes.node.isRequired,
+    children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
     title: PropTypes.node,
     minimizedTitle: PropTypes.node,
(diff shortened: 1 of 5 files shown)


src/components/ui/Panel.jsx
```diff
@@ -111,6 +111,10 @@ export default class Panel extends React.Component {
   }
 
   holdResizer = (event, forceResize = false) => {
+    if (this.state.size === 'default') {
+      this._lastDefaultHeight = this.panelDOMElement.offsetHeight;
+    }
```
amatissart, line 116: This `offsetHeight` should not be saved if the event has been triggered during the transition to the `default` state (before its final height is reached).

PR #731, opened 2020-08-05
Qmaps 1440 masq footer

G-Ray:
## Description
- Fix missing `isMobile` check in `Panel.jsx`
- Remove FavoriteMasqFooter from favorites, and related assets.

## Screenshots
image


src/components/ui/Panel.jsx
```diff
@@ -237,7 +237,7 @@ export default class Panel extends React.Component {
               onClick={() => this.handleHeaderClick()}
               {...(isMobile && forceResizeHandlers)}
             >
-              {resizable && size === 'minimized' && minimizedTitle
+              {resizable && isMobile && size === 'minimized' && minimizedTitle
```
amatissart, line 240: Maybe this `isMobile` check should be defined in the `onClick` action ?  
It does not make sense to update the `size` state on desktop, right ?

G-Ray: :+1:

PR #733, opened 2020-08-05
Remove MasqFavoriteModal

G-Ray:
## Description
- Remove MasqFavoriteModal

## Why
Masq is being retired


src/panel/poi/PoiPanel.jsx
```diff
@@ -222,12 +219,6 @@ export default class PoiPanel extends React.Component {
     if (this.state.isPoiInFavorite) {
       store.del(poi);
     } else {
-      if (this.isMasqEnabled) {
-        const isLoggedIn = await store.isLoggedIn();
-        if (!isLoggedIn) {
-          await openMasqFavModalAndWaitForClose();
```
bbecquet, line 228: You can remove the `async` on the function as it doesn't contain any `await` now.

What do you think about adding this lint rule https://eslint.org/docs/rules/require-await to detect these cases? There are apparently a lot of them in the project, probably the results of refacto and sometimes misunderstandings.

G-Ray: :+1: Good idea about adding this rule to the linter

PR #734, opened 2020-08-05
Refactor telemetry without class

G-Ray:
## Description
Remove Telemetry class in favor of functions

## Why
simplicity

### DIFF
--- a/src/libs/telemetry.js
+++ b/src/libs/telemetry.js
@@ -5,68 +5,80 @@ import Error from '../adapters/error';
 
 const telemetry = nconf.get().telemetry;
 const system = nconf.get().system;
-const telemtryEventUrl = 'events';
+const telemetryEventUrl = 'events';
 const uniqEventList = [];
+const events = {};
 
-export default class Telemetry {
-  constructor() {}
+/*
+  This converts "/src/libs/telemetry.js" events into a map where you can use an event as follow:
+  'app_start' event will be accessible like this "Telemetry.APP_START" and its value will be the
+  original (so 'app_start').
+*/
+telemetryModule.events.forEach(event => {
+  events[event.toUpperCase()] = event;
+});
 
-  static add(event, type, source, data) {
-    if (event) {
-      if (type && source) {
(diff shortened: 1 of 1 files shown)


src/libs/telemetry.js
```diff
@@ -5,68 +5,81 @@ import Error from '../adapters/error';
...
-    const data = { type: event };
-    if (typeof extra_data === 'object' && extra_data !== null) {
-      Object.keys(extra_data).forEach(key => {
-        data[key] = extra_data[key];
-      });
-    }
-    const telemetryUrl = `${system.baseUrl}${telemtryEventUrl}`;
-    return Ajax.post(telemetryUrl, data);
+function send(event, extra_data) {
+  const telemetryUrl = `${system.baseUrl}${telemetryEventUrl}`;
```
bbecquet, line 42: It made more sense to have the early return conditions before declaring stuff that may not be used.

src/libs/telemetry.js
```diff
@@ -5,68 +5,81 @@ import Error from '../adapters/error';
...
-// This converts "/src/libs/telemetry.js" events into a map where you can use an event as follow:
-//
-// 'app_start' event will be accessible like this "Telemetry.APP_START" and its value will be the
-// original (so 'app_start').
-telemetryModule.events.forEach(event => {
-  Telemetry[event.toUpperCase()] = event;
-});
+function buildInteractionData({ source, template, id, zone, element, category }) {
+  const data = {
+    'event': 'click',
```
bbecquet, line 64: Nit: You can remove the unnecessary quotes for object keys.

src/libs/telemetry.js
```diff
@@ -5,68 +5,81 @@ import Error from '../adapters/error';
...
+
+  return {
+    'front_search_user_interaction_data': data,
+  };
+}
+
+export default {
+  add,
+  addOnce,
+  send,
```
bbecquet, line 82: Looks like `send` isn't publicly needed so it may be safer not to export it.

bbecquet (approved):

PR #736, opened 2020-08-06
[ESLint] No 'async' without 'await'

bbecquet:
## Description
Following discussion [here](https://github.com/QwantResearch/erdapfel/pull/733#discussion_r466225601), add an ESLint rule to prevent declaring functions as `async` when it's not needed i.e. they don't include the `await` instruction.
Apply the rule to clean the current codebase.

## Why
Useless and can lead to false assumptions when reading the code.

### DIFF
--- a/src/libs/telemetry.js
+++ b/src/libs/telemetry.js
@@ -29,7 +29,7 @@ export default class Telemetry {
     }
   }
 
-  static async send(event, extra_data) {
+  static send(event, extra_data) {
     if (!telemetry.enabled) {
       return;
     } else if (typeof event === 'undefined') {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -95,7 +95,8 @@
         "object-curly-spacing": ["error", "always"],
         "object-shorthand": "error",
         "react/prop-types": 0,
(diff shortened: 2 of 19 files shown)


src/libs/telemetry.js
```diff
@@ -29,7 +29,7 @@ export default class Telemetry {
     }
   }
 
-  static async send(event, extra_data) {
+  static send(event, extra_data) {
```
amatissart, line 32: I wonder if `async` may be useful for readability in such cases.  
Here, it makes explicit that `send()` returns a Promise (returned by `Ajax.post`)

bbecquet: Actually I wonder too…
I don't find this system very reliable as nothing ensures this is systematic. You would have to know the full call chain of every function to properly tag it `async` or not.

But there is also the fact that `async` ensures that the return value **is** a Promise, warping any non-promise value in a new promise if needed. So, technically, this syntax-only PR may break something if some parts of the code rely on being always able to do a `.then()` on the return value of this function. I'm pretty sure there is no such code, but I wonder if this was used intentionnaly for now.

bbecquet:
This isn't that useful. Let's close this for now, it's easy to bring back anyways.

PR #738, opened 2020-08-10
Fix touchmove on minimized panel: 'document' can't receive events after target has been removed

amatissart:
## Description
This PR suggests to attach the move handler to the target where the touchstart/mousestart event has been trigger (instead of `document`)
This is necessary keep receiving the events, even if the target is removed from the DOM.

## Why
The Panel component renders a `<span className="minimizedTitle">` in the minimized state only.
When this <span> is the target of touchmove/mousemove events, it is then removed (`size` is set to 'default') and the events
can't bubble up to the document, where touchmove/mousemove handlers are attached.

This should fix a bug where a mouse/touch drag from the minimized state could start a resize action and leave the panel in an intermediate state before the move ends.

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -57,6 +57,7 @@ export default class Panel extends React.Component {
   constructor(props) {
     super(props);
     this.moveHandler = null;
+    this.moveTarget = null;
     this.panelContentRef = React.createRef();
     this.state = {
       holding: false,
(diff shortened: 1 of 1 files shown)


bbecquet:
Sadly this fix doesn't seem to work :(
In my PR the problem still exists. It can also be observed on the 'details' part of the maximized view : if you start your slide from this area to reduce the panel, the DOM element disappears during the drag and the event is "lost". At least the root cause of this bug is well-known now :)

src/components/ui/Panel.jsx
```diff
@@ -119,10 +121,13 @@ export default class Panel extends React.Component {
     this.removeListeners();
 
     this.moveHandler = event => this.move(event, forceResize);
+    this.moveTarget = event.target;
+    // It is important to attach the move handler to the current target (instead of `document`)
+    // in order to keep receiving the events, even if the target is removed from the DOM.
```
bbecquet, line 126: I don't think this solves anything. In both cases the target element is the same, and if it suddenly disappears from the DOM it stops bubbling `touch/mousemove` events. The element where the listener is attached doesn't matter here.

amatissart: By attaching the `touchmove` event handler to the target of the `touchstart` event, events no longer need to bubble up as the listener function is called directly. At least, that's what I had observed on both Chrome and Firefox on Android, and I could't reproduce the bug after this change.

But I suppose I may have missed something about the conditions when the bug appears :shrug:

PR #742, opened 2020-08-18
[POI panel] Redesign mobile minimized and default content

bbecquet:
## Description
New redesign of the POI panel content, for the three different panel sizes on mobile, according to the latest designs.
Except for:
- the title font size => *too big, needs to big discussed*
- the position of reviews, between the title and the POI class => *needs more component refacto, as the title and class are managed as a single component for now*

*Note: as a side effect of these changes, POI items in result lists change too. But this brings them closer to the final design of this part*

## Screenshots
|Before|After|
|---|---|
|www qwant com_maps_places__type=restaurant(iPhone 6_7_8 Plus)|localhost_3000_place_pj_56927563@Les_Bios_Arts(iPhone 6_7_8 Plus) (2)|
|www qwant com_maps_place_pj_50015630@Le_Zinc_Du_March%C3%A9(iPhone 6_7_8 Plus)|localhost_3000_(iPhone 6_7_8 Plus)|
|www qwant com_maps_place_pj_50015630@Le_Zinc_Du_March%C3%A9(iPhone 6_7_8 Plus) (1)|localhost_3000_(iPhone 6_7_8 Plus) (1)|

### DIFF
--- a/src/components/PoiItem.jsx
+++ b/src/components/PoiItem.jsx
@@ -5,7 +5,6 @@ import OsmSchedule from 'src/adapters/osm_schedule';
 import ReviewScore from 'src/components/ReviewScore';
 import PoiTitleImage from 'src/panel/poi/PoiTitleImage';
 import classnames from 'classnames';
-import MultilineAddress from 'src/components/ui/Address';
 
 const PoiItem = ({ poi,
   withOpeningHours,
@@ -15,15 +14,6 @@ const PoiItem = ({ poi,
   ...rest
 }) => {
   const reviews = poi.blocksByType?.grades;
-  const address = poi.address || {};
-
-  const Address = () =>
-    poi.subClassName !== 'latlon'
-      ? <div className="u-text--subtitle poiItem-address">
-        <MultilineAddress address={address} omitCountry={poi.type !== 'admin'} />
-      </div>
-      : null
-  ;
 
   const Reviews = () =>
     reviews
@@ -45,7 +35,6 @@ const PoiItem = ({ poi,
   return <div className={classnames('poiItem', className)} {...rest}>
     <div className="poiItem-left">
       <PoiTitle poi={poi} withAlternativeName={withAlternativeName} />
-      <Address />
       <Reviews />
       <OpenStatus />
     </div>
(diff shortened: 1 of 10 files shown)


xem (commented):
Hello,
The code seems very nice.

Visually, and in terms of UX, I have some remarks that are not blocking for this review but maybe to discuss with design or confirm that it is already planned to address them later:

1) Non-PJ pois don't have the "return to list" link on top, and it leaves a big blank space that may not be ideal (maybe put the title higher in that case, aside to the "X" button?)

image

2) On Firefox and Chrome in Mobile simulator mode, it seems impossible to fold the panel completely, by dragging the handle at the bottom of the screen. Is it broken, or only broken on desktop with mobile simulator?

3) In french UI, the PJ icon at the end of "Retour à la liste" link may be a bit too close to the resize handle, and (in all languages) the text doesn't seem properly aligned vertically to the arrow
image

4) Could we maybe prevent the user to from dragging the handle upwards if there's nothing else to see inside the panel? Doing so only reveals more blank space and may be a bit frustrating.
image

xem (approved):

bbecquet:
@xem 

Thanks for the review :)
Here are some answers about your questions in 

* Points 1 and 3
  I agree the header part of the panel isn't ideal for now, either when empty or when display the back action. I left it out of this PR on purpose, to be addressed separately, as there are still some design questions with when this is displayed and such.

* Point 2
  There has always been problems with the desktop browsers simulator with these complex touch interactions. Especially, they often don't detect the end of the swipe, so the panel is dragged but hard to release or doesn't change position. But I can make it work sometimes, and more importantly I confirm this works on real mobile browsers.

* Point 4
  There is no case where the fully deployed panel displays nothing more. Even if the POI has no extra information, there is still the OSM contribution block or for simple positions the category block. If you encountered an empty space when the panel is maximized (and not still dragging), it probably means the Idunn API fetch failed, which happens quite a lot these days on dev :/
  
  But if you mean we should have content *during the resizing* between default and maximized, I totally agree :) The current behavior is a limitation of the dynamic height implementation. As explained in https://github.com/QwantResearch/erdapfel/pull/725, we decided to address this resize problem separately. And there's already a task in the bug tracker :)

PR #743, opened 2020-08-18
Progressive focus: send zoom level with autocomplete queries

remi-dupre:
## Description

Pass the zoom option to Idunn in order to allow a refinement of the focus mode. Currently, only the coordinates are passed when the zoom level is at least 11 which only allows for focus with a rather big scale.

This is the integration of changes in Idunn from https://github.com/QwantResearch/idunn/pull/162

---

**Update** by @amatissart:
This PR has been updated with a new setting `focusMinZoom` (default value: 11).  
The default behavior is unchanged, but it's now possible to allow focused search for lower zooms in the deploy configuration.

### DIFF
--- a/src/libs/suggest.js
+++ b/src/libs/suggest.js
@@ -12,8 +12,6 @@ import { suggestResults } from 'src/adapters/suggest_sources';
 
 const geocoderConfig = nconf.get().services.geocoder;
 const SUGGEST_MAX_ITEMS = geocoderConfig.maxItems;
-const SUGGEST_USE_FOCUS = geocoderConfig.useFocus;
-const SUGGEST_FOCUS_MIN_ZOOM = 11;
 
 export const selectItem = (selectedItem, { query, replaceUrl = false, fromQueryParams } = {}) => {
   if (selectedItem instanceof Poi) {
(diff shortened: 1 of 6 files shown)


src/libs/suggest.js
```diff
@@ -63,11 +61,10 @@ export const getInputValue = item => {
 
 export const fetchSuggests = (query, options = {}) =>
   suggestResults(query, {
-    withCategories: options.withCategories || false,
-    useFocus: options.useFocus || SUGGEST_USE_FOCUS,
-    focusMinZoom: options.focusMinZoom || SUGGEST_FOCUS_MIN_ZOOM,
-    maxItems: options.maxItems || SUGGEST_MAX_ITEMS,
-    maxFavorites: options.maxFavorites || !query ? 5 : 2,
+    withCategories: options.withCategories ?? false,
```
bbecquet, line 64: Is this change just a question of code style preference?

bbecquet (approved):

src/libs/suggest.js
amatissart: Not exactly, `optional_value ?? default_value` should be preferred to `optional_value || defaut_value` here. It doesn't make a difference when the default is `false`. But for other params below, the default value is `true` and the optional value was always ignored:  
`false ?? true` evaluates to `false`  
whereas
`false || true` evaluates to `true`

bbecquet: Ah yes of course! Sorry for the dumb question :D
I always try to give default false to boolean options for this kind of reason, but sometimes true is cool too.

PR #744, opened 2020-08-19
Ignore .idea config folder

xem:
## Description
Hide the uncommited .idea/* config files created by certain IDEs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ stats.json
 
 .env
 .env.*
+/.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ tmp/
 *.log
 .vscode
 .DS_Store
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@ stats.json
 
 .env
 .env.*
+/.idea/
```
bbecquet, line 25: I think `.idea` is a folder from PhpStorm/WebStorm, not VS Code (but I'm not sure which one you use)

If we add it, let's regroup it with similar dev env specificities, like `.vscode` and `.DS_Store` at the top of file and use the same form (simply `.idea`).

xem: OK!
I've used both IDEs so I may have mixed up which one produces which folder.

bbecquet (approved):

PR #746, opened 2020-08-19
QMAPS-1648 update the X icon's style

xem:
## Description

The X icon that closes a panel now has the following style:

- #0c0c0e on mobile
- #59595f on desktop / #0c0c0e on hover / no background circle on hover

Notes for @bbecquet / C.Ziegler:
- The X icon that empties a search field wasn't modified (to discuss: is it ok like that?)
- The X icon that closes the burger menu wasn't modified (to discuss: it is white. Will it stay white after Masq is removed?)
- I didn't make a separate "x-icon" component (yet), I don't think it belongs to this ticket which only asked for a little redesign.
If it's still necessary to create a component, may we do it in another ticket?

image
image

### DIFF
--- a/src/scss/includes/components/panel.scss
+++ b/src/scss/includes/components/panel.scss
@@ -33,7 +33,6 @@
     margin-left: auto;
 
     &:hover {
-      background: $background_hover;
       color: $primary_text;
     }
   }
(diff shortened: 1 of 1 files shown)


bbecquet (approved):
> If it's still necessary to create a component, may we do it in another ticket?

Nope, you're right :) I was impatient to create a new micro-component for that but there is really 0 added benefit for now.

PR #749, opened 2020-08-21
Change POI panel size dynamically with user interactions

bbecquet:
## Description
Adjust the POI panel size automatically depending on user interaction:
 - if the user drags/zoom the map while a POI panel is open on its default size, the panel gets automatically minimized
 - it the user clicks on a POI while the panel is minimized, if will be restored to default.

Both actions require external elements (the map and the router) to change the panel size imperatively. 
This level of control would require hacks to be done properly with the current implementation of the size only managed as internal state by the `<Panel>` component. So I did what is commonly done in such case in React: I lifted the state up.
The panel size is now managed as a top-level app state, in the `<PanelManager>` component, and can be changed by it in response to events or life-cycles. The size is passed as prop to the `<Panel>` component, with the help of a React context.  think this change will get even more useful later `<Panel>` still manages the complex drag and drop interactions, but they are just one way among others to change the size.

## Why
Improve mobile screen use to react to user interactions.

## Screenshots
Peek 21-08-2020 17-16

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -149,6 +149,9 @@ Scene.prototype.initMapBox = function() {
       });
     }
 
+    this.mb.on('dragstart', () => { fire('map_user_interaction'); });
+    this.mb.on('pitchstart', () => { fire('map_user_interaction'); });
+
     this.mb.on('moveend', () => {
       const { lng, lat } = this.mb.getCenter();
       const zoom = this.mb.getZoom();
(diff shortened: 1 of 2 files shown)


xem (approved):
Code and local test are OK, however is it normal that the size change is made with no transition?
(I think it would be more comfortable visually, but it's not blocking, so I approve the PR)

bbecquet:
@xem 

> however is it normal that the size change is made with no transition?
>

Still the same problem mentioned in https://github.com/QwantResearch/erdapfel/pull/725. Using `auto` as height means CSS transitions on this property can't work (transitions still work for other panels which use `50%` or `50px`).
We agreed to try to find a way around this separately, as it's less problematic than the problems already addressed but not trivial at all, so it would have made everything far more difficult to solve at the same time.

PR #751, opened 2020-08-24
Manage search input content with history state

bbecquet:
## Description
- Make the category label persistent in the main search input, when going from a POI list to a single POI of the list.
- Remove it when the user exits exit the category (ex: closing it explicitely, clicking elsewhere on the map, opening another feature, etc.)

To achieve that, instead of managing the content at the individual component level (`PoiPanel` and `CategoryPanel`), we do it at the same level as the router.
The logic is that content is dependant on the app navigation state (`poiFilters`, pushed as part of the history state of the POI list and a single POI). So if we update it at each route change, it will always be up-to-date with the application state, including when navigating back/forward.

This also makes the code cleaner, regrouping imperative manipulations of this search input in a single component.

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -136,7 +136,7 @@ const Suggest = ({
         inputNode.value = getInputValue(item);
         close();
         if (onSelect) {
-          onSelect(item);
+          onSelect(item, { query: inputNode.value });
         }
       }}
       onClear={onClear}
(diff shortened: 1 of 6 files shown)


bbecquet:
Putting this PR back in draft as it currently breaks the current feature of keeping the POI query search in the input. Will have to change some stuff in how the suggest works.

xem (approved):
Great

PR #755, opened 2020-08-26
QMAPS-1649 search field redesign

xem:
## Description
Update the display and behavior of the main search field on mobile and desktop.
For each device, there are 4 views with different colors/icons/sizes:

Unfocused empty
image
image

Focused empty
image
image

Unfocused filled

image
image

Focused filled
image

image

Besides this redesign:
- the HTML code has changed (reordering of elements)
- the JS code was enhanced to represent the focused and filled states with css classes on the top bar.
- the tests have been fixed.
- the return arrow on mobile now clears the field

Other detail asked by design team but to do in a different ticket:
- Another style for unfocused filled while we're scrolling the suggests on mobile, or while a list item has been selected (in this case the arrow must disappear)
- Gradient on the right of the field when there's too much text: 
image

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -83,13 +83,6 @@ const Suggest = ({
       fetchItems(value);
       setIsOpen(true);
       setLastQuery(value);
-      const topBarHandle = document.querySelector('.top_bar');
-      if (value.length > 0) {
-        topBarHandle.classList.add('top_bar--search_filled');
-      }
-      else {
-        topBarHandle.classList.remove('top_bar--search_filled');
-      }
     };
 
     const handleKeyDown = async event => {
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -83,6 +83,13 @@ const Suggest = ({
       fetchItems(value);
       setIsOpen(true);
       setLastQuery(value);
+      const topBarHandle = document.querySelector('.top_bar');
+      if (value.length > 0) {
+        topBarHandle.classList.add('top_bar--search_filled');
+      }
+      else {
+        topBarHandle.classList.remove('top_bar--search_filled');
+      }
     };
 
     const handleKeyDown = async event => {
--- a/src/panel/PanelManager.jsx
+++ b/src/panel/PanelManager.jsx
@@ -160,13 +160,22 @@ export default class PanelManager extends React.Component {
     const searchInput = document.querySelector('#search');
     const topBarHandle = document.querySelector('.top_bar');
 
-    searchInput.onfocus = () => {
+    searchInput.addEventListener('focus', () => {
       topBarHandle.classList.add('top_bar--search_focus');
-    };
+    });
 
-    searchInput.onblur = () => {
+    searchInput.addEventListener('blur', () => {
       topBarHandle.classList.remove('top_bar--search_focus');
-    };
+    });
+
+    searchInput.addEventListener('input', () => {
+      const value = searchInput.value;
+      if (value.length > 0) {
+        topBarHandle.classList.add('top_bar--search_filled');
+      } else {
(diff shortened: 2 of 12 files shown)


src/components/ui/Suggest.jsx
```diff
@@ -83,6 +83,13 @@ const Suggest = ({
       fetchItems(value);
       setIsOpen(true);
       setLastQuery(value);
+      const topBarHandle = document.querySelector('.top_bar');
```
bbecquet, line 86: This `<Suggest>` component is also used by the direction feature, where this effect shouldn't happen.
And in React you should avoid direct DOM manipulations like this, to prevent side effects breaking the rendering cycle.

Both these points show this should live elsewhere. As it looks like the things we already do in the `initTopBar` method of `PanelManager`, this is probably where this should go.

src/scss/includes/menu.scss
```diff
@@ -2,7 +2,9 @@
 #react_menu__container {
   align-self: stretch;
```
bbecquet, line 3: If you turn this element to `position: absolute`, this rule related to `flex` shouldn't have any effect anymore.

src/ui_components/search_input.js
```diff
@@ -34,6 +34,8 @@ export default class SearchInput {
       const isMobile = isMobileDevice();
       const isActive = document.activeElement.id === inputElement.id;
       inputElement.value = '';
+      const topBarHandle = document.querySelector('.top_bar');
+      topBarHandle.classList.remove('top_bar--search_filled');
```
bbecquet, line 38: Why is this needed? In any cas, looks like it's linked to my other comment and should be put in `initTopBar()` too.

xem: The "_filled" class is needed because sole special style will apply only when we have started writing in the field.

The DOM manipulation will be removed according to your comment

src/panel/PanelManager.jsx
```diff
@@ -131,13 +131,22 @@ export default class PanelManager extends React.Component {
...
-    };
+    });
 
-    searchInput.onblur = () => {
+    searchInput.addEventListener('blur', () => {
       topBarHandle.classList.remove('top_bar--search_focus');
-    };
+    });
+
+    searchInput.addEventListener('input', () => {
```
bbecquet, line 142: :+1:

bbecquet:
@xem 
On Chrome and WebKit browsers, mobile and desktop, the magnifying glass icon isn't vertically centered with the content
Capture d’écran de 2020-09-01 17-20-59
Capture d’écran de 2020-09-01 17-21-44

bbecquet:
@xem follow-up to the previous comment, the alignment of the burger menu changed too (and there is now less space on the left than on the right of it).

|Master|This branch|
|---|---|
|Capture d’écran de 2020-09-01 17-39-01|Capture d’écran de 2020-09-01 17-39-05|

Favoring `display: flex` for this kind of layout makes alignment issues easier.

xem:
I fixed the icons alignment.
Didn't need flex for such a small scope (if you don't mind).

bbecquet (approved):
I'm really not a fan of the overall added complexity and animation is lagging on low-end devices, but this is  consistent with what the design wants, so :+1: and gg with not getting lost in all these states.

We will probably get back to it and simplify stuff, that's the normal lifecycle.
I take note of these improvements for later:
 - add the opacity transition on logo and burger during the expansion animation
 - try to make this animation smoother
 - use the same SVG for the 3 states of the magnifying glass, instead of 3 different files
 - **bigger but more useful**: migrate the search bar to a React component, so it's self-contained and we can get rid of all the code to manipulate it from elsewhere

xem:
@bbecquet 
- for the magnifying glass icons (and some others) I opened this ticket: QMAPS-1719 
- other thing TODO later: custom state on mobile when the suggests are being scrolled (it's like the unfocused filled state, but without the return arrow)

PR #757, opened 2020-08-28
[Fix] Don't minimize panel when using back link

bbecquet:
## Description
Stop event bubbling on backToList/Favs links of the POI detailed view, so they don't trigger the panel header (which resizes the panel)


bbecquet:
Haha, I'm losing my head, I already opened it here https://github.com/QwantResearch/erdapfel/pull/754

PR #758, opened 2020-08-28
[Panel] Simplify touch event handling and fix on iOS

bbecquet:
## Description
This PR simplifies parts of the `<Panel>` component related to user gestures on mobile.

The root reason was several bugs, especially a big one on iOS which prevented most user interaction on the POI Panel. While debugging, it appears that Safari event manager seemed sometimes confused by React re-renders happening between the `touchstart` and `touchend` event. This lead to the [normal event firing chain](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent#Event_order) being interrupted, so `click` event at the end was never called, and interactive elements inside the panel didn't respond.

The simplification consists mainly of two parts:

**1. avoid unnecessary state/props update**
- don't update the height state on `touchstart`, just init some vars and attach listeners. Updating the state on `touchmove` if the user starts a dragging move is enough.
- don't update the height state/props on `touchend` if nothing actually changed.

As a result, if the user only performed a "click" action in the panel content, the panel state won't change at all and the normal event chain won't be disturbed.
We also spare some React update cycles, which can benefit perceived performance.

**2. simplify the event handlers**
As another result of the event chain being complete, we don't need to manage separately the case of clicks on the panel header. The dragging handlers can be attached to the whole panel, and the click handler on the header is enough.
This allows us to remove the code which distinguished both cases using the `forceResize` var. This makes the event handling code far easier to follow.

Another related simplification would be to stop listening to `mouse*` events for the drag interactions. `touch*` should be enough theoretically. But let's do it separately only if we are sure.

## Why
 - Fixing an important issue with iOS which prevented buttons and links to be clicked inside the panel in some situations 
 - Overall simplification of this component, which is the main UI building block of our app

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -215,7 +215,6 @@ class Panel extends React.Component {
                 'panel-header',
                 { 'panel-resizeHandle': resizable && isMobile }
               )}
-              ref={element => this.handleElement = element}
               onClick={() => isMobile && this.handleHeaderClick()}
             >
               {resizable && isMobile && size === 'minimized' && minimizedTitle
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -58,7 +58,6 @@ class Panel extends React.Component {
 
   constructor(props) {
     super(props);
-    this.moveHandler = null;
     this.panelContentRef = React.createRef();
     this.state = {
       holding: false,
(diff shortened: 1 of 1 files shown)


xem (commented):
Can't test on iOS but the code seems great

amatissart (approved):

PR #760, opened 2020-08-28
Unfocus input search on submit by keyboard

bbecquet:
## Description
Unfocus the main input search after validation in the following case:
 - typing something
 - highlighting an item in the suggestion dropdown 
 - selecting this item with the enter Key

## Why
All the other ways of validating a search blur the, there's no reason this path should have a different result.

### DIFF
--- a/src/components/ui/SuggestsDropdown.jsx
+++ b/src/components/ui/SuggestsDropdown.jsx
@@ -83,7 +83,6 @@ const SuggestsDropdown = ({
 
       if (key === 'Enter') {
         if (highlighted !== null) {
-          e.preventDefault();
           onSelect(suggestItems[highlighted]);
         }
       }


src/components/ui/SuggestsDropdown.jsx
```diff
@@ -83,7 +83,6 @@ const SuggestsDropdown = ({
 
       if (key === 'Enter') {
         if (highlighted !== null) {
-          e.preventDefault();
```
bbecquet, line 86: I don't think this was useful for something else :D

xem (approved):
Works for me!

src/components/ui/SuggestsDropdown.jsx
amatissart: Arf, it was actually necessary to avoid the `keypress` event to be propagated to the input :(  
As a result, a new search will be triggered on Enter, and may select a completely different result from the one that was highlighted.

PR #762, opened 2020-09-02
QMAPS-1703 add information subtitle in poi panel when applicable

xem:
## Description
If at least one piece of information is displayed, add an "information" subtitle
NB: this should be translated as "Informations" in french.

## Screenshots
City / Hotel (subtitle):
image
image

LatLon poi (no subtitle):
image

### DIFF
--- a/src/panel/poi/PoiBlockContainer.jsx
+++ b/src/panel/poi/PoiBlockContainer.jsx
@@ -43,6 +43,17 @@ export default class PoiBlockContainer extends React.Component {
       {displayCovidInfo &&
         <CovidBlock block={covidBlock} countryCode={this.props.poi.address.country_code} />}
       <Divider />
+      {
+        (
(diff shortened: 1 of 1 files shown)


src/panel/poi/PoiBlockContainer.jsx
```diff
@@ -43,6 +43,17 @@ export default class PoiBlockContainer extends React.Component {
       {displayCovidInfo &&
         <CovidBlock block={covidBlock} countryCode={this.props.poi.address.country_code} />}
       <Divider />
+      {
+        (
+          (this.props.poi.address && this.props.poi.subClassName !== 'latlon')
+          || websiteBlock
```
bbecquet, line 49: I've been wondering for some time if there is a more elegant way to write this, and generally the block by block stuff. Sure there is, but it would require some re-org of the files first, and managing some exceptions, so :+1: for being pragmatic :smile:

bbecquet (approved):

PR #773, opened 2020-09-07
Manage "category markers" as a MapBox-gl layer

bbecquet:
## Description
Implements the POI markers as a MapBox-GL layer, with a dynamic GeoJSON source, instead as div Marker objects. This is a necessary step to be able to display these POI names on the map with correct layout.

- add functions to convert POI objects to GeoJSON (**TODO: unit tests to add.**)
- define MapBox-GL style for the new layer
- declare the new source and layer, and convert listeners to the new model
- declare a single new icon to almost match the existing style. **For now we use the same default icon for all types of POIs, so this is a regression**. I think it's acceptable to move on for now. But if we want the same look as currently, we have to convert all our icon types to the MapBox-GL system, which is *really* different from div Markers.
- manage the active states (hovered or selected) with bigger icon size. We use another layer, as MapBox "feature states" (that allow to toggle some data on features in an optimized way) don't work with the `icon-size` property :( An alternative would be to go back to Markers to manage these state. I would happily discuss it with you :)

I also found that there is a little (sadly expected) performance hit on drawing speed. Let's talk if it's worth it.


src/adapters/icon_manager.js
```diff
@@ -84,4 +84,13 @@ export function createIcon(iconOptions, name, hoverEffect = false) {
   return element.firstElementChild;
 }
 
+export function createDefaultMapIcon() {
```
G-Ray, line 87: great ! Can we set image dimension as parameters ?

G-Ray: and `src` ? As I guess we will need to use other icons :framed_picture:

src/adapters/pois_styles.js
```diff
@@ -0,0 +1,20 @@
+
```
G-Ray, line 1: :+1:

src/adapters/scene_category.js
```diff
@@ -1,21 +1,66 @@
...
 export default class SceneCategory {
   constructor(map) {
     this.map = map;
-    this.markers = [];
+
+    createDefaultMapIcon().then(imageData => {
+      this.map.addImage('pin_with_dot', imageData);
+    });
+
+    this.map.addSource(DYNAMIC_POIS_LAYER, {
```
G-Ray, line 20: Can we group similar code together ?
For instance:
1) Adding Sources
2) Adding Layers
3) Adding events handlers

Also, we can reduce layer/source code using an array like:
```js
const sources = [DYNAMIC_POIS_LAYER, ACTIVE_POIS_LAYER]
sources.map((source) => {
  this.map.addSource(source, {
    type: 'geojson',
    data: emptyFeatureCollection,
    promoteId: 'id'
  })

  this.map.addLayer({
    ...filteredPoisStyle,
    source,
    id: source,
  });
})
```

src/adapters/scene_category.js
```diff
@@ -1,21 +1,66 @@
...
+      type: 'geojson',
+      data: emptyFeatureCollection,
+      promoteId: 'id', // tells MapBox-GL to use this property as internal feature identifier
```
G-Ray, line 23: I don't understand mapbox's documentation about this `promoteId` property, and don't know why this is not the default behavior :confused:

src/adapters/scene_category.js
```diff
@@ -1,21 +1,66 @@
...
+    });
+    this.map.addLayer({
+      ...filteredPoisStyle,
+      source: DYNAMIC_POIS_LAYER,
+      id: DYNAMIC_POIS_LAYER,
+    });
+    this.map.on('click', DYNAMIC_POIS_LAYER, e => {
+      e.cancelMapClick = true; // Hack as MapBox events don't have stopPropagation
```
G-Ray, line 31:
```suggestion
      e.originalEvent.stopPropagation()
```

src/adapters/scene.js
```diff
@@ -123,6 +123,9 @@ Scene.prototype.initMapBox = function() {
     // which are thrown *after* two separate click events are thrown
     this.clickDelayHandler = null;
     this.mb.on('click', e => {
+      if (e.cancelMapClick) {
```
G-Ray, line 126:
```suggestion
      if (e.originalEvent.cancelBubble) {
        // Hack as MapBox events don't have stopPropagation
        return
      }
```

src/adapters/scene_category.js
G-Ray: See relevant suggestion here : https://github.com/QwantResearch/erdapfel/pull/773/files#r485606896 :)

G-Ray: We can rely on the `originalEvent` property to set `cancelBubble` to `true` using `stopPropagation()`.
It just implies to check for `e.originalEvent.cancelBubble` where we don't want the mapbox event to be propagated.

src/adapters/scene_category.js
```diff
@@ -99,13 +126,16 @@ export default class SceneCategory {
...
-      if (highlight) {
-        marker.classList.add('active');
-      } else {
-        marker.classList.remove('active');
-      }
+    const pois = this.selectedPoi ? [this.selectedPoi] : [];
+    if (highlight && poi !== this.selectedPoi) {
+      pois.push(poi);
     }
+    this.map.getSource(ACTIVE_POIS_LAYER).setData(poisToGeoJSON(pois));
```
G-Ray, line 133:
```suggestion
  const pois = [
    ...this.selectedPoi ? [this.selectedPoi] : [],
    ...this.selectedPoi !== highlight && poi ? [poi] : [],
  ];
  this.map.getSource(ACTIVE_POIS_LAYER).setData(poisToGeoJSON(pois));
```

What do you think about using spreads operators ?
:dog2:  Such immutable. Much functional. :dog:

src/adapters/scene_category.js
```diff
@@ -27,13 +72,17 @@ export default class SceneCategory {
     listen('click_category_poi', state => {
       this.selectPoi(state);
     });
+    listen('clean_marker', () => {
+      this.selectPoiMarker(null);
+    });
+  }
+
+  getPointedPoi = mapMouseEvent => {
+    const feature = this.map.queryRenderedFeatures(mapMouseEvent.point)[0];
```
amatissart, line 81: It should be possible to use `mapMouseEvent.features` directly, instead of calling `queryRenderedFeatures()`.

src/adapters/scene_category.js
```diff
@@ -1,21 +1,66 @@
...
+    this.map.addSource(ACTIVE_POIS_LAYER, {
+      type: 'geojson',
+      data: emptyFeatureCollection,
+      promoteId: 'id',
+    });
+    this.map.addLayer({
+      ...hoveredPoiStyle,
+      source: ACTIVE_POIS_LAYER,
+      id: ACTIVE_POIS_LAYER,
+    });
```
amatissart, line 61: Is it really necessary to define a separate source and layer to represent the active POI ?
This is a use-case where the "feature state" mechanism provided by mapbox-gl is supposed to be appropriate.
https://docs.mapbox.com/help/tutorials/create-interactive-hover-effects-with-mapbox-gl-js/

And the performance would probably be better too, as the source data would be updated less often.

src/adapters/icon_manager.js
bbecquet: Yup, good idea. It was a quick way of getting an image blob from this particular SVG but it sure looks like something that should be generic :)

src/adapters/scene_category.js
bbecquet: MapBox-GL really needs unique identifiers to work, so by default it doesn't trust the data source for that and has its own internal system to assign ids. I think that was the safest way for them, and it's ok to be able to override it :)

src/adapters/scene_category.js
bbecquet: If you really think that's more readable, why not.

src/adapters/scene_category.js
bbecquet: This is what I explained previously, this is indeed something that looks like a use-case for feature state. So that's what I tried happily at first… only to re-discover that `icon-size` and `icon-image` are "layout" properties, that don't support being set with `feature-state` (probably as it requires too much redraws).
That's another of the points where the Mapbox-GL doc is hard to parse and not explicit enough.
This is the closest official inspiration I used: https://docs.mapbox.com/android/maps/examples/icon-size-change-on-click/ (the same case but on the Android SDK :/)

src/adapters/scene_category.js
bbecquet: Thanks. I did that this way because the `mapMouseEvent.features` technique doesn't work for `mouseleave`… but I actually don't need to know the POI in this case.

src/adapters/scene_category.js
amatissart: Also the geojson standard defines an optional `id` field for the feature (outside of the `properties` object). So I think that what mapbox-gl would use by default.

src/adapters/scene_category.js
bbecquet: I used a similar form, but even simpler IMO.

src/adapters/scene_category.js
bbecquet: :+1:  The current code is the result of incremental POC development, where grouping by layers seemed more straightforward, but I agree this can be improved. I'm just waiting to be sure about how many layers we use at the end and I'll try to make it more compact and readable.

bbecquet:
Let's close this one as the alternative works better https://github.com/QwantResearch/erdapfel/pull/779

PR #776, opened 2020-09-09
QMAPS-1685 show all the panel's content during resize on mobile

xem:
## Description
On mobile, when the PoI panel appears in "default" form, it needs to contain all the panel's data already, and this data should be visible immediately when the panel is being resized to "maximized" form.
This must also work when resizing from "minimized" to "maximized" (in that case, the actions buttons and the rest of the panel should appear instantly during unfolding)

## Screenshots
resize

### DIFF
--- a/src/panel/poi/PoiPanel.jsx
+++ b/src/panel/poi/PoiPanel.jsx
@@ -224,20 +224,22 @@ export default class PoiPanel extends React.Component {
   }
 
   renderContent = (poi, { size: panelSize, isMobile }) => {
-    if (isMobile && panelSize === 'minimized') {
-      return <div className="poi_panel__content">
-        <PoiItem poi={poi} withAlternativeName />
-      </div>;
-    }
-
     return <div className="poi_panel__content">
-      <PoiItem
-        poi={poi}
-        withAlternativeName
-        withOpeningHours
-        className="u-mb-20"
-        onClick={this.center}
-      />
+      {isMobile && panelSize === 'minimized' ?
+        <PoiItem
+          poi={poi}
+          withAlternativeName
+          className="u-mb-20"
+        />
+        :
+        <PoiItem
+          poi={poi}
+          withAlternativeName
+          withOpeningHours
+          className="u-mb-20"
+          onClick={this.center}
+        />
+      }
       <div className="u-mb-8">
         <ActionButtons
           poi={poi}
@@ -248,7 +250,7 @@ export default class PoiPanel extends React.Component {
           toggleStorePoi={this.toggleStorePoi}
         />
       </div>
-      {(!isMobile || panelSize === 'maximized') && <div className="poi_panel__fullContent">
+      <div className="poi_panel__fullContent">
         <PoiBlockContainer poi={poi} covid19Enabled={covid19Enabled} />
         {isFromPagesJaunes(poi) &&
           <div className="poi_panel__info-partnership u-text--caption">
@@ -263,7 +265,7 @@ export default class PoiPanel extends React.Component {
           <CategoryList />
         </div>}
         {isFromOSM(poi) && <OsmContribution poi={poi} />}
-      </div>}
+      </div>
     </div>;
   }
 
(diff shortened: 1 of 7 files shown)


xem:
@bbecquet @G-Ray J'ai corrigé le test "a minima" cat maintenant infoTitle est constamment présent dans le panel.
A votre avis, est-il nécessaire de rajouter un test qui vérifie que lorsqu'on clique sur le header du panel et qu'il passe en default ou minified, alors la div fullContent est correctement cachée?

G-Ray:
Ça à l'air de bien fonctionner !
Est-ce que tu penses que le retour des transitions peut-être réalisé simplement grâce à ce changement ?
Pour le test supplémentaire, ça peut être intéressant de tester un ou quelques élements invisibles en plus.

src/panel/poi/PoiPanel.jsx
```diff
@@ -224,20 +224,22 @@ export default class PoiPanel extends React.Component {
...
+        />
+        :
+        <PoiItem
+          poi={poi}
+          withAlternativeName
+          withOpeningHours
+          className="u-mb-20"
+          onClick={this.center}
+        />
+      }
```
G-Ray, line 242: the only difference between the two components instances are the props `withOpeningHours` and `onClick`.
You can just add conditions to these two props to check if it's `isMobile && panelSize === 'minimized'`

xem:
Je compte retourner faire des essais de transitions dans la branche de l'autre ticket quand celui-ci sera mergé. Mais en tout cas non, ça ne sera pas trivial car il doit subsister quelques heights "auto", donc faudra voir en trichant un peu ce que ça donne. :)
Pour les tests d'éléments invisibles, j'attendais justement que vous validiez cette approche à base de height: 0 / overflow: hidden avant de modifier le test en conséquence.

src/panel/poi/PoiPanel.jsx
bbecquet: You can even spare the `isMobile` test, as `panelSize` will never equal `minimized` if we aren't on mobile.

src/scss/includes/panels/poi_panel.scss
```diff
@@ -78,6 +78,11 @@ $BLOCK_ICON_FONT_SIZE: 16px;
   .poi_panel__action__direction {
     flex-grow: 1;
   }
+
+  .minimized:not(.panel--holding) & {
+    height: 0;
```
bbecquet, line 83: Just to be sure, I'm curious about this choice. Didn't it work with simply `display: none/block`?

bbecquet (commented):
Nice and pragmatic solution, tested on real devices. Works great… as long as the design doens't change more between the three states :)

src/scss/includes/panels/poi_panel.scss
xem: It could have been a display toggle indeed, but as stated in another discussion, using the height could be part of the solution to reintroduce transitions between default abd minimized. (didn't have time to test it yet)

bbecquet: I think this is unrelated. You'll need to get the real height of the whole panel at some points. Having just a fixed height for only a subpart of it doesn't give any hint about the height of the rest. Using `auto` is still mandatory if we want to adapt to possibly very long titles and the presence/absence of some data (reviews, etc.).
So if this was the only reason to use `height` instead of `display`, you can simplify it by using the latter. Let's not introduce dependencies between separate tickets.
Even if it turns we need height in the other fix, it's pretty straightforward to change.

bbecquet (approved):
Ok to go back to the CSS-based version of that for now.
Now that the PoiPanel inner content doesn't depend on the panel size anymore, you can refactor the component to just get rid of the `renderContent` method and put its content directly in the `render`.

G-Ray (dismissed):

xem:
I removed renderContent as suggested, please revalidate :)

bbecquet (approved):

PR #777, opened 2020-09-09
QMAPS-1715 hide top-bar shadows when the current panel is maximized and not held

xem:
## Description
- hide top-bar shadows when the current panel is maximized and not held
- show the shadow again as soon as the panel is being dragged down or clicked to return to default size

## Why
to make the top-bar and panel's header merge seamlessly

## Screenshots
topbar

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -145,6 +145,11 @@ class Panel extends React.Component {
     }
 
     this.setState({ currentHeight, holding: true });
+
+    // Show top_bar shadows when holding
+    const topBar = document.querySelector('.top_bar');
+    topBar.classList.remove('top_bar--no-shadow');
+
   }
 
   /**
@@ -171,6 +176,14 @@ class Panel extends React.Component {
 
     if (newSize !== this.props.size) {
       this.props.setSize(newSize);
+
+      // Hide top_bar shadows when maximized
+      const topBar = document.querySelector('.top_bar');
+      if (newSize === 'maximized') {
+        topBar.classList.add('top_bar--no-shadow');
+      } else {
+        topBar.classList.remove('top_bar--no-shadow');
+      }
     }
     if (this.state.holding || this.state.currentHeight) {
       this.setState({ holding: false, currentHeight: null });
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -145,11 +145,6 @@ class Panel extends React.Component {
     }
 
     this.setState({ currentHeight, holding: true });
-
-    // Show top_bar shadows when holding
-    const topBar = document.querySelector('.top_bar');
-    topBar.classList.remove('top_bar--no-shadow');
-
   }
 
   /**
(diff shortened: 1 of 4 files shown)


xem:
@bbecquet I resolved this by adding a class to top-bar. I hope it's clean enough, as I didn't find any cleaner way to achieve this effect.

src/components/ui/Panel.jsx
```diff
@@ -145,6 +145,11 @@ class Panel extends React.Component {
     }
 
     this.setState({ currentHeight, holding: true });
+
+    // Show top_bar shadows when holding
+    const topBar = document.querySelector('.top_bar');
```
G-Ray, line 150: I think we can add topBar on `this` to manipulate it through the component

bbecquet: By doing this top bar manipulation in `<Panel>`, you're breaking the separation of concerns and introduce a new coupling between separated parts. You also introduce a new risk of inconsistent state if you forget some transitions. For exemple, this code doesn't manage the case of dragging from `'maximized'` and releasing to `'maximized'`. The shadow is restored but not removed at the end.

To avoid such problems, as this new class depends entirely on the `size` state, you need to find where it's actually stored and managed.
As we needed to control the size of the panel from outside, [we recenlty introduced a change in that](https://github.com/QwantResearch/erdapfel/pull/749): `size` is now passed as a prop to this component and managed in the state of PanelManager. So this is where your new logic should be moved. 

And that's good news, as there are already a lot of direct manipulation of the top bar element there :) We're beginning to see a future separated component emerge, when we'll be able to render the top bar in React.

bbecquet:
@xem see mainly my [big comment](https://github.com/QwantResearch/erdapfel/pull/777#discussion_r486189028) about where this should be moved to.

We also need to address the problem of the white to grey separation. I propose we move forward and change the style of mobile list panels here, as it's straightforward enough. Here is an example with the category panel. We need to confirm it though and check that the two other grey panels (Route result and Favorites) are ok.

|Currently|Proposed|
|---|---|
|localhost_3000_places__type=restaurant(iPhone 6_7_8) (5)|localhost_3000_places__type=restaurant(iPhone 6_7_8) (6)|

src/components/ui/Panel.jsx
xem: I totally underastand the problem shown here, though I'm don't think I can reliably update the topBar shadows in PanelManager: this component doesn't know when the panel is maximized, nor when it is being dragged. Only Panel knows that. I'm not sure how to change my code to move the logic here. @bbecquet

bbecquet: > PanelManager doesn't know when the panel is maximized

Yes it does. As I explained in the previous comment, `PanelManager` is actually where the current panel size is stored now, in the state, so it can be controlled by stuff outside of the `Panel` component itself.
You can find it here: https://github.com/QwantResearch/erdapfel/blob/master/src/panel/PanelManager.jsx#L32
And as any React component state, you can set it, read it, and react to changes on it by using `componentDidUpdate`.

> nor when it is being dragged

That point is true, only the `Panel` knows that. But actually here you don't need to know when it's dragged, only when it's set to maximized.

bbecquet: @xem Ok, I think I see your point. It's probably better visually if we restore the shadow as soon as we are dragging, is it what you wanted?

G-Ray: @xem What about using z-index to cover the top bar-shadow ? It would be so much easier.
This way, we would only need to handle panel's shadow visibility

xem: @bbecquet yep it's what I meant.
Thanks a lot for these tips, I'm on it!

@G-Ray indeed, a bigger z-index in the panel_container div can invert the order of the panel and the top bar. I'm digging this way too!

xem: It works perfectly! We now have a pure CSS fix for this ticket :D

G-Ray (approved):

PR #778, opened 2020-09-11
Display NoResultPanel when no result are returned for query

G-Ray:
## Description
Display no result message in a panel when there is not result for current query (from search input or url query param)

## Screenshots
desktop

mobile gif

### DIFF
--- a/src/panel/NoResultPanel.jsx
+++ b/src/panel/NoResultPanel.jsx
@@ -1,0 +1,37 @@
+/* globals _ */
+import React from 'react';
+
+import { Panel } from 'src/components/ui';
+import SearchInput from '../ui_components/search_input';
+
+const close = () => window.app.navigateTo('/');
+
+const handleSearchClick = e => {
+  e.preventDefault();
+  close();
+  SearchInput.setInputValue('');
+  SearchInput.select();
+};
+
+const NoResultPanel = () => {
+  return (
+    <Panel
+      white
+      close={close}
+    >
+      <div
+        style={{ padding: '0 16px 32px 16px' }}
+      >
+        <p className="u-mb-8 u-text--smallTitle">{_('No results found.')}</p>
+        <p className="u-text--subtitle u-mb-20">
+          {_('Check the spelling of your search or add more details, such as city or country.')}
+        </p>
+        <a onClick={handleSearchClick} href="#">
+          {_('Try a new search query')}
+        </a>
+      </div>
+    </Panel>
+  );
+};
+
+export default NoResultPanel;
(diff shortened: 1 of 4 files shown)


bbecquet:
Can you put a screenshot of the resulting panel in the desc please?

src/panel/category/CategoryPanelError.jsx
```diff
@@ -1,22 +1,26 @@
...
-      {zoomIn
-        ? _('Please zoom in the map to see the results for this category.', 'categories')
-        : _('No results found. Please zoom out of the map.', 'categories')
-      }
-    </p>
-    <button onClick={geoloc}>
-      <span className="icon-pin_geoloc" /> {_('Search around my position', 'categories')}
-    </button>
-  </div>;
+  return (
```
bbecquet, line 11: Not a fan of putting this unrelated change in the same PR.

src/scss/includes/panels/categories.scss
```diff
@@ -38,14 +38,6 @@
 
 .category__panel__error {
   padding: 30px 40px 50px;
```
bbecquet, line 40: If you want to go on with this change for visual consistency, let's set the bottom padding to the same value as the other panel…

src/panel/NoResultPanel.jsx
```diff
@@ -0,0 +1,34 @@
...
+      white
+      close={close}
+    >
+      <Flex
+        flexDirection="column"
+        alignItems="center"
+        justifyContent="center"
+        style={{ padding: '0 16px 32px 16px' }}
+      >
+        <span className="u-mb-8 u-text--heading6">{_('No results found.')}</span>
```
bbecquet, line 25: I would add some more margin here, the two elements are really close.

bbecquet (commented):
Is this design Panel also validated on mobile?

src/components/ui/Flex.jsx
```diff
@@ -8,6 +8,7 @@ const Flex = forwardRef(({
   className,
   justifyContent,
   alignItems = 'center',
+  flexDirection,
```
bbecquet, line 11: Changes on this file aren't required anymore by the rest of the diff, they can be removed.

src/panel/NoResultPanel.jsx
```diff
@@ -0,0 +1,36 @@
...
+import React from 'react';
+
+import { Panel } from 'src/components/ui';
+
+const close = () => window.app.navigateTo('/');
+
+const handleSearchClick = e => {
+  e.preventDefault();
+  close();
+  window.__searchInput.searchInputHandle.value = '';
```
bbecquet, line 11: To be tested, but you should be able to avoid this global stuff manip by importing `SearchInput` and calling `setInputValue` then `select`.

G-Ray: I confirm it works! thansk

bbecquet (approved):

PR #779, opened 2020-09-14
Manage "category markers" as GL layer, DOM markers for active state

bbecquet:
## Description
Alternative implementation to https://github.com/QwantResearch/erdapfel/pull/773.
Same overall principle:
 - the set of map markers from a category or intention search is now using its own MapBox-GL layer and style, like the other POI layers
 - this layer is using a dynamic vector source as data. We inject the filtered POIs as GeoJSON data in this source
 - we are still using a single pin marker for all POI classes.

**Difference:**
MapBox-GL has very limited support to manage dynamic changes on map features. Although the API has the [`feature-state` mechanism](https://docs.mapbox.com/help/tutorials/create-interactive-hover-effects-with-mapbox-gl-js/), you can change only some limited style properties on-the-fly with it. This doesn't work for all "layout properties", including icon or text size. 
As a result, the [previous PR](https://github.com/QwantResearch/erdapfel/pull/773) used a different vector source/layer/style to achieve the dynamic effect. We injected dynamically the hovered and/or selected POI in it, also as GeoJSON. But this caused a big perceived performance degradation, as the map was fully redrawn each time…
The current PR proposes another way around this performance hit, by re-introducing DOM marker overlays only on user interaction cases.

I took a shorcut here: as the hover icon covers the normal icon entirely, I didn't change anything on the normal layer while hovering. If we needed to hide it, we would be able to use `feature-state` in this case with `icon-opacity` set to 0, as this property [supports it](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#paint-symbol-icon-opacity)

Peek 14-09-2020 10-55


src/adapters/pois_styles.js
```diff
@@ -0,0 +1,10 @@
+
+export const filteredPoisStyle = {
+  type: 'symbol',
+  layout: {
+    'icon-image': 'pin_with_dot',
+    'icon-allow-overlap': true,
```
amatissart, line 6: I am not sure it would make any difference here, but it would be more consistent to me to set [`icon-ignore-placement`](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-icon-ignore-placement) too:
```js
'icon-ignore-placement': true,
```

src/adapters/pois_styles.js
```diff
@@ -0,0 +1,10 @@
...
+    'icon-size': 0.5,
+    'icon-anchor': 'bottom',
```
amatissart, line 8: (Very minor)

Given the shape of the icon this anchor is the right thing to do :+1: 
Note that it was not the case before, so this makes the popup position a bit off:

|Before|After|
|-|-|
|image|image|

src/adapters/scene_category.js
```diff
@@ -1,21 +1,63 @@
...
+    });
+    this.map.on('click', DYNAMIC_POIS_LAYER, e => {
+      e.originalEvent.stopPropagation();
+      const poi = this.getPointedPoi(e);
+      this.selectPoi({
+        poi,
+        pois: this.pois,
+        poiFilters: this.poiFilters,
+      });
+    });
```
amatissart, line 47: Possible edge-case here:
when several POIs are very close from each other and the mouse moved to another poi after hover (without triggering `mouseleave`), the clicked `poi` (from the dynamic layer) may be different from the highlighted one which is described in the current popup.

An alternative would be to always select the highlighted poi here, independently from `e` ? With potentially other bugs ?

amatissart (commented):
The performance and the rendering looks good to me. I just made a few minor comments.

The different icons (representing the poi class/subclass) are lost, but this could probably be fixed in a separate PR.

src/adapters/pois_styles.js
bbecquet: Ah yes, thanks! More street names are visible this way.

src/adapters/pois_styles.js
bbecquet: Ok, I've changed it globally in the Popup util for now, it would be too much refacto here to adapt to all cases. I put a value that ensures the popup doesn't overlap with these hovered marker while not being too far for the "normal" POIs.

src/adapters/scene_category.js
bbecquet: Haha, interesting edge case, you're always good with finding them :D
How I understand it, it's more a problem with the hovered POI (and therefore the popup) not being correcly updated than a problem with the clicked one, right?
Peek 14-09-2020 18-07

amatissart: You're right. Ideally the highlighted poi and the popup would always follow the mouse, to be consistent the feature triggered on click.
But that's probably not feasible with only `mouseenter`/`mouseleave` events. So I suppose we would need to listen to `mousemove` events and introduce more complexity; I am not sure it's worth it.

bbecquet: @amatissart @xem I gave it a try with `mousemove`. For the user it works by just replacing `mouseenter` by it. But it triggers invisible re-renders on each mouse movement, so I checked the POI actually changed before touching the marker or popup. I think it's not too complex after all.

src/adapters/scene_category.js
```diff
@@ -1,21 +1,68 @@
...
+    this.hoveredMarker = new Marker({
+      element: createIcon({ disablePointerEvents: true }),
+      anchor: 'bottom',
+    });
+    this.selectedMarker = new Marker({
+      element: createIcon({}),
+      anchor: 'bottom',
+    });
+
+    createMapGLIcon('./statics/images/map/pin_map_dot.svg', 50, 60)
```
G-Ray, line 25: There is a lot of code in this constructor. I think we can hide the complexity of each instruction by defining standalone functions. e.g: 
```js
this.map.on('click', handlePoiClicked).
```

bbecquet:
@G-Ray I pushed another commit with an hopefully easier to read structure for the whole class. Thanks for the suggestions.

tests/units/geojson.js
```diff
@@ -0,0 +1,47 @@
+import { poiToGeoJSON, poisToGeoJSON } from '../../src/libs/geojson';
+import IdunnPOI from '../../src/adapters/poi/idunn_poi';
+const mockPoi1 = require('../__data__/poi.json');
+const mockPoi2 = require('../__data__/poi2.json');
+
+describe('geojson', () => {
```
G-Ray, line 6: :+1:

G-Ray (approved):

xem (dismissed):
Approving code and local test.
As discussed on Slack, the markers hover effect and tooltip position need a bit of changes.
Also, the tests don't pass yet, but I suppose they will soon :)

amatissart (approved):

PR #783, opened 2020-09-16
QMAPS-1723 make the panel close button bigger on mobile

xem:
## Description
Make the panel close button bigger on mobile (35x50px)

## Why
Misclicks on the upper half of the button often provoke a panel minimization instead of closing it
25x25px is too small for touch inputs

## Screenshots
before:
image

after:
image

### DIFF
--- a/src/scss/includes/components/panel.scss
+++ b/src/scss/includes/components/panel.scss
@@ -58,6 +58,8 @@
 
     &-close {
       color: $primary_text;
+      height: 50px;
+      width: 35px;
     }
 
     &--white {


amatissart:
Thanks! We will need to test this change on a real iOS device.  
I wonder if using a proper `<button>` element could also help with this, instead of the current `<div onClick=...>` built with the Flex component.

G-Ray:
The button changed position in the following PR : https://github.com/QwantResearch/erdapfel/pull/786
The close button is now bigger. Could you test there and suggest any changer if it's still needed please ?

G-Ray:
@xem I let you close this as the fix has already been merged ?

PR #785, opened 2020-09-16
QMAPS-1722 fix long touch on map on mobile

xem:
## Description
Implement a long touch on the map (500ms) that works on both Android and Safari

## Why
Mapbox-gl-js oncontextmenu event on the map only worked for Android for a long time, and now doesn't even work for android.
So we reimplement it with a cancellable timeout triggered on touchstart, and many conditions that are able to cancel it (2 fingers on screen, touchmove, touchend, gesturestart, etc...).
This implementation contextmenu-free should also fix the behavior on iOS (as well as Android), to be confirmed on real device.

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -23,6 +23,8 @@ const baseUrl = nconf.get().system.baseUrl;
 
 const store = new LocalStore();
 
+let longTouchTimeout = null;
+
 function Scene() {
   this.currentMarker = null;
   this.popup = new PoiPopup();
@@ -145,10 +147,29 @@ Scene.prototype.initMapBox = function() {
       }, this.DOUBLE_TAP_DELAY_MS);
     });
 
+    // Long touch on mobile (500ms)
+    // Custom implementation because the contextmenu event isn't supported by MapBox.
+    // Long touch is initiated with a touchstart event
+    // It is canceled if many fingers touch the map, or if one of these events is initiated:
+    // touchmove, touchcancel, touchend, pointerdrag, pointermove, moveend, gesturestart, gesturechange, gestureend
     if (isMobileDevice()) {
-      this.mb.on('contextmenu', e => {
-        this.clickOnMap(e.lngLat, null, { longTouch: true });
+      this.mb.on('touchstart', e => {
+        if (e.originalEvent.touches.length === 1) {
+          longTouchTimeout = setTimeout(() => {
+            this.clickOnMap(e.lngLat, null, { longTouch: true });
+          }, 500);
+        }
       });
+
+      this.mb.on('touchend', this.cancelLongTouch);
+      this.mb.on('touchcancel', this.cancelLongTouch);
+      this.mb.on('touchmove', this.cancelLongTouch);
+      this.mb.on('pointerdrag', this.cancelLongTouch);
+      this.mb.on('pointermove', this.cancelLongTouch);
+      this.mb.on('moveend', this.cancelLongTouch);
+      this.mb.on('gesturestart', this.cancelLongTouch);
+      this.mb.on('gesturechange', this.cancelLongTouch);
+      this.mb.on('gestureend', this.cancelLongTouch);
     }
 
     this.mb.on('dragstart', () => { fire('map_user_interaction'); });
@@ -429,6 +450,10 @@ Scene.prototype.mobileButtonVisibility = function(selector, visible) {
       item.classList.add('hidden');
     }
   }
+};
+
+Scene.prototype.cancelLongTouch = () => {
+  clearTimeout(longTouchTimeout);
 };
 
 export default Scene;
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -20,10 +20,8 @@ import { fire, listen } from 'src/libs/customEvents';
 import locale from '../mapbox/locale';
 
 const baseUrl = nconf.get().system.baseUrl;
-
 const store = new LocalStore();
-
-let longTouchTimeout = null;
+const LONG_TOUCH_DELAY_MS = 500;
 
 function Scene() {
   this.currentMarker = null;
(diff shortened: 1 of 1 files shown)


src/adapters/scene.js
```diff
@@ -145,10 +147,29 @@ Scene.prototype.initMapBox = function() {
...
+    // It is canceled if many fingers touch the map, or if one of these events is initiated:
+    // touchmove, touchcancel, touchend, pointerdrag, pointermove, moveend, gesturestart, gesturechange, gestureend
     if (isMobileDevice()) {
-      this.mb.on('contextmenu', e => {
-        this.clickOnMap(e.lngLat, null, { longTouch: true });
+      this.mb.on('touchstart', e => {
+        if (e.originalEvent.touches.length === 1) {
+          longTouchTimeout = setTimeout(() => {
+            this.clickOnMap(e.lngLat, null, { longTouch: true });
+          }, 500);
```
G-Ray, line 160: can we use/modify DOUBLE_TAP_DELAY_MS ?

src/adapters/scene.js
```diff
@@ -23,6 +23,8 @@ const baseUrl = nconf.get().system.baseUrl;
 
 const store = new LocalStore();
 
+let longTouchTimeout = null;
```
bbecquet, line 26: You can move it inside your `if(mobileDevice()) {}` block below. That way all the code related to the long touch is close and can be read and maintained more easily.

src/adapters/scene.js
```diff
@@ -431,4 +452,8 @@ Scene.prototype.mobileButtonVisibility = function(selector, visible) {
   }
 };
 
+Scene.prototype.cancelLongTouch = () => {
```
bbecquet, line 455: See the comment above. You don't act on any part of the object here so it doesn't need to be a class method.

src/adapters/scene.js
```diff
@@ -145,10 +147,29 @@ Scene.prototype.initMapBox = function() {
...
-        this.clickOnMap(e.lngLat, null, { longTouch: true });
+      this.mb.on('touchstart', e => {
+        if (e.originalEvent.touches.length === 1) {
+          longTouchTimeout = setTimeout(() => {
+            this.clickOnMap(e.lngLat, null, { longTouch: true });
+          }, 500);
+        }
       });
+
+      this.mb.on('touchend', this.cancelLongTouch);
```
bbecquet, line 164: Instead of the big redundancy in these instructions and comments, it would be simpler to declare an array with a good name and loop.

```js
const longTouchCancellingEvents = [ 'touchend', 'touchcancel', 'touchmove' /*…*/ ];
const cancelLongTouch = () => { clearTimeout(longTouchTimeout); }
longTouchCancellingEvents.forEach(event => { this.mb.on(event, cancelLongTouch);  });
```

That way everything related to cancellation is grouped, you don't need the class method and you don't need the redundant comment with the event list, as the code is self-explanatory.

src/adapters/scene.js
bbecquet: That would be interesting indeed to explore the possible relationship with these two values. Can they be the same? How did you decide about 500ms?
I would just avoid modifying the current value of DOUBLE_TAP_DELAY_MS as this is the value used by MapBox internal implementation and it seems right UX-wise.

src/adapters/scene.js
```diff
@@ -145,10 +147,29 @@ Scene.prototype.initMapBox = function() {
       }, this.DOUBLE_TAP_DELAY_MS);
     });
 
+    // Long touch on mobile (500ms)
+    // Custom implementation because the contextmenu event isn't supported by MapBox.
+    // Long touch is initiated with a touchstart event
+    // It is canceled if many fingers touch the map, or if one of these events is initiated:
+    // touchmove, touchcancel, touchend, pointerdrag, pointermove, moveend, gesturestart, gesturechange, gestureend
```
bbecquet, line 154: We don't need to list every events in the comments here, they already very readable in the code (see my other comment).
BTW, did you establish this list yourself? Or if it's inspired by a documented example, it would be more interesting to have the link to it in these comments.

bbecquet (commented):
Some comments about coding style, but overall that's a great improvement!
I tested it on real iOS and Android devices, it works really great :+1:

src/adapters/scene.js
```diff
@@ -145,10 +147,29 @@ Scene.prototype.initMapBox = function() {
...
     if (isMobileDevice()) {
```
bbecquet, line 155: I think we should get rid of this condition. Listening to touch-specific events should be enough. And that way we'll support big screens (not considered as mobile) with touch support.
More generally, we should use `isMobileDevice` only when there aren't more reliable alternatives.

src/adapters/scene.js
xem: Long press duration is not a magic value unfortunately. It's specific to Mobile OS's and on certain devices you can even modify the long touch duration in the OS settings, but afaik there's no way to read it in JS.
I chose 500ms as it seems to be the default value on Android.
I would not recommend using DOUBLE_TAP_DELAY_MS as it has a wrong name and 300ms seems too short in my opinion.
However I recommend creating a LONG_TOUCH_DELAY_MS constant equal to 500 or any other value you prefer.
@bbecquet @G-Ray

src/adapters/scene.js
xem: I thought about doing it initially, but decided against it as it seemed a bit less simple to read in this form.
However, I agree to do the change you propose, it's fine too. :)

src/adapters/scene.js
bbecquet: So what I understand is this number comes from knowledge of the mobile platform, and you use it so our app behavior is consistent with user interaction on native apps.
I agree defining a LONG_TOUCH_DELAY_MS const would be nice. And if you have a link to a doc or example justifying its 500ms value, you can add it in comments like we did for DOUBLE_TAP_DELAY_MS.

bbecquet (approved):
Thanks for the UI improvement and for the refacto :+1: It looks less threatening to maintain this way :D

PR #786, opened 2020-09-16
Refactor Panel navigation with new design

G-Ray:
## Description
- Fix close button position/design
- Now, components using `Panel` can use `renderHeader` to render a header on top of the panel.
- Add `PanelNav` component on desktop to pass as a `renderHeader` props to Panel
- Removed obsolete `title` prop from Panel.
- Remove obsolete `close` prop from Panel.
- Simpler Panel drawer/header style, no absolute positioning anymore.
- styles cleanup

## Screenshots
|Before|After|
|---|---|
|image|image|
|image | image|
image|image|

Please note : On the bottom-left screenshot, there is no close button anymore. The close action can still be done from the search bar, which should be the targetted UX behavior. If you think it's too early to remove this button, let me know :)

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -4,7 +4,6 @@ import classnames from 'classnames';
 import { fire } from 'src/libs/customEvents';
 import { DeviceContext } from 'src/libs/device';
 import { PanelContext } from 'src/libs/panelContext';
-import Flex from 'src/components/ui/Flex';
 
 const getEventClientY = event => event.changedTouches
   ? event.changedTouches[0].clientY
@@ -39,7 +38,6 @@ function getTargetSize(previousSize, startHeight, endHeight, maxSize) {
 class Panel extends React.Component {
   static propTypes = {
     children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
-    title: PropTypes.node,
     minimizedTitle: PropTypes.node,
     resizable: PropTypes.bool,
     size: PropTypes.string,
@@ -192,8 +190,8 @@ class Panel extends React.Component {
 
   render() {
     const {
-      children, title, minimizedTitle,
-      resizable, className, white, size, renderNav } = this.props;
+      children, minimizedTitle,
+      resizable, className, white, size, renderHeader } = this.props;
     const { currentHeight, holding } = this.state;
 
     return (
@@ -209,19 +207,17 @@ class Panel extends React.Component {
             onTransitionEnd={() => this.updateMobileMapUI()}
             {...(isMobile && resizable && this.getEventHandlers())}
           >
-            {(isMobile || title) && <Flex
-              justifyContent="space-between"
-              className={classnames(
-                'panel-header',
-                { 'panel-resizeHandle': resizable && isMobile }
-              )}
-              onClick={this.handleHeaderClick}
-            >
-              {resizable && isMobile && size === 'minimized' && minimizedTitle
-                ? <span className="minimizedTitle">{minimizedTitle}</span>
-                : title}
-            </Flex>}
-            {renderNav}
+            {isMobile &&
+              <div
+                className="panel-header"
+                onClick={() => this.handleHeaderClick()}
+              >
+                {resizable && isMobile && <div className="panel-resizeHandle"/>}
+                {resizable && isMobile && size === 'minimized' && minimizedTitle
+                && <span className="minimizedTitle u-text--subtitle">{minimizedTitle}</span>}
+              </div>
+            }
+            {size !== 'minimized' && renderHeader}
             <div
               className="panel-content"
               ref={this.panelContentRef}
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -209,10 +209,10 @@ class Panel extends React.Component {
           >
             {isMobile &&
               <div
-                className="panel-header"
+                className="panel-drawer"
                 onClick={() => this.handleHeaderClick()}
               >
-                {resizable && isMobile && <div className="panel-resizeHandle"/>}
+                {resizable && isMobile && <div className="panel-handle"/>}
                 {resizable && isMobile && size === 'minimized' && minimizedTitle
                 && <span className="minimizedTitle u-text--subtitle">{minimizedTitle}</span>}
               </div>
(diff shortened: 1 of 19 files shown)


src/scss/includes/components/panel.scss
```diff
@@ -13,13 +13,15 @@
...
   width: $panel_width;
 
-  &-header {
-    padding: 0 15px;
-    min-height: 40px;
-    color: #5c6f84;
-    span {
-      width: 100%;
-    }
+  &-drawer {
```
bbecquet, line 16: According to the design, this drawer should be 20px high. Here we see there isn't enough vertical space between the panel top and the content.

src/scss/includes/components/panel.scss
```diff
@@ -83,41 +78,17 @@
...
-        position: absolute;
-        top: 8px;
-        left: 50%;
-        width: 40px;
-        height: 5px;
-        transform: translateX(-50%);
-        border-radius: 2.5px;
-        background-color: #e0e1e6;
-      }
+    .panel-handle {
```
bbecquet, line 86: What is the point of making it a real element instead of pseudo? It was the same amout of CSS but less DOM.

src/scss/includes/components/panel.scss
```diff
@@ -36,13 +38,6 @@
       color: $primary_text;
     }
   }
-
-  &-content {
```
bbecquet, line 40: I understand you want to get rid of that, but some size limitation is still necessary on desktop so that the content is scrollable. Having the header optional probably makes things more complicated now. Maybe we can find a better way by limiting the `.panel` height itself and use flex for the internal layout between header and content?

src/components/ui/index.js
```diff
@@ -23,4 +26,5 @@ export {
   PlaceholderText,
   Suggest,
   Text,
+  PanelNav,
```
bbecquet, line 29: :+1:

src/components/ui/Panel.jsx
```diff
@@ -211,27 +207,17 @@ class Panel extends React.Component {
...
-              {close &&
-              <Flex
-                justifyContent="center"
-                className="panel-close"
-                title={_('Close')}
-                onClick={e => {e.stopPropagation(); close(e);}}
+            {isMobile &&
+              <div
+                className="panel-drawer"
+                onClick={() => this.handleHeaderClick()}
```
bbecquet, line 213: When `resizable` is false this shouldn't be listened to.

src/components/ui/Panel.jsx
```diff
@@ -211,27 +207,17 @@ class Panel extends React.Component {
...
               >
-                <i className="icon-x" />
-              </Flex>}
-            </Flex>
+                {resizable && isMobile && <div className="panel-handle"/>}
```
bbecquet, line 215: You can remove `isMobile` on these two lines as we are already in a block assuming it's true.

bbecquet, line 215: Related to my comment about the grey handle being a pseudo-element, this could simply be replaced by an optional classname on `.panel-drawer`.

bbecquet (commented):
The back button looks really good, and it seems we are close to a great overall simplification :+1: 

Some comments about structure and a bug (desktop panels not limited in height anymore).
Another one that had no line change to refer to: the favorite panel lost its title.

Capture d’écran de 2020-09-18 15-49-04
Capture d’écran de 2020-09-18 15-48-56

src/scss/includes/components/panel.scss
G-Ray: Done, thanks

src/scss/includes/components/panel.scss
G-Ray: The minimized title can be "pushed" from the drawer by making it a real elem

src/scss/includes/components/panel.scss
G-Ray: Yep, I remove at one time trying to fix an issue. I fixed it, thanks

G-Ray:
Favorite title has been fixed/tweaked, thanks

src/components/ui/Panel.jsx
```diff
@@ -211,27 +207,17 @@ class Panel extends React.Component {
...
-                onClick={e => {e.stopPropagation(); close(e);}}
+            {isMobile && resizable &&
+              <div
+                className="panel-drawer"
+                onClick={() => this.handleHeaderClick()}
               >
-                <i className="icon-x" />
-              </Flex>}
-            </Flex>
+                {<div className="panel-handle"/>}
```
bbecquet, line 215: You don't need the braces anymore.

src/scss/includes/components/panel.scss
bbecquet: Meh, not really convinced.

src/scss/includes/utils.scss
```diff
@@ -68,6 +68,10 @@
   font-weight: bold;
 }
 
+.u-mt-8 {
```
bbecquet, line 71: Not a fan of having utils for both top and bottom. I think we agree in trying to stack elements by using mostly bottom margins, and this would complexify stuff.

src/scss/includes/components/panel.scss
```diff
@@ -12,14 +14,19 @@
...
-    padding: 0 15px;
-    min-height: 40px;
-    color: #5c6f84;
-    span {
-      width: 100%;
-    }
+  display: flex;
+  flex-direction: column;
+
+  &-drawer {
```
bbecquet, line 20: It needs a `flex-shrink: 0` too.
This would solve it being actually 13px and too close to the panel titles below. This removes the need for a top margin on the Favorites title.

bbecquet (commented):
A lot of small visual changes were introduced as side effects, most of them look a bit odd.

The "Directions" title is too close from the handle:
Capture d’écran de 2020-09-21 15-28-15

The header grew in height:
Capture d’écran de 2020-09-21 15-28-08

Same here. We'll remove it soon but still:
Capture d’écran de 2020-09-21 15-28-00

The bottom margin seems unbalanced compared to the distance with the handle
Capture d’écran de 2020-09-21 15-26-53

xem (commented):
- Local tests are ok for me
- Code is OK besides the open discussions
- I only have two remarks, one is for design (why don't we write somewhere "return to the list" on mobile) and the other is for issue QMAPS-1723 (can we make the button 35x35 px to make it easier to touch on mobile?)

xem (approved):

bbecquet (approved):

PR #788, opened 2020-09-18
[PoC] React server-side rendering: first step

bbecquet:
## Description
A quick PoC to lay the first bricks of server-side rendering of React components.

There are many goals for that (first load perf, crawlability, etc.), but the first would be to simplify our top bar/main search implementation by moving in to a React component. The target is to have one render tree with shared state, instead of using DOM selections and global functions to make React and static HTML communicate. Hopefully we'll reduce overall complexity, ease top bar maintenance and avoid frequent bugs concerning the search.

SSR is not trivial to set-up from scratch and there are many gotchas, so it's better if we go slowly :slightly_smiling_face: 
This PR is a basis so we can discuss the process and explain all details. It includes only the very first steps:
 1. ensure React can be used on the server (In practice, this means mainly using Babel on the server code too so JSX is recognized)
 2. define a new step when serving the index page, so we can render React components to strings and inject them in the HTML template. Also pass necessary bits (i18n and config).
 3. define a new `TopBar.jsx` component and render it in place of the current static markup
 4. ensure event listeners mapped to global functions are working

That's it. It doesn't do anything more for now, that's pure server-side rendering with no client counterparts. The same as the current implementation, except with React instead of HTML. Fun stuff and real benefits will come later, but it needs these first step to be well understood first.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env", "@babel/preset-react"]
+}
--- a/bin/SSR/reactSSR.js
+++ b/bin/SSR/reactSSR.js
@@ -1,0 +1,16 @@
+/* globals module */
+
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
+import TopBar from '../../src/components/TopBar.jsx';
+
+module.exports = function(html, { i18n, config }) {
+  const topBar = ReactDOMServer.renderToString(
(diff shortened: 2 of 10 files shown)


bbecquet:
Closing this as it's stale and only served as a support for explanation. Not sure it will be done like this in the future, and in any case it's not hard to extract parts.

PR #789, opened 2020-09-18
[Telemetry] Improve metrics of POI panel actions

bbecquet:
## Description
Add new telemetry events or some details to existing events in POI panel metrics:
 - clicks on Itinerary button
 - clicks on OSM view/edit links
 - target of the Share feature ('facebook', 'twitter' or 'copy')
 - status of the POI in the favorite store

Better reviewed commit by commit.

### DIFF
--- a/local_modules/telemetry/index.js
+++ b/local_modules/telemetry/index.js
@@ -32,6 +32,7 @@ module.exports = {
     'poi_osm_favorite', // Favorite toggle
     'poi_osm_phone',
     'poi_osm_website',
+    'poi_osm_itinerary',
     /* Pages Jaunes Poi */
     'poi_pages_jaunes_open',
     'poi_pages_jaunes_go',
@@ -39,6 +40,7 @@ module.exports = {
     'poi_pages_jaunes_phone',
     'poi_pages_jaunes_website',
     'poi_pages_jaunes_reviews',
+    'poi_pages_jaunes_itinerary',
     /* Map */
     'localise_trigger',
     /* Covid-19 */
--- a/local_modules/telemetry/index.js
+++ b/local_modules/telemetry/index.js
@@ -26,6 +26,7 @@ module.exports = {
     'poi_backtolist',
     'poi_restore',
     'poi_share',
+    'poi_share_to',
     /* OSM */
     'poi_osm_open',
     'poi_osm_go',
--- a/local_modules/telemetry/index.js
+++ b/local_modules/telemetry/index.js
@@ -33,6 +33,8 @@ module.exports = {
     'poi_osm_phone',
     'poi_osm_website',
     'poi_osm_itinerary',
+    'poi_osm_osm_view',
+    'poi_osm_osm_edit',
     /* Pages Jaunes Poi */
     'poi_pages_jaunes_open',
     'poi_pages_jaunes_go',
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -27,8 +27,6 @@ export default class DirectionPanel extends React.Component {
   constructor(props) {
     super(props);
 
-    Telemetry.add(Telemetry.ITINERARY_OPEN);
-
     this.vehicles = [modes.DRIVING, modes.WALKING, modes.CYCLING];
     if (this.props.isPublicTransportActive) {
       this.vehicles.splice(1, 0, modes.PUBLIC_TRANSPORT);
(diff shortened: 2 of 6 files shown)


src/panel/poi/ActionButtons.jsx
```diff
@@ -17,6 +17,10 @@ const ActionButtons = ({
     return handler(e);
   };
 
+  const onShare = target => {
+    Telemetry.add(Telemetry.POI_SHARE_TO, { target, id: poi.id });
```
G-Ray, line 21: do we need to specify the poi.id ?

bbecquet: Maybe not. I'm 100% sure it will be asked in the future, but let's remove them for now.

local_modules/telemetry/index.js
```diff
@@ -26,19 +26,24 @@ module.exports = {
...
     'poi_restore',
     'poi_share',
+    'poi_share_to',
     /* OSM */
     'poi_osm_open',
     'poi_osm_go',
     'poi_osm_favorite', // Favorite toggle
     'poi_osm_phone',
     'poi_osm_website',
+    'poi_osm_itinerary',
```
G-Ray, line 36: it does not seem to be used anywhere ?

src/panel/direction/DirectionPanel.jsx
```diff
@@ -27,15 +27,7 @@ export default class DirectionPanel extends React.Component {
...
-    Telemetry.add(Telemetry.ITINERARY_OPEN,
-      props.poi && Telemetry.buildInteractionData({
-        id: props.poi.id,
-        source: props.poi.meta ? props.poi.meta.source : props.poi.name,
-        template: 'simple',
-        zone: 'detail',
-        element: 'itinerary',
-      })
-    );
+    Telemetry.add(Telemetry.ITINERARY_OPEN);
```
G-Ray, line 30: shouldn't we move this in a DidMount lifecycle ?

local_modules/telemetry/index.js
```diff
@@ -26,19 +26,24 @@ module.exports = {
...
+    'poi_osm_osm_view',
+    'poi_osm_osm_edit',
     /* Pages Jaunes Poi */
     'poi_pages_jaunes_open',
     'poi_pages_jaunes_go',
     'poi_pages_jaunes_favorite', // Favorite toggle
     'poi_pages_jaunes_phone',
     'poi_pages_jaunes_website',
     'poi_pages_jaunes_reviews',
+    'poi_pages_jaunes_itinerary',
```
G-Ray, line 46: I'm unable to find if this is used somewhere ?

bbecquet: This is because of the system which builds poi event keys by concatenating their source and the event action… 
In https://github.com/QwantResearch/erdapfel/pull/781 I introduced the `sendPoiEvent` helper, so you have to look for a call like `Telemetry.sendPoiEvent(poi, 'itinerary');`.
It will send a `poi_pages_jaunes_itinerary` event for PagesJaunes POI and a `poi_osm_itinerary` for OSM POI.

`sendPoiEvent` was meant only as a technical simplification, but it didn't address the root of the problem, which is it's hard to find call points for these events…

Here I mimicked what was done before, because it thought it was important that we have this disctinction. But actually I'm not sure (what is sure with telemetry anyway?…)
Maybe we should just get rid of the distinction of POI sources in the event names, even for the preexisting ones. That way we could send `Telemetry.POI_ITINERARY`, that's it.

WDYT @G-Ray @amatissart ?

G-Ray: Oh yes, you explained that to me earlier, however.
Ideally, I think we should use define one event per poi, and pass its type only as data, if we can afford to lose previous data...

It's not a big deal if we keep it like this, `sendPoiEvent` is indeed a simplification.

src/panel/direction/DirectionPanel.jsx
bbecquet: It will be called the same number of times here, as component instances are re-instantiated before each mount, but I agree it would be cleaner. I'll move it.

local_modules/telemetry/index.js
amatissart: I agree it would be cleaner to merge this `*_osm_*` and `*_pages_jaunes_*` events.  
But I am afraid that should wait for another PR, and hopefully learn more about what these events will be used for in the future.

G-Ray (approved):

amatissart (approved):

PR #791, opened 2020-09-22
[Clean-up] Remove Masq code and assets

bbecquet:
## Description
Clean-up everything related to Masq. Please refer to individual commit titles for what is actually removed.

Compared to the check list exposed to the team some time ago, here is what is **not done** in this PR:
 - *cleaning Masq-related translation strings*. This is a bit laborious to do manually and I'm not sure there isn't a better and automatic way. If not, I'll add it.
 - *simplifying the store abstraction system*. This is a bit trickier than just deleting stuff, so I plan to do it separately. For now the abstraction layer is still here but with only the `localStorage` implementation behind. 

## Why
Less stuff to load for the user, less stuff to maintain for us.

In production build:

|Master|This branch|Diff|
|---|---|---|
|bundle.js (508 KiB)|bundle.js (474 KiB)| -6.69 %|
|public/css/app.css    108 KiB|public/css/app.css   99.3 KiB|-3.59 %|


package.json
```diff
@@ -54,7 +54,6 @@
     "json-loader": "0.5.7",
     "mapbox-gl": "^1.11.1",
     "mapbox-gl-js-mock": "https://github.com/QwantResearch/mapbox-gl-js-mock.git#7eaee7b",
-    "masq-lib": "git+https://github.com/QwantResearch/masq-lib.git#v0.14.3",
```
amatissart, line 57: `"detect-browser"` can be removed too. 
It was only used in `src/libs/masq.js` to disable Masq on unsupported browsers.

bbecquet: Thank you, it got past my radar indeed :)

G-Ray (approved):
Look at these stats :scream: -1,963 LoC

PR #792, opened 2020-09-22
Additional categories, used by intention detection only

amatissart:
## Description
Following https://github.com/QwantResearch/idunn/pull/166, this PR defines more categories to be used in "intentions" returned by the search results.

These new categories have no icon or color, as they don't appear as action button anywhere.  

Let me know if you find this `.filter()` mechanism too implicit. Maybe we could define 2 separate lists of categories with categories buttons on one hand, and searched-only categories on the other hand ?

### DIFF
--- a/src/components/CategoryList.jsx
+++ b/src/components/CategoryList.jsx
@@ -11,6 +11,7 @@ const handleCategoryClick = category => {
 const CategoryList = ({ className, limit = Number.MAX_VALUE }) =>
   <div className={className}>
     {CategoryService.getCategories()
+      .filter(c => c.iconName)
       .slice(0, limit)
       .map(category => <MainActionButton
         key={category.name}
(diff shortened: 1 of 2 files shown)


G-Ray:
I think the comment you wrote is explicit enough :)
Having all the categories defined in the same place, with different properties is the way to go I think.

G-Ray (approved):

src/components/CategoryList.jsx
```diff
@@ -11,6 +11,7 @@ const handleCategoryClick = category => {
 const CategoryList = ({ className, limit = Number.MAX_VALUE }) =>
   <div className={className}>
     {CategoryService.getCategories()
+      .filter(c => c.iconName)
```
bbecquet, line 14: :heart: for the code style but it may be useful to add a comment here or use a named function as filter argument.

bbecquet (dismissed):
Don't mind my previous comment, it's quite easy to understand this way :)

bbecquet (approved):

PR #795, opened 2020-09-23
Qmaps 1575 poi list redesign

xem:
## Description
Redesign PoI List for desktop & mobile, PJ & OSM.
- No more grey background
- No more cards
- PJ partnership message is fixed at the bottom of the screen (on mobile) / below the list (on desktop)
- If PJ partnership message is present, allow 36 extra pixels of scroll at the bottom of the items list
- If an item has no photo, use the category icon + a corresponding background color on a square with a border radius similar to the one in the photos (8px now)
- Hover an item on desktop: no more red left border

Notes about this PR:
- The loading of more PoI on scroll is not done in this branch.
- The re-shadowing of the top bar on scroll is not done in this branch.
- I decided to hide the PJ partnership message when there's no result.
- The Categories background colors are removed from the code and are now directly generated from the icons colors (take the RGB value of the icon color and add an Alpha = 0.157)
- I adjusted some little details when we move the panel between two sizes.

### Mobile

PJ:
image
image
image

image

OSM:

image
image
image
(ignore the "undefined" in the field, it'll be fixed soon on the API side)

### Desktop

PJ:

image

OSM:

image


src/components/CategoryList.jsx
```diff
@@ -13,8 +13,8 @@ const CategoryList = ({ className, limit = Number.MAX_VALUE }) =>
         label={category.label}
         icon={category.iconName}
         iconStyle={{
-          color: category.color,
-          backgroundColor: category.backgroundColor,
+          color: category.color, // RGB icon color
+          backgroundColor: category.color + '28', // RGBA background color equal to the icon color + 0.157 alpha
```
G-Ray, line 17: Great if this is the targetting color is the same ! I tried to do something similar to this once, without success

bbecquet: Nice and simple way to do that indeed :+1:

Now **let's move it to an util function**, so it's easier to name/comment/debug/test/make evolve. And re-use, as I see you use the same operation in this PR.

src/components/ui/Panel.jsx
```diff
@@ -202,6 +203,7 @@ class Panel extends React.Component {
             className={classnames('panel', size, className, {
               'panel--white': white,
               'panel--holding': holding,
+              'panel--pj': dataSource === 'pages_jaunes',
```
bbecquet, line 206: Let's think about separation of concerns. `Panel` is a generic component which doesn't need to know about its content. So it shouldn't complexified with stuff specific to a single feature.
As it's already possible to pass a custom `className` to it, you have a proper alternative which doesn't require you to modify `Panel` code for that.

src/panel/category/CategoryPanel.jsx
```diff
@@ -5,13 +5,13 @@ import Panel from 'src/components/ui/Panel';
...
 import PoiItemListPlaceholder from './PoiItemListPlaceholder';
 import CategoryPanelError from './CategoryPanelError';
-import CategoryPanelHeader from './CategoryPanelHeader';
 import Telemetry from 'src/libs/telemetry';
 import nconf from '@qwant/nconf-getter';
 import IdunnPoi from 'src/adapters/poi/idunn_poi';
 import { getVisibleBbox } from 'src/panel/layouts';
 import { fire, listen, unListen } from 'src/libs/customEvents';
 import { boundsFromFlatArray, parseBboxString, boundsToString } from 'src/libs/bounds';
+import { sources } from '../../../config/constants.yml';
```
bbecquet, line 14: To avoid these long chains of `../` and always having to know where you are in the tree, you can import from `'config/constants.yml'`, as there is a webpack alias on `config` like we use for `src`.

src/panel/category/CategoryPanel.jsx
```diff
@@ -159,11 +159,19 @@ export default class CategoryPanel extends React.Component {
 
     return <Panel
       resizable
-      renderHeader={<CategoryPanelHeader dataSource={dataSource} loading={initialLoading} />}
-      minimizedTitle={_('Show results', 'categories')}
+      renderHeader=""
```
bbecquet, line 162: Removing the prop should be enough.

src/scss/includes/panels/categories.scss
```diff
@@ -6,45 +6,40 @@
...
+
+  & ~ .category__panel__pj {
+    display: none;
+  }
 }
 
 @media (max-width: 640px) {
   .category__panel__pj {
-    height: 20px;
+    position: fixed;
```
bbecquet, line 35: It should be part of the panel content.
By moving it below instead of inside, you're making your life very complicated (having to change complex panel height computations and such, introducing a lot of new rules, etc.) and new bugs are appearing:

No border radius:

Capture d’écran de 2020-09-24 11-26-32


No shadow:

Capture d’écran de 2020-09-24 11-26-19

src/scss/includes/components/panel.scss
```diff
@@ -52,6 +52,11 @@ $bottom-margin: 40;
   .panel {
     max-height: calc(100vh - (92px + 40px)); // 92px is panel.getBoundingClientRect().top
 
+    &.panel--pj {
+      max-height: calc(100vh - (92px + 40px + 36px));
```
bbecquet, line 56: See my other comment about the position of the footer… At best you shouldn't have to change that.

src/scss/includes/components/panel.scss
```diff
@@ -68,7 +73,7 @@ $bottom-margin: 40;
     width: 100vw;
     height: 50%;
     position: absolute;
-    background: #f4f6fa;
+    background: #ffffff;
```
bbecquet, line 76: White or not is currently controlled by a prop which sets the `--white` variant on `.panel` (see some selectors below). So if all panels are white now, this means we should also remove all the code associated to that.

The best way here is just to set the `white` prop on `<CategoryPanel>`, and we do the cleaning in another PR. As to clean things up entirely we would have many files not related to CategoryPanel to change, and this would make this PR more confusing.

src/scss/includes/components/panel.scss
```diff
@@ -97,19 +102,30 @@ $bottom-margin: 40;
...
     .panel-handle {
       width: 40px;
       height: 5px;
-      margin: 4px auto;
+      margin: 8px auto;
       border-radius: 2.5px;
       background-color: #e0e1e6;
     }
+
+    &.minimized:not(.panel--holding) {
```
bbecquet, line 113: What is the impact of these global changes on other panels?

src/scss/includes/components/itemList.scss
```diff
@@ -1,42 +1,31 @@
...
-      position: absolute;
-      width: 4px;
-      height: 100%;
-      background-color: #FF3B4A;
-    }
   }
 }
 
 @media (max-width: 640px) {
+  .panel--pj .itemList {
```
bbecquet, line 14: Same idea as my comment on Panel. We are in the css file of a generic component, so you can think of it as an external lib, not editable. These specific rules should be in the only place they apply, meaning `categoryPanel`.

It's ok to target generic selectors in the specific context which use them, but the opposite isn't true (and I know some other files of the project break this rule, but we are making things better step by step :))

src/components/ui/Panel.jsx
G-Ray: Also, Panel has been refactored in master some time ago. We need to rebase this code on that.

src/scss/includes/components/panel.scss
G-Ray: The variable is rather `$background` right ?

src/scss/includes/panels/categories.scss
```diff
@@ -6,45 +6,40 @@
...
-    font-size: 12px;
-    color: $secondary_text;
-
-    .icon-icon_info {
-      font-size: 16px;
-      margin-right: 4px;
-    }
-  }
+  height: 36px;
+  position: absolute;
```
G-Ray, line 10: I don't think it need to be with absolute positioning ?

src/panel/category/CategoryPanel.jsx
xem: sorry, my IDE added this line automatically and I didn't catch it

src/components/ui/Panel.jsx
```diff
@@ -203,7 +203,6 @@ class Panel extends React.Component {
             className={classnames('panel', size, className, {
               'panel--white': white,
               'panel--holding': holding,
-              'panel--pj': dataSource === 'pages_jaunes',
```
bbecquet: You also have to remove `dataSource` in the propTypes declaration and 10 lines above when you get it from props.

src/panel/category/CategoryPanel.jsx
```diff
@@ -149,29 +150,24 @@ export default class CategoryPanel extends React.Component {
       if (hasError) {
         panelContent = <CategoryPanelError zoomIn={zoomIn} />;
       } else {
-        panelContent = <PoiItemList
+        panelContent =
+        <PoiItemList
           pois={pois}
           selectPoi={this.selectPoi}
           highlightMarker={this.highlightPoiMarker}
+          dataSource={dataSource}
```
bbecquet, line 158: It's a bit sad to have to change PoiItemList for this specific stuff. As a result you have to maintain it at 2 different levels in the tree. A best way would be to add the block here, in a `Fragment` with the `PoiItemList` element.

src/panel/category/CategoryPanel.jsx
```diff
@@ -149,19 +150,27 @@ export default class CategoryPanel extends React.Component {
...
     }
 
     return <Panel
+      white
       resizable
-      renderHeader={<CategoryPanelHeader dataSource={dataSource} loading={initialLoading} />}
-      minimizedTitle={_('Show results', 'categories')}
-      className="category__panel"
+      minimizedTitle={_('Unfold to show the results', 'categories')}
+      className={classnames('category__panel', { 'panel--pj': dataSource === 'pages_jaunes' })}
```
bbecquet, line 173: in the lines above, you use `dataSource === sources.pagesjaunes` for the same test. This one should be the same.

src/scss/includes/components/itemList.scss
```diff
@@ -1,42 +1,24 @@
...
-    }
   }
 }
 
 @media (max-width: 640px) {
+
   .itemList {
     padding-bottom: 12px;
-    margin: 0 15px;
+    margin: 0;
```
bbecquet, line 17: You should get the same result by just removing the `margin: 0` lines, as it's the default.

src/scss/includes/panels/categories.scss
```diff
@@ -6,47 +6,48 @@
...
 
 .category__panel__items {
   animation: appear 600ms forwards;
 }
 
 .category__panel__error {
-  padding: 16px 16px 32px 16px
+  padding: 16px 16px 32px 16px;
+
+  & ~ .category__panel__pj {
```
bbecquet, line 28: In the React component you chose not to render the block in case of error, so this rules seems unnecessary.

src/scss/includes/panels/categories.scss
```diff
@@ -6,47 +6,48 @@
...
+@media (min-width: 641px) {
+  .category__panel {
+
+    &.panel--pj {
+      .panel-content {
+        overflow: hidden;
+      }
+
+      .category__panel__items {
+        height: calc(100vh - 170px);
```
bbecquet, line 42: Here you reintroduce a hardcoded computation with some "magic" value, that we made disappeared for the panel content. Maybe it's needed, but I'm not sure. What was the problem with the flex-based layout?

xem: When I try to do it similarly to panel-header and panel-content, with the proper columns and shrink properties, the items container doesn't scroll anymore... 🤔

bbecquet: I tried the flex-based layout in the inspector, it worked:
```css
.category__panel.panel--pj .panel-content {
   overflow: hidden;
   display: flex;
   flex-direction: column;
}
.category__panel.panel--pj .category__panel__items {
   overflow-y: auto;
}
.category__panel__pj {
   /* position: absolute; to be removed */
   flex-shrink: 0;
}
```

G-Ray: yes IIRC we tested this

src/libs/colors.js
```diff
@@ -0,0 +1,4 @@
+// Compute a light rgba background color based on an icon's color
```
G-Ray, line 1: :+1:

src/scss/includes/panels/categories.scss
```diff
@@ -61,6 +62,25 @@
...
+
+  .category__panel__pj {
+    position: fixed;
+    bottom: 0;
+    height: 36px;
+    border-radius: 0;
+  }
+
+  .minimized:not(.panel--holding) {
+    height: 36px;
```
G-Ray, line 79: The minimized height should be correctly handled by th Panel itsel now

src/scss/includes/panels/categories.scss
xem: thanks for the tip! I was close, now it works :)

G-Ray (approved):

PR #797, opened 2020-09-24
Add label styling to POI marker layer

bbecquet:
## Description
Add labels on category POI map markers. 
Also implement their selected state (red text instead of black).

Capture d’écran de 2020-09-28 11-48-23

I used [text-variable-anchor](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-text-variable-anchor) so the label is allowed to be moved on the sides of the icon to fit without collisions, but there are still some overlapping with other icons :/ Mapbox doesn't always chose to move the label where it could possibly avoid things. After many tries I think it's related to Mapbox implementation and we can't obtain much better for now.
I suggest we move on with this state, as it's still globally ok and an improvement over no label at all.

### DIFF
--- a/src/adapters/pois_styles.js
+++ b/src/adapters/pois_styles.js
@@ -14,13 +14,14 @@ export const filteredPoisStyle = {
     'text-allow-overlap': false,
     'text-ignore-placement': false,
     'text-optional': true,
-    'text-anchor': 'top',
-    'text-radial-offset': 0.25,
+    'text-variable-anchor': ['top', 'bottom-left', 'bottom-right'],
+    'text-offset': [1.5, 0.5],
+    'text-justify': 'auto',
   },
   paint: {
     'text-color': ['case', ['==', ['feature-state', 'selected'], true], '#900014', '#0c0c0e'],
     'text-halo-color': 'white',
     'text-halo-width': 1,
-    'text-translate': [0, 4],
+    'text-translate': [0, -2],
   },
 };
--- a/src/adapters/pois_styles.js
+++ b/src/adapters/pois_styles.js
@@ -7,5 +7,20 @@ export const filteredPoisStyle = {
     'icon-ignore-placement': true,
     'icon-size': 0.5,
     'icon-anchor': 'bottom',
+
+    'text-font': [ 'Noto Sans Bold' ],
+    'text-size': 10,
+    'text-field': '{name}',
+    'text-allow-overlap': false,
+    'text-ignore-placement': false,
+    'text-optional': true,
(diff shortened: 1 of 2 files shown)


amatissart:
Hum, indeed the limitations of mapbox-gl seem to be quite strict here...

My contribution with very little improvements:
```diff
diff --git a/src/adapters/pois_styles.js b/src/adapters/pois_styles.js
index da1dcf85..1c26f642 100644
--- a/src/adapters/pois_styles.js
+++ b/src/adapters/pois_styles.js
@@ -14,13 +14,13 @@ export const filteredPoisStyle = {
     'text-allow-overlap': false,
     'text-ignore-placement': false,
     'text-optional': true,
-    'text-anchor': 'top',
+    'text-variable-anchor': ['top','top-left','top-right'],
     'text-radial-offset': 0.25,
   },
   paint: {
     'text-color': ['case', ['==', ['feature-state', 'selected'], true], '#900014', '#0c0c0e'],
     'text-halo-color': 'white',
     'text-halo-width': 1,
-    'text-translate': [0, 4],
+    'text-translate': [0, 2],
   },
 };
```

bbecquet:
@amatissart I used `bottom-left`/`bottom-right` and adjusted offsets and translations, I got this result. This isn't so bad in terms of left/right positioning, but too bad about colisions with the icons themselves :/

Capture d’écran de 2020-09-25 13-52-45

G-Ray:
The labels are great looking ! Collisions are not perfect indeed.

src/adapters/pois_styles.js
```diff
@@ -7,5 +7,21 @@ export const filteredPoisStyle = {
...
+    'text-field': '{name}',
+    'text-allow-overlap': false,
+    'text-ignore-placement': false,
+    'text-optional': true,
+    'text-variable-anchor': ['top', 'bottom-left', 'bottom-right'],
+    'text-offset': [1.5, 0.5],
+    'text-justify': 'auto',
+  },
+  paint: {
+    'text-color': ['case', ['==', ['feature-state', 'selected'], true], '#900014', '#0c0c0e'],
```
G-Ray, line 22: We should share colors variables in JS too. It's ok for this PR, we need to move forward.

G-Ray: Would it be possible to display markers above labels in all cases to workaround the issue ? Lika an order priority. I don't know if it would be ugly.

G-Ray (approved):

amatissart (approved):

PR #798, opened 2020-09-24
Use lang without locale in .po filenames for consistency with Transifex configuration

amatissart:
## Description
Supported locales are listed in `config/constants.yml`.  
In practice the distinction between languages (`en`, `de`, `ca`, etc.) and locales (`en_GB`, `de_DE`, `ca_ES`, etc.) is only used for datetime formatting purposes, and has no impact on translated messages for now. 

This PR suggests to define a single `.po` file per language when all locales use the same messages for this language.
In this case, the `<locale>.po` is now defined as a symbolic link to the corresponding `<lang>.po` file. 
For example, `fr_FR.po` is now a symbolic link to `fr.po`. 

## Why
We are using Transifex to track and update messages to be localized, and allow external translators to integrate their contributions.
In Transifex, translators are sometimes restricted to a list of languages where no locale is specified. So this will make possible to integrate their translations into the project, while preserving existing messages. Moreover, this change will also be more consistent with other Qwant projects that are using Transifex.


bbecquet (approved):

G-Ray (approved):
:+1:

PR #800, opened 2020-09-25
Fix long touch on map interaction on Firefox mobile

bbecquet:
## Description
After the merge of https://github.com/QwantResearch/erdapfel/pull/785, we discovered a bug on Firefox mobile: after correctly opening a "POI anywhere" on a "long touch", the panel was immediately closed as if the user had clicked on an empty part of the map. This didn't occured for very long touches.

The root cause seems to be Firefox has a longer delay (> 500 ms) than other browsers when he still considers a prolonged touch event a click. So, even after our timeout expired and we triggered the action of a long touch (opening the POI anywhere panel), if the user releases their finger, the browser fires a `touchend` event and then a `click` event as the synthesis of this action. We interpret this `click` event like a single tap on an empty part of the map, which closes the panel.
When the user keeps their finger longer on the screen, `touchcancel` is fired instead of `touchend` and this isn't considered a `click` by Firefox, explaining why it worked.
On other browsers, no `click` event is fired at all when the touch delay is > 500ms.

This is a fix proposition. We explicitely set a flag to exit early on the next `click` event received after considering a long touch.
To be sure it doesn't apply on the next events or when not needed on other browsers, we reset the flag when we know we won't consider the current interaction as a long touch (the timeout is cancelled early).

(What a mess…)

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -122,8 +122,15 @@ Scene.prototype.initMapBox = function() {
     // we have to delay click event resolution to make time for possible double click events,
     // which are thrown *after* two separate click events are thrown
     this.clickDelayHandler = null;
+    // switch used to prevent 'click' events to have an effect
+    // after a touch interaction has been considered a longtouch
+    this.cancelClickAfterLongTouch = false;
     this.mb.on('click', e => {
       if (e.originalEvent.cancelBubble) {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


xem (approved):
Nice catch!
It's sad we have to add this fix but I think it's not too messy.

bbecquet:
This alternative version is cleaner https://github.com/QwantResearch/erdapfel/pull/802

PR #803, opened 2020-09-28
Remove symlinks in language/messages for each supported locale

amatissart:
## Description
The language code (`en`, `fr`, `br` etc.) is used to fetch the .po file corresponding to each locale and its potential fallback list of languages.

## Why
Symbolic links seem to be poorly supported on Windows :crying_cat_face:, and we do not plan to use different messages for locales with a single language code.


bbecquet (approved):

xem (approved):
works for me! shouldn't we also rename xx_XX.js into xx.js?

PR #806, opened 2020-09-28
Fix mobile map buttons position

G-Ray:
## Description
Compute mobile buttons position according to panel's height from bottom.

## Why
I missed this in the PR about panel


bbecquet (approved):
Nice, and the code gets cleaner too :+1:

PR #811, opened 2020-10-01
Panel: white as only background color

bbecquet:
## Description
Set white as the new default (and only) color for the Panel component.
Previously it was gray for mobile panels, with white as an explicit prop. So we can simplify the API of Panel with one less prop.

## Why
We don't have gray background designs any more.
Following https://github.com/QwantResearch/erdapfel/pull/795, the generic item list style changed so items are now full-width, no-border, white background. This doesn't fit anymore with gray backgrounds, so favorites and route results seem broken, with a gray header. They will soon be explicitely redesigned, but until now let's fix them at least.

## Screenshots
|Before|After|
|---|---|
|localhost_3000_routes__origin=latlon_48 87674_2 33186 destination=latlon_48 88692_2 35252 mode=driving pt=true(iPhone 6_7_8) (2)|localhost_3000_favs_(iPhone 6_7_8)|
|localhost_3000_routes__origin=latlon_48 87674_2 33186 destination=latlon_48 88692_2 35252 mode=driving pt=true(iPhone 6_7_8) (1)|localhost_3000_routes__origin=latlon_48 87674_2 33186 destination=latlon_48 88692_2 35252 mode=driving pt=true(iPhone 6_7_8)|

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -49,7 +49,6 @@ class Panel extends React.Component {
     setSize: PropTypes.func,
     marginTop: PropTypes.number,
     className: PropTypes.string,
-    white: PropTypes.bool,
   }
 
   static defaultProps = {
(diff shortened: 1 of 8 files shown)


G-Ray (approved):
:+1:

PR #813, opened 2020-10-01
Fix direction panel results

bbecquet:
## Description
Fixes two separate problems touching the direction panel results:
 - on desktop, the result were not scrollable anymore, due to changes in how the panel content is built with flex stacking instead of computing inner content size.
 - on mobile, the route details view was displayed inside the panel, instead of fullscreen, because of the problem of fixed-positioned element in translated ones. => fixed with a React portal


G-Ray (approved):
:+1:

PR #814, opened 2020-10-01
Direction form, white background with gradient

G-Ray:
## Description
- Remove backgrounds
- Define one `direction-panel` class for mobile & desktop
- gradient around direction form on focus

## Why
Implement new design

## Screenshots
|state|before|after|
|---|---|---|
|unfocus|image|image|
|focus|image|image|


src/scss/includes/direction.scss
```diff
@@ -1,6 +1,32 @@
-.direction_panel {
-  background: url("../images/panel_top.svg") #f4f6fa repeat-x !important;
-  background-size: 100% 5px;
+.direction-panel {
+  background-color: $background;
```
bbecquet, line 2: These 4 rules aren't required at all on desktop now that you reverted to a Panel.
And on mobile, only the `background-color` and shadow are required.

src/scss/includes/direction.scss
```diff
@@ -1,6 +1,32 @@
...
+  @include card_shadow();
+  width: 400px;
+
+  position: absolute;
+
+  @media (min-width: 641px) {
```
bbecquet, line 8: This can be removed too.

src/scss/includes/direction.scss
```diff
@@ -1,6 +1,32 @@
...
+    @include panel_radius();
+  }
+
+  @media (max-width: 640px) {
+    position: fixed;
+    min-height: 42px;
```
bbecquet, line 14: I don't think this is needed (but maybe I'm wrong, the style is tricky during the input focus).

src/scss/includes/direction.scss
```diff
@@ -1,6 +1,32 @@
...
+    width: 100%;
+    top: 0;
+    left: 0;
+    z-index: 3;
+
+    .itinerary_title {
```
bbecquet, line 20: This can be removed as the same exact rules are defined below outside of the media query :)

src/scss/includes/direction.scss
```diff
@@ -9,6 +35,10 @@
   }
 }
 
+.panel-content {
```
bbecquet, line 38: This is applied on the global scope. If that's on purpose, it should be defined in the panel stylesheet. Otherwise, it should be scoped with a parent selector.

src/scss/includes/colors.scss
```diff
@@ -36,3 +36,5 @@ $action-blue-lighter: #e5eeff;
 $red-base: #ff1d3c;
 $green-base: #83c458;
 $orange-base: #ff851e;
+
+$gradient: linear-gradient(to right, #216fff, #af27cc);
```
bbecquet, line 40: Let's give it a more specific name. What about `$focus-gradient`?

bbecquet (commented):
Nice simplification. Just some little clean-up suggestions, that should make the result even more simple.

src/scss/includes/direction.scss
G-Ray: some style are needed for mobile through, so i moved them

src/scss/includes/direction.scss
bbecquet: This selector can still be removed.

src/scss/includes/direction.scss
```diff
@@ -977,9 +975,19 @@ button.direction_shortcut {
...
+
+    &::before {
+      content: '';
+      position: fixed;
+      top: 0; right: 0; bottom: 0; left: 0;
+      margin: -3px;
+      border-radius: inherit;
+      z-index: 1;
+      background: $focus-gradient;
+    }
```
amatissart, line 990: @G-Ray Did you mean to commit this change ? The gradient is not visible on mobile on my side.

G-Ray: I did not want to, sorry I remove it rn

bbecquet (approved):

amatissart (commented):
Very minor: 
Peek 02-10-2020 17-37

The destination icon is visible on focus during a brief moment.

PR #815, opened 2020-10-02
Factorize panel close button and fix interaction on PoiPanel

bbecquet:
## Description
 - Define a new generic component `CloseButton`, that encapsulates its own style for proper size, alignment, hover, etc. Can be extended later.
 - Use this component in PoiPanel, improves design conformity and reduces style specifity.
 - Use this component in Panel, when using the `onClose` prop
 - Also reduce the width of mobile Panel clickable area for auto-resize, so it isn't cliked by mistake when trying to close the panel

## Screenshot
|Before|After|
|---|---|
|localhost_3000_place_pj_05353141@Kozo_Makita(iPhone 6_7_8)|localhost_3000_favs(iPhone 6_7_8) (1)|
|localhost_3000_favs_(iPhone 6_7_8) (4)|localhost_3000_favs_(iPhone 6_7_8) (3)|

### DIFF
--- a/src/components/ui/CloseButton.jsx
+++ b/src/components/ui/CloseButton.jsx
@@ -1,0 +1,14 @@
+/* global _ */
+import React from 'react';
+import cx from 'classnames';
+
+const CloseButton = ({ className, onClick }) =>
+  <button
+    className={cx('closeButton', className)}
+    title={_('Close')}
+    onClick={onClick}
+  >
+    <i className="icon-x" />
+  </button>;
+
+export default CloseButton;
--- a/src/components/ui/CloseButton.jsx
+++ b/src/components/ui/CloseButton.jsx
@@ -2,9 +2,9 @@
 import React from 'react';
 import cx from 'classnames';
 
-const CloseButton = ({ className, onClick }) =>
+const CloseButton = ({ className, position, onClick }) =>
   <button
-    className={cx('closeButton', className)}
+    className={cx('closeButton', { [`closeButton--${position}`]: position }, className)}
     title={_('Close')}
     onClick={onClick}
   >
--- a/src/components/ui/CloseButton.jsx
+++ b/src/components/ui/CloseButton.jsx
@@ -1,6 +1,7 @@
 /* global _ */
 import React from 'react';
 import cx from 'classnames';
+import PropTypes from 'prop-types';
 
 const CloseButton = ({ className, position, onClick }) =>
   <button
@@ -10,5 +11,11 @@ const CloseButton = ({ className, position, onClick }) =>
   >
     <i className="icon-x" />
   </button>;
+
+CloseButton.propTypes = {
+  className: PropTypes.string,
+  position: PropTypes.oneOf(['topRight']),
+  onClick: PropTypes.func.isRequired,
+};
 
 export default CloseButton;
--- a/src/scss/includes/components/closeButton.scss
+++ b/src/scss/includes/components/closeButton.scss
@@ -6,9 +6,12 @@
   justify-content: center;
   font-size: 22px;
   cursor: pointer;
-  margin-top: -9px;
-  margin-right: -9px;
   color: $primary_text;
+
+  &--topRight {
+    margin-top: -9px;
+    margin-right: -9px;
+  }
 
   @media (min-width: 641px) {
     color: $secondary_text;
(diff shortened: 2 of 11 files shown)


src/components/ui/CloseButton.jsx
```diff
@@ -0,0 +1,14 @@
+/* global _ */
+import React from 'react';
+import cx from 'classnames';
+
+const CloseButton = ({ className, onClick }) =>
```
G-Ray, line 5: :+1: If we need to handle multiple buttons with only an icon later, I think we would add the relevant behavior into `Button.jsx`

src/scss/includes/components/closeButton.scss
```diff
@@ -0,0 +1,21 @@
+.closeButton {
+  height: 35px;
+  width: 35px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-size: 22px;
+  cursor: pointer;
+  position: relative;
+  top: -9px;
```
G-Ray, line 10: I think it's best if CloseButton totally ignores where it will be positioned

bbecquet: For me it's not really positionning stuff but a shift to ensure the cross is visually aligned with the surrounding content but the actual clickable size is larger. So this is better if it's not managed by the calling sites.
But I understand your point. I'll try doing it with margins instead.

src/components/ui/CloseButton.jsx
bbecquet: `<Button>` already supports the case of icon-only buttons (like on POI actions), but with a standard bordered style and no need to 'cheat' to expand the visible area. I think the case of close button is specific-enough to always be managed separately.

G-Ray: as long as it is specific to close button yes

src/scss/includes/components/closeButton.scss
```diff
@@ -0,0 +1,20 @@
+.closeButton {
+  height: 35px;
+  width: 35px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-size: 22px;
+  cursor: pointer;
+  margin-top: -9px;
```
G-Ray, line 9: why a negative margin is needed here ? Shouldn't setting w/h dimensions be enough ?

bbecquet: That's what I explained here https://github.com/QwantResearch/erdapfel/pull/815#discussion_r498821104 and referred to also there https://github.com/QwantResearch/erdapfel/pull/815#discussion_r498833356.
Capture d’écran de 2020-10-05 13-52-59

We need to 'cheat' so that the button content (the cross) is aligned to the content, but the actual active area should be larger than that. I think it's best if this is the responbility of the button itself, as it's dependant of its size, and part of what justified to make a separate component.
But I understand if it bothers you that it's directly in the class. Maybe we can put it a variant like `.closeButton--topRight' and map a prop to it?

G-Ray: I missed your previous comment has it had been marked as resolved, sorry.
The button would be reusable anywhere if we can pass props to specify margins :+1:

bbecquet: Just to be sure we're on the same line: I don't want to pass to margin values themselves, as they depend on the button style itself. I want to pass props which say that the button is used on the right/top of a container, so it can be shifted.
Like:
`<CloseButton onClose={foo} position="topRight" />`
and for CSS:
```scss
.closeButton {
  // general style

  &--topRight {
    margin-top: -9px;
    margin-right: -9px;
  }
}
```

G-Ray: yep it's ok :+1:

src/components/ui/CloseButton.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+const CloseButton = ({ className, position, onClick }) =>
+  <button
+    className={cx('closeButton', { [`closeButton--${position}`]: position }, className)}
+    title={_('Close')}
+    onClick={onClick}
+  >
+    <i className="icon-x" />
+  </button>;
+
```
G-Ray, line 13: Can we add propTypes ? (mainly for position)

G-Ray (approved):

PR #819, opened 2020-10-05
regroup scss properties in nested classes

G-Ray:
## Description
- Regroup scss properties with nested classes
- Split styles into 3 files : panel, form and input.

## Why
`direction.scss` is harder and harder to read/edit.


bbecquet (approved):
Thanks for the renaming effort!
For the structural refactoring, I don't find it much more readable this way (except for the split), but I think we all agree it's just a first step and real simplifications will comme later. In the meantime it's great if you better find your way in it :+1:

PR #820, opened 2020-10-06
Redesign vehicle selector

bbecquet:
## Description
Better look for vehicle selector, according to the latest designs.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran 2020-10-06 à 11 01 03|Capture d’écran 2020-10-06 à 11 01 21|
|Capture d’écran de 2020-10-06 10-58-39|Capture d’écran de 2020-10-06 10-54-49|

### DIFF
--- a/src/scss/includes/vehicleSelector.scss
+++ b/src/scss/includes/vehicleSelector.scss
@@ -69,7 +69,7 @@
         content:'';
         width: 100%;
         height: 4px;
-        background: $focus-gradient;
+        background: $active-gradient;
         border-radius: 4px 4px 0 0;
       }
     }
--- a/src/scss/includes/vehicleSelector.scss
+++ b/src/scss/includes/vehicleSelector.scss
@@ -1,11 +1,11 @@
-.itinerary_vehicles {
+.vehicleSelector {
   display: flex;
   justify-content: center;
   margin: 0px 32px;
   padding: 0px 32px 16px;
 }
 
-.itinerary_vehicle_button {
+.vehicleSelector-button {
   display: flex;
   align-items: center;
   justify-content: center;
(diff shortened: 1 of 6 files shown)


src/scss/includes/vehicleSelector.scss
```diff
@@ -0,0 +1,77 @@
...
+    &--active {
+      .vehicleSelector-buttonLabel {
+        color: $primary_text;
+      }
+
+      &:after{
+        content:'';
+        width: 100%;
+        height: 4px;
+        background: $focus-gradient;
```
G-Ray, line 72: according to mockups, it's a blue gradient. Can we define it in `colors.scss`, like `focus-gradient` ?

bbecquet: Yeah…

bbecquet: Done :)

G-Ray:
nice,

the only thing is we got the same hover issue as me on touch devices:
Peek 2020-10-06 16-23

On mobile devices, we can apply style only when element `active` for instance.

bbecquet:
@G-Ray I don't define any `hover` on mobile. Your screenshot is on the desktop design.

G-Ray:
> @G-Ray I don't define any `hover` on mobile. Your screenshot is on the desktop design.

my bad you r right...

Maybe we should review our hover effect for other touch devices (tablet devices) at one point.

bbecquet:
@G-Ray The proper way is probably to use a feature detection media query.
https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#Targeting_media_features

G-Ray (approved):

xem (approved):

PR #821, opened 2020-10-06
Direction: rework reverse button

G-Ray:
## Screenshots
image

### DIFF
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -84,11 +84,12 @@ export default class DirectionForm extends React.Component {
           onChangePoint={(input, point) => this.onChangePoint('destination', input, point)}
           ref={this.destinationRef}
         />
-        <div
-          className="direction-invert-origin-destination icon-reverse"
+        <button
+          className="direction-invert-origin-destination"
           onClick={this.onReverse}
-          title={_('Invert start and end', 'direction')}
-        />
+          title={_('Invert start and end', 'direction')}>
+          <i className="icon-reverse" />
+        </button>
       </form>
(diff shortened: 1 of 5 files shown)


src/scss/includes/direction-form.scss
```diff
@@ -32,19 +32,23 @@
 
     .direction-form-separator {
       border-top: 1px solid #c8cbd3;
-      margin: 0 37px;
+      margin: 0 52px 0 44px;
```
G-Ray, line 35: we might try to create a flex layout for the form elements later

bbecquet: Isn't it possible right now?

G-Ray: I would like to push this change and the font changes before, just to have the general visual elements in place before changing layout/DOM :)

G-Ray: I just added minor font changes

src/panel/direction/DirectionForm.jsx
```diff
@@ -84,11 +84,12 @@ export default class DirectionForm extends React.Component {
           onChangePoint={(input, point) => this.onChangePoint('destination', input, point)}
           ref={this.destinationRef}
         />
-        <div
-          className="direction-invert-origin-destination icon-reverse"
+        <button
```
bbecquet, line 87: Great to use a real button, but in this case you need to add a `type="button"` attribute, as the default type of a HTML button is `submit`, which will try to submit the form it belongs too… which will open a new page :D

bbecquet (dismissed):

bbecquet (approved):

PR #822, opened 2020-10-06
QMAPS-1752 suggest

xem:
- make suggest list full-height on mobile (top bar +direction panel)
- show more text in suggest list items (only 60px of space on the left for the icon)
- direction panel: hide "your position" when we started typing, and no border below "your position" when it's visible.

## Screenshots

image
image

image
image

### DIFF
--- a/src/scss/includes/direction.scss
+++ b/src/scss/includes/direction.scss
@@ -742,6 +742,6 @@ button.direction_shortcut {
   }
 
   .autocomplete_suggestions {
-    height: calc(100vh - 64px);
+    height: 100vh;
   }
 }
--- a/src/scss/includes/direction.scss
+++ b/src/scss/includes/direction.scss
@@ -740,4 +740,8 @@ button.direction_shortcut {
       }
     }
   }
+
+  .autocomplete_suggestions {
+    height: calc(100vh - 64px);
+  }
 }
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -154,11 +154,9 @@ input[type="search"] {
   width: 60px;
   flex-shrink: 0;
   text-align: center;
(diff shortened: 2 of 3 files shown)


src/scss/includes/direction.scss
```diff
@@ -740,4 +740,8 @@ button.direction_shortcut {
       }
     }
   }
+
+  .autocomplete_suggestions {
```
G-Ray, line 744: I see no difference by commenting this rule ?

xem: when it's commented, the directions suggest list on mobile doesn't take all the height of the screen as required:

image

image

src/scss/includes/search_form.scss
```diff
@@ -154,11 +154,9 @@ input[type="search"] {
   width: 60px;
   flex-shrink: 0;
   text-align: center;
-  margin-right: 14px;
```
bbecquet, line 157: This rule is here so (on desktop at least), the suggestion labels are vertically aligned with the input content. Let's wait for a more general scope redesign of all suggests to change it.

bbecquet: Your screenshots show exactly what I describe:
|Before|After|
|---|---|
|95210149-7b2a2280-07eb-11eb-86d7-6d0a4129b7fc|95210298-a3b21c80-07eb-11eb-8c02-ac32363304a3|

And yes, this alignment is on purpose :smile: 

My point is you are changing something outside of the scope of directions only, so we should manage it we care, and there is no rush to do it here.

xem: Thomas just validated this change even if it applies to both Directions and Top bar.
Let's merge it like that and wait for a future redesign of this panel 👍

bbecquet (approved):

PR #823, opened 2020-10-06
Direction: Use flexbox to position form content

G-Ray:


src/scss/includes/direction-form.scss
```diff
@@ -46,20 +48,18 @@
...
+    .direction-invert-button {
       display: flex;
       align-items: center;
       justify-content: center;
-      width: 36px;
-      height: 36px;
-      right: 8px;
-      top: 22px;
-      cursor: pointer;
+      min-width: 36px;
```
bbecquet, line 55: Why these `min-*`? It looks like it's allowed grow, whereas it shouldn't. Maybe what you want is `flex-grow: 0` and `flex-shrink: 1`?

PR #824, opened 2020-10-07
Qmaps 1748/1751 - Direction panel tweaks

G-Ray:
## Description
- Update title and placeholders
- add closeButton on mobile
- add focus icons states
- fix direction's form overflowing along with suggestions, using `renderHeader` props.
- reverse button styles is now applied depending of its disabled state, instead of hover => Button component now accept a disabled prop

## Why
New Design

## Screenshots
desktop
mobile

Icon color change on focus. The colored icons are now displayed only on `valid` state (once validated).
image


src/panel/direction/DirectionForm.jsx
```diff
@@ -89,6 +89,7 @@ export default class DirectionForm extends React.Component {
         </div>
 
         <button
+          disabled={originInputText === '' && destinationInputText === ''}
```
bbecquet, line 92: Great for the disabled! It's far more consistent like this.
And now it really looks like it should be a standard <Button> component :) It will just need to add `disabled` as a prop to it.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -310,13 +313,8 @@ export default class DirectionPanel extends React.Component {
         : <Panel
           className="direction-panel"
           onClose={this.onClose}
-          renderHeader={(
-            <div>
-              {title}
-            </div>
-          )}
+          renderHeader={form}
```
bbecquet, line 316: Isn't it a bit dangerous to use the header for that?

src/panel/direction/DirectionPanel.jsx
```diff
@@ -310,13 +313,8 @@ export default class DirectionPanel extends React.Component {
         : <Panel
           className="direction-panel"
           onClose={this.onClose}
```
bbecquet, line 315: The onClose doesn't work because the CloseButton managed by Panel is covered by the form element :/

src/scss/includes/utils.scss
```diff
@@ -54,6 +54,14 @@
   font-weight: normal;
 }
 
+.u-text--title {
```
bbecquet, line 57: I know it's the design but I'm a bit worried this will be too big for the string to fit in some languages…

bbecquet (commented):
Another comment about something not in the diff, but probably should be:
The horizontal space for the origin/destination field icons is too large. The code says 64px, but the design says 48px. 
Capture d’écran de 2020-10-08 10-18-50

src/panel/direction/DirectionPanel.jsx
bbecquet: I would just put that in the children with the others, define no renderHeader at all, and set a `z-index: 1` to the Panel CloseButton.

G-Ray:
> Another comment about something not in the diff, but probably should be:
> The horizontal space for the origin/destination field icons is too large. The code says 64px, but the design says 48px.
> Capture d’écran de 2020-10-08 10-18-50

Actually, it's 64px on desktop, et 48px on mobile... I did not see it

src/panel/direction/DirectionPanel.jsx
G-Ray: what do you mean by "dangerous" ? This is needed to separate the real, **scrollable** panel content, and its header which is really just a children.

src/panel/direction/DirectionPanel.jsx
G-Ray: I set a z-index value to 1 on closeButton, is it ok for you ?

src/scss/includes/utils.scss
G-Ray: let's see how it goes ?

src/panel/direction/DirectionForm.jsx
G-Ray: Disabled props should now be properly handled by the Button component :record_button:

src/components/ui/Button.jsx
```diff
@@ -4,13 +4,19 @@ import PropTypes from 'prop-types';
...
   const Tag = href ? 'a' : 'button';
 
   return <Tag
     type={type}
-    className={classnames('button', { [`button--${variant}`]: variant }, className)}
+    disabled={disabled}
+    className={classnames(
+      'button',
+      { [`button--${variant}`]: variant },
+      { ['button--disabled']: disabled },
```
bbecquet, line 17: We don't need a variant for that. You can target with the `:disabled` CSS selector.

bbecquet (dismissed):

bbecquet (approved):

PR #826, opened 2020-10-09
Tweak category list padding and footer

bbecquet:
## Description
Adjust paddings and PagesJaunes footer on the desktop category list, to match the design.

Let's focus on this panel only here, but it would be good too to soon make a pass over all our panels to ensure consistency of paddings. Introducing the spacing vars used by the design would be useful.

## Screenshots
|Before|After|
|---|---|
|localhost_3000_places__type=restaurant (1)|localhost_3000_places__type=restaurant|

### DIFF
--- a/src/scss/includes/panels/categories.scss
+++ b/src/scss/includes/panels/categories.scss
@@ -5,6 +5,7 @@
 }
 
 .category__panel__pj {
+  width: 100%;
   height: 32px;
   border-top: 1px solid $grey-light;
   line-height: 32px;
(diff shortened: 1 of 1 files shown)


G-Ray:
I agree, also I guess padding could be set only by `Panel.jsx` directly.

G-Ray:
image
divider does not take all width on mobile

bbecquet:
@G-Ray I fixed the footer width on mobile. Was too quick to remove the `width: 100%` :)

G-Ray (approved):

PR #828, opened 2020-10-09
rewrite restorePoints with async/await

G-Ray:


src/panel/direction/DirectionPanel.jsx
```diff
@@ -16,6 +16,15 @@ import { fire, listen, unListen } from 'src/libs/customEvents';
 import * as address from '../../libs/address';
 import { CloseButton, Flex } from '../../components/ui';
 
+const setMarker = (type, poi) => {
+  window.execOnMapLoaded(() => {
+    fire(`set_${type}`, poi);
```
amatissart, line 21: I am not very fond of using a dynamic value here. It would make future refactoring particularly harder if we can't even `grep` for the event name. Maybe we could merge `set_destination` and `set_origin` into a single event by adding another parameter.

bbecquet (approved):

PR #829, opened 2020-10-09
Tweak mobile direction field when focused

bbecquet:
## Description
Just some hardcoded tweaks so that the focused mobile direction fields look more alike the general search field.
The right way to solve the inconsistencies would be to use the same markup/CSS for both (with some flex <3), but this is more work (especially if we want the same React component), so let's keep it for later and fix only the look for now.
 - adjust height and padding
 - adjust size and positions of the back icon

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2020-10-09 17-00-51|Capture d’écran de 2020-10-09 16-58-28|

### DIFF
--- a/src/scss/includes/direction-field.scss
+++ b/src/scss/includes/direction-field.scss
@@ -18,11 +18,11 @@
         position: fixed;
         display: block;
         z-index: 1;
-        top: 5px;
+        top: 8px;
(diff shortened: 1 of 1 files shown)


xem (approved):
Great

PR #830, opened 2020-10-09
Qmaps 1751 - Set current location as origin on mobile

G-Ray:
## Description

- On mobile devices, if origin is not set as a props to `DirectionPanel` on mount, set current location as origin.
- geolocation: export functions instead of class
- asyncify checkPrompt
- checkPrompt -> showGeolocationModalIfNeeded

## Screenshots

Screencast when opening a default direction panel, and when opening an itinerary from a POI panel :
Peek 2020-10-12 10-48

### DIFF
--- a/src/adapters/poi/specials/navigator_geolocalisation_poi.js
+++ b/src/adapters/poi/specials/navigator_geolocalisation_poi.js
@@ -1,7 +1,7 @@
 /* global _ */
 
 import Poi from '../poi';
-import GeolocationCheck from 'src/libs/geolocation';
+import * as GeolocationCheck from 'src/libs/geolocation';
 
 export const navigatorGeolocationStatus = {
   PENDING: 'pending',
--- a/src/adapters/poi/specials/navigator_geolocalisation_poi.js
+++ b/src/adapters/poi/specials/navigator_geolocalisation_poi.js
@@ -24,7 +24,7 @@ export default class NavigatorGeolocalisationPoi extends Poi {
   }
 
   async geolocate(options = { displayErrorModal: true }) {
-    await GeolocationCheck.checkPrompt();
+    await GeolocationCheck.showGeolocationModalIfNeeded();
     return new Promise((resolve, reject) => {
       this.status = navigatorGeolocationStatus.PENDING;
       navigator.geolocation.getCurrentPosition(position => {
--- a/src/adapters/poi/specials/navigator_geolocalisation_poi.js
+++ b/src/adapters/poi/specials/navigator_geolocalisation_poi.js
@@ -23,7 +23,7 @@ export default class NavigatorGeolocalisationPoi extends Poi {
     return window.__navigatorGeolocalisationPoi;
   }
 
-  async geolocate() {
+  async geolocate(options = { displayErrorModal: true }) {
     await GeolocationCheck.checkPrompt();
     return new Promise((resolve, reject) => {
       this.status = navigatorGeolocationStatus.PENDING;
@@ -34,7 +34,11 @@ export default class NavigatorGeolocalisationPoi extends Poi {
         if (error.code === 1) {
           this.status = navigatorGeolocationStatus.FORBIDDEN;
         }
-        GeolocationCheck.handleError(error);
+
+        if (options.displayErrorModal) {
+          GeolocationCheck.handleError(error);
+        }
+
         reject(error);
       });
     });
--- a/src/adapters/poi/specials/navigator_geolocalisation_poi.js
+++ b/src/adapters/poi/specials/navigator_geolocalisation_poi.js
@@ -1,7 +1,7 @@
 /* global _ */
 
 import Poi from '../poi';
-import * as GeolocationCheck from 'src/libs/geolocation';
+import * as Geolocation from 'src/libs/geolocation';
 
 export const navigatorGeolocationStatus = {
   PENDING: 'pending',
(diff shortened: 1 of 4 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -56,11 +59,24 @@ export default class DirectionPanel extends React.Component {
...
-  componentDidMount() {
+  async componentDidMount() {
     Telemetry.add(Telemetry.ITINERARY_OPEN);
     document.body.classList.add('directions-open');
     this.dragPointHandler = listen('change_direction_point', this.changeDirectionPoint);
     this.setPointHandler = listen('set_direction_point', this.setDirectionPoint);
+
+    if (!this.props.origin && isMobileDevice()) {
+      // If authorized, set origin to current position on mobile
+      const isGeolocAvailable = await new Promise(resolve =>
```
bbecquet, line 70: It seems this is not blocking, as the field gets filled with "My position" before the user has actually accepted to enable it. As a result, even if here refuses it will still be there.

G-Ray: Yeah I'm on it

src/adapters/poi/specials/navigator_geolocalisation_poi.js
```diff
@@ -24,7 +24,9 @@ export default class NavigatorGeolocalisationPoi extends Poi {
   }
 
   async geolocate() {
-    await GeolocationCheck.checkPrompt();
+    const isAuthorized = await GeolocationCheck.checkPrompt();
```
bbecquet, line 27: Shouldn't you pass a callback here?

src/libs/geolocation.js
```diff
@@ -7,22 +7,39 @@ const geolocationPermissions = {
...
-    return window.navigator.permissions.query({ name: 'geolocation' }).then(async p => {
-      if (p.state === geolocationPermissions.PROMPT) {
-        if (window._GEO_QUESTION_ASKED !== true) {
+    return window.navigator.permissions
+      .query({ name: 'geolocation' })
+      .then(async p => {
+        if (p.state === geolocationPermissions.DENIED) {return cb(false);}
+        if (p.state === geolocationPermissions.GRANTED) {return cb(true);}
+        if (p.state === geolocationPermissions.PROMPT) {
+          if (window._GEO_QUESTION_ASKED === true) {return cb(false);}
```
bbecquet, line 29: I'm not sure I understand this case. Can you explain to which scenario it corresponds?

src/libs/geolocation.js
```diff
@@ -7,22 +7,39 @@ const geolocationPermissions = {
...
+
           window._GEO_QUESTION_ASKED = true;
```
bbecquet, line 31: Let's use your refactoring to stop using a global for that :)

src/adapters/poi/specials/navigator_geolocalisation_poi.js
```diff
@@ -24,7 +24,11 @@ export default class NavigatorGeolocalisationPoi extends Poi {
   }
 
   async geolocate() {
-    await GeolocationCheck.checkPrompt();
+    const isAuthorized = await new Promise(resolve =>
+      GeolocationCheck.checkPrompt(isAuthorized => resolve(isAuthorized))
+    );
+    if (!isAuthorized) {return;}
```
bbecquet, line 30: You are changing the signature of the function by not returning a rejecting Promise here. It may disturb other parts of the app relying on that. https://github.com/QwantResearch/erdapfel/blob/master/src/panel/direction/DirectionInput.jsx#L63-L72

src/libs/geolocation.js
G-Ray: It seems we don't want to display the geolocation modal multiple times if the user has already seen it, while authorization is still in status 'PROMPT'

src/libs/geolocation.js
```diff
@@ -7,30 +7,30 @@ const geolocationPermissions = {
...
-  static handleError(error) {
-    if (error.code === 1) {
-      // PERMISSION_DENIED
-      fire('open_geolocate_denied_modal');
-    } else {
-      fire('open_geolocate_not_activated_modal');
-    }
+  const p = await window.navigator.permissions.query({ name: 'geolocation' });
+  if (p.state !== geolocationPermissions.PROMPT) {
+    // allowed on denied
```
xem, line 20: what does "allowed on denied" mean?

G-Ray: typo :100:

xem (dismissed):
As we said, let's merge that for now and finish this permission thing in the 1754 PR

src/panel/direction/DirectionPanel.jsx
```diff
@@ -56,11 +59,23 @@ export default class DirectionPanel extends React.Component {
...
-  componentDidMount() {
+  async componentDidMount() {
     Telemetry.add(Telemetry.ITINERARY_OPEN);
     document.body.classList.add('directions-open');
     this.dragPointHandler = listen('change_direction_point', this.changeDirectionPoint);
     this.setPointHandler = listen('set_direction_point', this.setDirectionPoint);
+
+    if (!this.props.origin && isMobileDevice()) {
+      // If authorized, set origin to current position on mobile
+      await GeolocationCheck.showGeolocationModalIfNeeded();
```
bbecquet, line 70: Why do you need to do it here explicitely, whereas it's already done in the `geolocate` method?

bbecquet (dismissed):

xem (approved):

PR #832, opened 2020-10-12
Qmaps 1754 - Popup redesign

xem:
## Description
Specs en résumé:

Sur desktop: jamais de popup custom! (à part en cas d'erreur)

Sur mobile:
- Afficher la popup redesignée à la première ouverture du panel itinéraire vierge, et si la géoloc n'a pas encore été autorisée pour ce navigateur
- Clic sur la croix: on retient le choix de l'utilisateur de ne pas requêter sa position aux futurs lancement du panel itinéraire. On affiche le formulaire vierge mais "your position" reste disponible dans le suggest de chaque champ.
- Clic sur "j'ai compris": on retient ce choix, et on requête sa position immédiatement (et on la requêtera à chaque future ouverture du panel itinéraire). Si l'user permet l'accès à sa position dans la popup native, on affiche sa position dans le champ de départ. Sinon, on affiche le formulaire vierge.
- Ouverture du panel itinéraire pas vierge (ex: depuis un poi ou un favori):
-> si l'autorisation de geoloc a déja été donnée: requêter la position et la mettre dans le champ départ. Le PoI lui est mis dans le champ arrivée, et l'itinéraire est calculé.
-> si l'autorisation de geoloc n'a pas été donnée: pas de requête de la position tant qu'on a pas choisi "ma position" dans le champ départ. Ce champ reste vide. L'autre contient le poi

Sur desktop et mobile:
- En cas de clic sur le bouton geoloc de MapBox: pas de popup custom, mais requêter la position de l'utilisateur et afficher la popup native si la permission n'a pas encore été reçue.
- Afficher les modal d'erreurs comme avant (mais avec le nouveau style) si la position est demandée mais pas accessible

## Why
Design.

## Screenshots
image

image


src/scss/includes/modals/index.scss
```diff
@@ -6,22 +6,28 @@
   left: 0;
   width: 100vw;
   height: 100vh;
-  background-image: linear-gradient(to bottom, #6c7694, #353c52);
+  background: rgba(12, 12, 14, 0.4);
```
G-Ray, line 9: don't forget to define colors in colors.scss each time ;)

src/panel/direction/DirectionPanel.jsx
```diff
@@ -56,27 +55,17 @@ export default class DirectionPanel extends React.Component {
     };
 
     this.restorePoints(props);
+
+    if (!this.state.origin && !this.state.destination && !props.origin && !props.destination) {
```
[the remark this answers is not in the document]
G-Ray: yeah, this is also where the current location by default is handled on mobile

G-Ray:
The popup is nice ! Just a minor tweak : the bottom padding height is a bit high isn't it ?
image

src/modals/GeolocationModal.jsx
```diff
@@ -45,24 +52,31 @@ function close() {
...
 export async function openAndWaitForClose() {
   return new Promise(resolve => {
-    open('PENDING', () => {
-      close();
-      resolve();
-    });
-  });
+    open(
+      'PENDING',
+      () => {
```
G-Ray, line 70: you can pass only one onClick handler, and take a status (false or true) as param
```js
  (isAuthorized) => {
    close();
    resolve(isAuthorized); // click "OK": allow native geolocation popup
  }
```

src/scss/includes/modals/index.scss
```diff
@@ -6,49 +6,54 @@
...
+  color: $grey-black;
+  font-weight: bold;
   margin-bottom: 8px;
 }
 
 .modal__subtitle {
   display: block;
-  font-size: 13px;
-  color: $secondary_text;
+  font-size: 14px;
```
G-Ray, line 54: these 3 font properties are same as `u-text--subtitle` I think ?

src/scss/includes/modals/index.scss
```diff
@@ -6,49 +6,54 @@
...
   position: absolute;
-  padding: 16px 10px;
   right: 0;
   top: 0;
-  color: $secondary_text;
-  font-size: 18px;
-  cursor: pointer;
 }
 
 .modal__title {
```
G-Ray, line 44: u-text--smallTitle ?

src/scss/includes/modals/index.scss
```diff
@@ -60,13 +65,20 @@
 }
 
 .modal__button {
-  width: 271px;
+  width: 100%;
   height: 35px;
```
bbecquet, line 69: Now that it's a standard button, you don't need any of these rules, except for `width`.

src/scss/includes/modals/index.scss
```diff
@@ -6,49 +6,54 @@
...
 .modal {
   position: relative;
-  border-radius: 3px;
+  border-radius: 12px;
   overflow: auto;
+  background: $background;
+  box-shadow: 0 8px 64px 0 rgba(12, 12, 14, .12);
+  text-align: center;
+  width: 305px;
+  padding: 35px 0 20px;
```
bbecquet, line 22: There is no top padding in the design.

src/libs/geolocation.js
```diff
@@ -10,20 +11,24 @@ const geolocationPermissions = {
...
-    return;
+    return false;
   }
 
+  // granted or denied
   const p = await window.navigator.permissions.query({ name: 'geolocation' });
   if (p.state !== geolocationPermissions.PROMPT) {
-    // allowed or denied
-    return;
+    return p.state === geolocationPermissions.GRANTED;
```
bbecquet, line 26: Not sure, but if permission has already been granted, we shouldn't show the modal again, right? So we should return false here, no?

src/adapters/poi/specials/navigator_geolocalisation_poi.js
```diff
@@ -23,8 +23,14 @@ export default class NavigatorGeolocalisationPoi extends Poi {
...
   }
 
-  async geolocate(options = { displayErrorModal: true }) {
-    await Geolocation.showGeolocationModalIfNeeded();
+  async geolocate(options = { displayErrorModal: true, displayDirectionModalIfNeeded: false }) {
+    let requestPosition = true;
+    if (options.displayDirectionModalIfNeeded) {
+      requestPosition = await Geolocation.showGeolocationModalIfNeeded();
+      if (!requestPosition) {
+        reject();
```
bbecquet, line 31: This looks like a copy-paste: this causes a linting error and will fail at runtime as `reject` is undefined. Not sure what you needed here, maybe just `return`?

src/modals/GeolocationModal.jsx
```diff
@@ -3,40 +3,47 @@ import React from 'react';
...
-    <div className="modal__maps">
-      <i className="icon-x modal__close" onClick={onClose} />
-      <h2 className="modal__title">{title}</h2>
-      <div className="modal__subtitle" dangerouslySetInnerHTML={{ __html: text }} />
-      <button className="modal__button" onClick={onClose}>{button}</button>
+    <div className={classnames('modal__maps', className)}>
+      <CloseButton onClick={onClose} />
+      <div className="modal__maps__content">
+        <h2 className="modal__title u-text--smallTitle">{title}</h2>
+        <div className="modal__subtitle u-text--subtitle" dangerouslySetInnerHTML={{ __html: text }} />
```
bbecquet, line 44: According to the linter, this line is too long.

xem: sorry, my local linter was misbehaving

src/adapters/poi/specials/navigator_geolocalisation_poi.js
xem: @bbecquet indeed the reject() alone was an error.
Though, I tried many ways to "return early" in this function (return, reject, ...), but every time, when closing the modal with the "x" button, I end up having this error in my console, which seems unrelated but only happens when the return early is in the code 🤔 
image

src/libs/geolocation.js
xem: No, for me it's right the way it is:

If geolocationPermissions.GRANTED is false, the user refused to be located, showGeolocationModalIfNeeded returns true, and the function that calls it (geolocate) does no native request.

If geolocationPermissions.GRANTED is true, the user accepted to be located, showGeolocationModalIfNeeded returns true, and geolocate continues by doing the native request.

Otherwise, it means the permission is pending, so the modal is shown by showGeolocationModalIfNeeded

src/adapters/poi/specials/navigator_geolocalisation_poi.js
bbecquet: The error isn't unrelated. It happens when you return early without setting a position on the `NavigatorGeolocalisationPoi`. As you didn't exit by rejecting, the calling code assumes everything went right and try to use the lat/lng property on this special POI object, but it doesn't have any…

Your problem can probably be solved by moving this block in the Promise below, so the method has only on exit point (the Promise with its two clear states), instead of two.

But I know this whole geolocation process is confusing, the previous code doesn't help. IMO it tries too much to be clever instead of using plain and simple isolated functions. The `showGeolocationModalIfNeeded` method is quite confusing to me, I'm never sure of how it should behave, hence my other comment.

Can we start by listing what `showGeolocationModalIfNeeded` should return in each case?

G-Ray: Previously, `showGeolocationModalIfNeeded` displayed a popup if needed (hence, the use has not given permission yer, and has not been asked to do so), and was "blocking" execution. It was not returning anything IIUC.
So, in any case, any caller of `showGeolocationModalIfNeeded` would still need to ask for geolocation using `poi.geolocate(options)`, which could return the location, or throw an error if location access is forbidden.

G-Ray: From now, we would like to know if user has accepted the popup, in which case we will try to call `geolocate()`, or do nothing if the popup has been dismissed.

So, `showGeolocationModalIfNeeded` does not really make sense now, we could name it as `isGeolocationModalAccepted` for instance. Its signature could be:

`const isGeolocationModalAccepted = function (): boolean`

Il would block until the user has clicked an option, and return :
- true when user clicked on accept button
- false when user clicked on close button

bbecquet: I would just name such function `isGeolocationAccepted`, without referring to a modal at all, since it many cases there won't be one.
But it looks like in this PR we are mixing two things:
 - the decision to display a custom modal (depending on the platform, the user action, the feacture)
 - the decision of the user to accept the geolocation (no: closing the custom modal or refusing the native browser modal, yes: accepting the native browser modal) 
Do you think your proposition fills all the needs?

xem: Yes, but it's improvable. If you're ok with this, I will separate the modal logic and the native popup logic into 2 functions, and move the "if(requestPosition) back inside the promise.
After merge, we'll see if we want to improve that part more deeply (but I think it'll be enough)
OK?

bbecquet: Great!

src/modals/GeolocationModal.jsx
```diff
@@ -3,40 +3,50 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import Modal from 'src/components/ui/Modal';
 import { listen } from 'src/libs/customEvents';
+import { Button, CloseButton } from 'src/components/ui';
+import classnames from 'classnames';
 
-const GeolocationModal = ({ status, onClose }) => {
+const GeolocationModal = ({ status, onClose, onClick }) => {
```
bbecquet, line 9: `onClick` isn't a precise enough name at this level. It gives no idea of what it does or what kind of element it applies to. `onAccept` would be clearer.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -65,11 +66,27 @@ export default class DirectionPanel extends React.Component {
...
+
+    // On mobile, and if both origin and destination are empty, check the browser's permissions
+    if (
+      !this.state.origin
+      && !this.state.destination
+      && !this.props.origin
+      && !this.props.destination
+      && isMobileDevice()
+    ) {
+      requestPosition = await Geolocation.showGeolocationModalIfNeeded();
```
bbecquet, line 80: I thought you were going to change what `showGeolocationModalIfNeeded` does, or at least its name.
Here we don't see clearly how the fact to show the modal (only once, and should be remembered) is related to requesting the position (possibly each time).
Both concepts should be separated.

src/libs/geolocation.js
```diff
@@ -10,20 +11,24 @@ const geolocationPermissions = {
 let hasPermissionModalOpenedOnce = false;
 
 export async function showGeolocationModalIfNeeded() {
+  if (!isMobileDevice()) {
```
bbecquet, line 14: You're doing this test in DirectionPanel, it can be removed from here to focus only on the permission stuff.

src/libs/geolocation.js
```diff
@@ -10,20 +11,24 @@ const geolocationPermissions = {
...
+    return false;
+  }
+
+  // Some browsers (Safari, etc) do not implement Permissions API
   if (!window.navigator.permissions) {
-    // Some browsers (Safari, etc) do not implement Permissions API
-    return;
+    return false;
```
bbecquet, line 20: If I understand correctly, this means we'll never request the position on Safari.

xem: Yes, in the current implementation, Safari won't see the Modal and won't request position when opening Directions panel.
We'll see after pushing in prod if the modal choice should rather be saved in localStorage (in that case, we can be sure it only appears once and can be enabled on Safari too without depending on the Permissions api)

src/panel/direction/DirectionPanel.jsx
```diff
@@ -65,11 +66,27 @@ export default class DirectionPanel extends React.Component {
...
+
+    // On mobile, and if both origin and destination are empty, check the browser's permissions
+    if (
+      !this.state.origin
+      && !this.state.destination
+      && !this.props.origin
+      && !this.props.destination
+      && isMobileDevice()
+    ) {
+      requestPosition = await Geolocation.directionModalOnFirstVisit();
```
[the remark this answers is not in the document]
amatissart: I agree that the permission status and the modal logic should be separated more clearly. 

For example, according to this PR description, the current location is supposed to be used as origin when the permission has already been accepted and the destination is set on initialization (eg, I clicked on "directions" from a Poi panel). No custom modal should be triggered in this case. Hence, t

The location is currently not used in this case. It's probably not a blocking issue, but separating the concerns would make future changes easier.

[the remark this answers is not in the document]
xem: Your proposition doesn't work @G-Ray 
isLocationAvailable can't return a simple boolean.
it must return something that tells directionPanel if the permission is prompt (to show the modal), or granted (to ask position) or denied (to do nothing).
So I made it return the permission status directly, then made directionPanel handle the Modal opening and the native popup opening.
Hope you'll like it.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -65,17 +67,32 @@ export default class DirectionPanel extends React.Component {
...
-        );
-      } catch (e) {
-        // ignore possible error
+
+    // on mobile, when no origin is specified, try auto-geoloc
+    if (isMobileDevice() && !this.state.origin && !this.props.origin) {
+      const geolocationPermission = await getGeolocationPermission();
+      let modalAccepted = false;
+
+      // on an empty from, if the user's position permission hasn't been asked yet, show modal
```
G-Ray, line 76: from typo

src/panel/direction/DirectionPanel.jsx
```diff
@@ -65,17 +67,32 @@ export default class DirectionPanel extends React.Component {
...
+      if (!this.state.destination && !this.props.destination
```
G-Ray, line 77: ok, it's easier to understand now. The code might be a bit complicated with nested if and try/catch, but it's ok

G-Ray (dismissed):

bbecquet (dismissed):

bbecquet (approved):

xem:
woohoo

PR #834, opened 2020-10-13
Add missing call to computeRoutes on mobile + loc

G-Ray:
## Description
Location was used as origin, but we forgot to call `this.update` (which call `this.computeRoutes` and`this.updateUrl`) :medal_sports:

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -69,7 +69,10 @@ export default class DirectionPanel extends React.Component {
       const origin = new NavigatorGeolocalisationPoi();
       try {
         await origin.geolocate({ displayErrorModal: false });
-        this.setState({ origin, originInputText: origin.name });
+        this.setState(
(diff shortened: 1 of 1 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -69,7 +69,10 @@ export default class DirectionPanel extends React.Component {
       const origin = new NavigatorGeolocalisationPoi();
       try {
         await origin.geolocate({ displayErrorModal: false });
-        this.setState({ origin, originInputText: origin.name });
+        this.setState(
+          { origin, originInputText: origin.name },
+          this.computeRoutes
```
bbecquet, line 74: In other cases we use `this.update`, which also updates the url. It would be better here also for consistence, except if you have a reason to only call `computeRoutes`.

G-Ray: ok, so let's call this.update instead

bbecquet (approved):

PR #838, opened 2020-10-13
Qmaps 1775 - Fix suggest overlap route results

bbecquet:
## Description
Two fixes on the relationship between the top (form) and bottom (result) parts of the direction mobile UI.
 - ensure the suggest dropdown is displayed above the result panel, instead of below
 - ensure the result panel maximized height is limited by the top part

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -329,8 +329,8 @@ export default class DirectionPanel extends React.Component {
           {!activePreviewRoute && origin && destination &&
             <Panel
               resizable
-              fitContent={['default']} 
-              marginTop={this.mobileFormRef.current.offsetHeight}
+              fitContent={['default']}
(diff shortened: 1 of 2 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -323,7 +325,7 @@ export default class DirectionPanel extends React.Component {
             />}
           </div>
           {!activePreviewRoute && origin && destination &&
-            <Panel resizable marginTop={160} >
+            <Panel resizable marginTop={this.mobileFormRef.current.offsetHeight} >
```
G-Ray, line 328: Great :+1:  I think it's safer to check for mobileFormRef.current before.
image

G-Ray (dismissed):

G-Ray:
I opened a new PR https://github.com/QwantResearch/erdapfel/pull/853 with 1 more commit to fix the remaining issue.

PR #840, opened 2020-10-13
Qmaps 1602 pin marker icons for filtered POI lists

bbecquet:
## Description
Uses this development branch of map-style-builder: https://github.com/QwantResearch/map-style-builder/pull/27
It generates new sprites including new pin-shaped icons:
sprite

Also adds a standalone SVG pin for hovered/selected states.

## Why
Distinct design between filtered POI results (typically categories) and map layer POIs.

## Screenshots
Capture d’écran de 2020-10-13 16-59-36

### DIFF
--- a/src/adapters/icon_manager.js
+++ b/src/adapters/icon_manager.js
@@ -95,3 +95,14 @@ export function createMapGLIcon(imageFile, width, height) {
     img.src = imageFile;
   });
 }
+
+export function createPinIcon(iconOptions) {
+  const element = document.createElement('div');
+  element.innerHTML = `
+    <div
+      class="marker ${iconOptions.className || ''}"
+      ${iconOptions.disablePointerEvents && 'style="pointer-events:none;"'}
+    ></div>
+  `;
+  return element.firstElementChild;
+}
--- a/src/adapters/icon_manager.js
+++ b/src/adapters/icon_manager.js
@@ -1,4 +1,5 @@
 import styleIcons from '@qwant/qwant-basic-gl-style/icons.yml';
+import classnames from 'classnames';
 
 const nameToClass = iconName => iconName.match(/^(.*?)-[0-9]{1,2}$/)[1];
 
@@ -96,13 +97,11 @@ export function createMapGLIcon(imageFile, width, height) {
   });
 }
 
-export function createPinIcon(iconOptions) {
+export function createPinIcon({ className, disablePointerEvents }) {
   const element = document.createElement('div');
-  element.innerHTML = `
-    <div
-      class="marker ${iconOptions.className || ''}"
-      ${iconOptions.disablePointerEvents && 'style="pointer-events:none;"'}
-    ></div>
-  `;
-  return element.firstElementChild;
+  element.className = classnames('marker', className);
+  if (disablePointerEvents) {
+    element.style = 'pointer-events:none;';
+  }
+  return element;
 }
--- a/src/adapters/pois_styles.js
+++ b/src/adapters/pois_styles.js
@@ -2,7 +2,7 @@
 export const getFilteredPoisStyle = ({ withName = true } = {}) => ({
   type: 'symbol',
   layout: {
-    'icon-image': ['concat', 'pin-', ['get', 'subclass']],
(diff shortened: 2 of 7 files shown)


src/adapters/pois_styles.js
```diff
@@ -2,10 +2,9 @@
 export const filteredPoisStyle = {
   type: 'symbol',
   layout: {
-    'icon-image': 'pin_with_dot',
+    'icon-image': ['concat', 'pin-', ['get', 'subclass']],
```
amatissart, line 5: The icon associated to the POI is not always identical to the subclass name.

Either this layer should integrate the same kind of `icon-image` rules (that are quite long), currently generated by the "icons" task in the map-style-builder to build other POIs layers used by vector tiles.
Or probably simpler, `poisToGeoJSON` should use the IconManager to evaluate the icon associated with each poi, and store it as an extra GeoJSON property.

src/adapters/icon_manager.js
```diff
@@ -95,3 +95,14 @@ export function createMapGLIcon(imageFile, width, height) {
...
+
+export function createPinIcon(iconOptions) {
+  const element = document.createElement('div');
+  element.innerHTML = `
+    <div
+      class="marker ${iconOptions.className || ''}"
+      ${iconOptions.disablePointerEvents && 'style="pointer-events:none;"'}
+    ></div>
+  `;
+  return element.firstElementChild;
```
G-Ray, line 107: `return element` should be equivalent ?

G-Ray:
Do you know if we could do something to prevent this flickering to happen when hovering + slight cursor moves ?

Peek 2020-10-16 11-19

src/adapters/icon_manager.js
bbecquet: No, it would create useless `<div>`s in the map element, and the style doesn't apply the same… That's why you have the ugly hover effect in your gif below, because you fiddled with that :)
But I get what bothers you, it's that we declare a `<div>` only to fill up with innerHTML. I just used the same syntax as the similar existing function, but here I can rewrite it in a better way.

src/adapters/scene_category.js
```diff
@@ -137,13 +147,23 @@ export default class SceneCategory {
     });
   }
 
+  setPoiFeatureState = (id, state) => {
+    this.map.setFeatureState({ id, source: DYNAMIC_POIS_LAYER }, state);
+  }
+
   highlightPoiMarker = (poi, highlight) => {
+    if (this.hoveredPoi) {
```
G-Ray, line 155: `this.hoveredPoi` does not seem to be use anywhere except in this `highlightPoiMarker` function, did I miss its use cases ?

bbecquet: We need to store it to be able to remove the hovered effect on it.

src/adapters/scene_category.js
```diff
@@ -4,14 +4,26 @@ import constants from '../../config/constants.yml';
...
 import { getFilteredPoisStyle } from 'src/adapters/pois_styles';
-import { createMapGLIcon, createIcon } from 'src/adapters/icon_manager';
 import { isMobileDevice } from 'src/libs/device';
+import { createMapGLIcon, createPinIcon } from 'src/adapters/icon_manager';
+import IconManager from 'src/adapters/icon_manager';
 
 const DYNAMIC_POIS_LAYER = 'poi-filtered';
 const mapStyleConfig = nconf.get().mapStyle;
 
+const poisToGeoJSON = pois => {
```
G-Ray, line 16: I think we can move this function to `src/libs/geojson`, this way `poiToGeoJSON` and `poisToGeoJSON` are side by side.

bbecquet: You're right in principle :) It was actually there at first, but referencing `IconManager` from `src/libs/geojson` made the unit tests crash.
The reason is `IconManager` itself imports a YAML file as as a dependency, but the yaml loader for Jest doesn't support the placeholder syntax we use in this file… This is a mess, I spent three whole hours on it, so to advance I moved the function calling IconManager in this file which isn't unit-tested.

G-Ray (approved):

amatissart (approved):

PR #842, opened 2020-10-14
Qmaps 1788 direction panel fitcontent

G-Ray:
## Description
- Panel now accept an array of strings (`default`, 'minimized`) for `fitContent` prop. This way, we can keep the existing behavior for PoiPanel, which need to be fitted in default and minimized states, meanwhile direction's results should be fitted only in default size.
Please note fitContent will have no effect if the content's size is higher than the DEFAULT_SIZE.
- Use Panel's `fitContent` prop in DirectionPanel.

## Why
Avoid useless space in panel when only one result in direction panel is displayed

### One result:
image

### multiple results
Defaut size is used as the content overflows DEFAULT_SIZE constant. Should we tweak the default size in this case ? Please note this is the current behavior.
image

### DIFF
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -192,11 +192,11 @@ class Panel extends React.Component {
     const panelHeight = this.panelContentRef.current.offsetHeight;
     const values = {
       'default': height -
-        (fitContent.includes('default') &&
+        (fitContent.indexOf('default') >= 0 &&
         (panelHeight + FIT_CONTENT_PADDING <= DEFAULT_SIZE)
           ? panelHeight + FIT_CONTENT_PADDING
           : DEFAULT_SIZE),
-      'minimized': height - (fitContent.includes('minimized')
+      'minimized': height - (fitContent.indexOf('minimized') >= 0
         ? panelHeight + FIT_CONTENT_PADDING
         : DEFAULT_MINIMIZED_SIZE),
       'maximized': 0,
--- a/src/components/ui/Panel.jsx
+++ b/src/components/ui/Panel.jsx
@@ -52,6 +52,7 @@ class Panel extends React.Component {
   }
 
   static defaultProps = {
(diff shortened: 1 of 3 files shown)


src/components/ui/Panel.jsx
```diff
@@ -191,10 +191,12 @@ class Panel extends React.Component {
...
-      'default': height - (fitContent
-        ? panelHeight + FIT_CONTENT_PADDING
-        : DEFAULT_SIZE),
-      'minimized': height - (fitContent
+      'default': height -
+        (fitContent.includes('default') &&
+        (panelHeight + FIT_CONTENT_PADDING <= DEFAULT_SIZE)
+          ? panelHeight + FIT_CONTENT_PADDING
+          : DEFAULT_SIZE),
+      'minimized': height - (fitContent.includes('minimized')
```
amatissart, line 199: Not sure if this function is called on desktop, but [`.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) is not available on IE11.

xem: Isn't it removed during build by the js transpiler ?

amatissart: I don't think so. The transpiler deals with the js syntax, but doesn't transform such method calls.

src/components/ui/Panel.jsx
```diff
@@ -44,7 +44,7 @@ class Panel extends React.Component {
     children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
     minimizedTitle: PropTypes.node,
     resizable: PropTypes.bool,
-    fitContent: PropTypes.bool,
+    fitContent: PropTypes.arrayOf(PropTypes.oneOf(['default', 'minimized'])),
```
bbecquet, line 47: Let's add a default value `[ ]` for this prop, to prevent some errors.

PR #843, opened 2020-10-14
Move minimizedTitle in content instead of drawer

G-Ray:
## Description
- Move minimizedTitle from drawer to content instead

## Screenshots
image

### DIFF
--- a/src/scss/includes/components/panel.scss
+++ b/src/scss/includes/components/panel.scss
@@ -80,12 +80,6 @@ $bottom-margin: 40px;
       }
     }
 
-    &.minimized {
-      .panel-drawer {
-        min-height: 50px;
-      }
-    }
-
     .panel-drawer {
       flex-shrink: 0;
     }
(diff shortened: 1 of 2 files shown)


xem (approved):

src/scss/includes/components/panel.scss
```diff
@@ -87,6 +87,7 @@ $bottom-margin: 40px;
     }
 
     .panel-drawer {
+      width: 100%;
       flex-shrink: 0;
     }
```
amatissart, line 92: Isn't that change in conflict with recent changes in https://github.com/QwantResearch/erdapfel/pull/815
where width is set explicitly at 50% on line 31 ?

G-Ray:
> Two different values of `width` are set for `.panel-drawer` in panel.scss. So this should be clarified before merging.

you are right, I missed search for `-drawer`. @bbecquet I'm not sur in which case setting width to 50% is needed ?

src/scss/includes/components/panel.scss
bbecquet: > I'm not sur in which case setting width to 50% is needed ?

@G-Ray it's a usability feature, related to the close button, so a user who missed the touch on the button by one pixel doesn't risk to minimize the panel.
I think the right way to fix this 2-lines issue is to move the title outside of the drawer.

amatissart (approved):

PR #850, opened 2020-10-16
Qmaps 1791 fix step by step

bbecquet:
## Description
"Quick fix" of the step-by-step mode design.
Remove the direction form element remaining behind, normalize margins and add a close button.

## Why
Just make things appear not broken, until the real redesign is prioritized.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran 2020-10-16 à 10 54 52|Capture d’écran 2020-10-16 à 10 54 23|

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -328,8 +328,10 @@ export default class DirectionPanel extends React.Component {
             <Panel resizable marginTop={160} fitContent={['default']}>
               {result}
             </Panel>}
-          {activePreviewRoute &&
-            <MobileRoadMapPreview steps={getAllSteps(activePreviewRoute)} />}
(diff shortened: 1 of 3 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -310,26 +310,28 @@ export default class DirectionPanel extends React.Component {
...
+          {!activePreviewRoute && <div className="direction-panel">
             <Flex
               className="direction-panel-header"
               alignItems="center"
               justifyContent="space-between">
               {title}
               <CloseButton onClick={this.onClose} />
             </Flex>
-            {!activePreviewRoute && form}
+            {form}
```
G-Ray, line 321: are you sure this is needed ? No side effect could occur by still rendering the form ?

G-Ray (approved):

PR #857, opened 2020-10-19
route steps design

G-Ray:
## Description
Update route steps according to new design :
- Updated fonts
- Updated layout : distance is now displayed under the direction step.
- Updated origin and destination icons, for now I used the new `svg` icon to be consistent with the new direction form. Integrating new icons asked by design + colors logic (depending on the type of poi) will be implemented in another changes set.
- Used Divider where possible and cleaned some css styles. Hopefully nothing is broken.

## Screenshots
mobile
desktop

### DIFF
--- a/src/scss/includes/direction.scss
+++ b/src/scss/includes/direction.scss
@@ -12,6 +12,18 @@
   box-shadow: 0 0 4px 0 rgba(0, 0, 0, 0.14);
 }
 
+.itinerary_result {
+  .itemList-item {
+    &:not(:last-child):after {
+      content: '';
+      background-color: #e0e1e6;
+      height: 1px;
+      margin: 0 33px;
+      display: block;
+    }
+  }
+}
+
 .itinerary_no-result {
   text-align: center;
   padding: 25px 50px;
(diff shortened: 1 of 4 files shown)


src/scss/includes/direction.scss
```diff
@@ -50,16 +50,18 @@
   overflow: auto;
 }
 
-.itinerary_leg {
-  position: relative;
-
+.itemList-item {
```
bbecquet, line 53: Make sure to scope that as you are modifying the global `.itemList-item` component here. Or if it's the target look and consistent with other uses in the app, we should modify it in its own SCSS.

bbecquet (commented):
Nice simplification.
Some specific rules are still need to make sure the `RoadMapStep` component looks good even when used in the mobile step-by-step context, as now the paddings look off.
Capture d’écran de 2020-10-19 15-50-06

G-Ray:
> Nice simplification.
> Some specific rules are still need to make sure the `RoadMapStep` component looks good even when used in the mobile step-by-step context, as now the paddings look off.
> Capture d’écran de 2020-10-19 15-50-06

Thanks, should be better now.
image

src/scss/includes/direction.scss
G-Ray: thanks, this should be a temporary fix, the delimiter has disappeared a few commits ago it seems, and it should be removed in a next commit due to design changes

bbecquet (approved):

PR #859, opened 2020-10-20
Don't focus the search when the MacOS Cmd 'key' is used

bbecquet:
## Description
Same thing as https://github.com/QwantResearch/erdapfel/pull/846 but for the Mac OS "Cmd" key, which is mapped on the `metaKey` key event property.

## Why
Don't disturb native OS functions.

### DIFF
--- a/src/ui_components/search_input.js
+++ b/src/ui_components/search_input.js
@@ -75,7 +75,7 @@ export default class SearchInput {
       }
       // KeyboardEvent.key is either the printed character representation or a standard value for specials keys
       // See https://developer.mozilla.org/fr/docs/Web/API/KeyboardEvent/key/Key_Values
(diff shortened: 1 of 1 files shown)


src/ui_components/search_input.js
```diff
@@ -75,7 +75,7 @@ export default class SearchInput {
       }
       // KeyboardEvent.key is either the printed character representation or a standard value for specials keys
       // See https://developer.mozilla.org/fr/docs/Web/API/KeyboardEvent/key/Key_Values
-      if (e.key.length === 1 && !e.ctrlKey) {
+      if (e.key.length === 1 && !e.ctrlKey && !e.metaKey) {
```
G-Ray, line 78: I guess we will use an array soon :trollface:

G-Ray (approved):

PR #861, opened 2020-10-20
Qmaps 1795 route summary info

G-Ray:
## Description
Define and reuse a `RouteComponentInfo` component on mobile and desktop. It renders route info like distance, duration, and "via" name.
- in `RouteResult` (desktop and mobile)
- in `MobileRouteDetails`
- Part of the new design on route summaries

## Why
New design

## Screenshots
desktop
mobile
mobile route details

### DIFF
--- a/src/panel/direction/MobileRouteDetails.jsx
+++ b/src/panel/direction/MobileRouteDetails.jsx
@@ -1,10 +1,10 @@
 /* global _ */
 import React, { useRef } from 'react';
-import RouteVia from './RouteVia';
 import RoadMap from './RoadMap';
-import { Badge, Button, CloseButton, Divider, Flex } from 'src/components/ui';
-import { formatDuration, formatDistance } from 'src/libs/route_utils';
+import { Button, Divider, Flex } from 'src/components/ui';
 import classnames from 'classnames';
+
+import RouteSummaryInfo from './RouteSummaryInfo';
 
 const MobileRouteDetails =
 ({ id, route, origin, destination, vehicle, toggleDetails, openPreview }) => {
@@ -18,22 +18,18 @@ const MobileRouteDetails =
   >
     <div className="mobile-route-details-header">
       <div className="mobile-route-details-header-content">
-        <Flex alignItems="center" justifyContent="space-between" className="u-text--title">
-          {formatDuration(route.duration)}
-          <CloseButton onClick={() => { toggleDetails(id); }} />
-        </Flex>
-
-        <RouteVia className="u-mb-8" route={route} vehicle={vehicle} />
-        <Badge className={isPublicTransport ? 'u-mb-20' : ''}>
-          {formatDistance(route.distance)}
-        </Badge>
+        <RouteSummaryInfo
+          route={route}
+          vehicle={vehicle}
+          onClose={() => toggleDetails(id)}
+        />
 
         {isPublicTransport &&
           <Button
             className="u-firstCap"
             onClick={() => { openPreview(id); }}
             variant="primary"
-            style={{ width: '100%' }}
+            style={{ width: '100%', marginTop: 20 }}
           >
             <Flex alignItems="center" justifyContent="center">
               <img className="u-mr-4" src="./statics/images/direction_icons/guide.svg" />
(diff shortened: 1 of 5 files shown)


src/panel/direction/RouteSummaryInfo.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+import RouteVia from './RouteVia';
+import { Badge, CloseButton, Flex } from 'src/components/ui';
+import { formatDuration, formatDistance } from 'src/libs/route_utils';
+
+const RouteSummaryInfo = ({ route, vehicle, onClose }) =>
+  <div>
+    <Flex alignItems="center" justifyContent="space-between" className="u-text--title">
+      {formatDuration(route.duration)}
+      {onClose && <CloseButton onClick={onClose} />}
```
bbecquet, line 11: Let's move the button outside, and use `RouteSummaryInfo` itself as a `Flex` child in `MobileRouteDetails`.

src/panel/direction/RouteResult.jsx
```diff
@@ -108,9 +108,9 @@ export default class RouteResult extends React.Component {
                   </div>
                   <PlaceholderText length={10} className="itinerary_leg_via_details"/>
                 </div>
-                <div className="itinerary_leg_info">
-                  <PlaceholderText length={5} className="itinerary_leg_duration" />
-                  <PlaceholderText length={7} className="itinerary_leg_distance" />
+                <div>
```
bbecquet, line 111: Thanks for giving some :heart: to the placeholders :)

G-Ray: It's not perfect at all, I just clean up deprecated classes. I will try to do another pass once the design is done

src/panel/direction/MobileRouteDetails.jsx
```diff
@@ -18,22 +18,20 @@ const MobileRouteDetails =
...
       <div className="mobile-route-details-header-content">
-        <Flex alignItems="center" justifyContent="space-between" className="u-text--title">
-          {formatDuration(route.duration)}
-          <CloseButton onClick={() => { toggleDetails(id); }} />
+        <Flex alignItems="flex-start" justifyContent="space-between">
+          <RouteSummaryInfo
+            route={route}
+            vehicle={vehicle}
+          />
+          <CloseButton onClick={() => toggleDetails(id)} />
```
bbecquet, line 26: What about using the prop `position="topRight"`, sot it looks better positionned with the content paddings.

src/panel/direction/RouteSummary.jsx
```diff
@@ -40,31 +40,32 @@ export default class RouteSummary extends React.Component {
...
-      <div className="itinerary_leg_via">
-        <RouteVia route={route} vehicle={vehicle} />
+
+      <Flex>
+        <div className={`itinerary_leg_icon ${getVehicleIcon(vehicle)}`} />
+        <RouteSummaryInfo
+          route={route}
+          vehicle={vehicle}
+        />
+      </Flex>
```
bbecquet, line 50: Looks like all elements could be in a single Flex, but as there are other redesign steps coming, let's wait a bit.

src/panel/direction/RouteSummaryInfo.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import RouteVia from './RouteVia';
+import { Badge, Flex } from 'src/components/ui';
+import { formatDuration, formatDistance } from 'src/libs/route_utils';
+
+const RouteSummaryInfo = ({ route, vehicle }) =>
+  <div>
+    <Flex alignItems="center" justifyContent="space-between" className="u-text--title">
```
bbecquet, line 9: You don't need the Flex anymore since the Button has been removed.

src/panel/direction/RouteSummaryInfo.jsx
```diff
@@ -1,14 +1,14 @@
...
 
 import RouteVia from './RouteVia';
-import { Badge, Flex } from 'src/components/ui';
+import { Badge } from 'src/components/ui';
 import { formatDuration, formatDistance } from 'src/libs/route_utils';
 
 const RouteSummaryInfo = ({ route, vehicle }) =>
   <div>
-    <Flex alignItems="center" justifyContent="space-between" className="u-text--title">
+    <span className="u-text--title">
```
bbecquet, line 9: This would be better as a div, so it doesn't depend on the next element to be on its own line

bbecquet (approved):

PR #862, opened 2020-10-20
direction step : add hover effects

G-Ray:
## Description
Recent changes removed any hover effects.
- Added `active-gradient`, and background color change

Behavior should change with next design changes, the active-gradient should be displayed on currently selected step.

## Screenshots
Peek 2020-10-20 15-58

### DIFF
--- a/src/scss/includes/direction.scss
+++ b/src/scss/includes/direction.scss
@@ -194,10 +194,30 @@
 
 .itinerary_roadmap_item-block {
   padding: 10px 0;
+
+  &:hover {
+    background-color: $grey-bright;
+    cursor: pointer;
+    position: relative;
+
+    &:before {
+      content: '';
+      position: absolute;
+      top: 12px;
+      left: 0;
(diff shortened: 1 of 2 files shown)


src/scss/includes/direction.scss
```diff
@@ -194,10 +194,30 @@
...
+    cursor: pointer;
+    position: relative;
+
+    &:before {
+      content: '';
+      position: absolute;
+      top: 0;
+      left: 0;
+      width: 4px;
+      height: 100%;
```
bbecquet, line 209: The gradient doesn't seem right. In the design, it has 12px vertical margins and rounder corners only on the right. The gradient direction also seems to be top->bottom, contrary to the current `$active-gradient` which is left->right.  
Capture d’écran de 2020-10-20 15-49-14

G-Ray: thanks, should be fixed now

bbecquet (approved):

PR #863, opened 2020-10-20
RouteSummary: Update hover/active style

G-Ray:
## Description
- Define a `grey-bright` hover effect on Summaries
- Define and reuse an `active-border-gradient` mixin.

## Screenshots
Peek 2020-10-20 16-50

### DIFF
--- a/src/scss/includes/direction.scss
+++ b/src/scss/includes/direction.scss
@@ -92,7 +92,6 @@
 }
 
 .itinerary_leg_summary {
-  border-left: 4px solid $background;
   display: flex;
   justify-content: space-between;
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/scss/includes/direction.scss
```diff
@@ -3,6 +3,17 @@
 @import "./direction-field.scss";
 @import "./mobileRouteDetails.scss";
 
+@mixin active-border-gradient {
```
bbecquet, line 6: :+1:

bbecquet (commented):
There is a white border on the left visible on hover, I think we want to avoid that.

G-Ray:
> There is a white border on the left visible on hover, I think we want to avoid that.

thanks, it's fixed

bbecquet (approved):

PR #864, opened 2020-10-20
Direction Summary: use same btns on desktop/mobile + see less btn

G-Ray:
## Description
- Update Details buttons on desktop
- When active, display `See less` btn
- Display steps/share buttons only if current direction is active

## Screenshots
Peek 2020-10-21 17-13

### DIFF
--- a/src/panel/direction/RouteSummary.jsx
+++ b/src/panel/direction/RouteSummary.jsx
@@ -16,6 +16,7 @@ export default class RouteSummary extends React.Component {
     toggleDetails: PropTypes.func.isRequired,
     openPreview: PropTypes.func.isRequired,
     selectRoute: PropTypes.func.isRequired,
+    isActiveDetails: PropTypes.bool.isRequired,
   }
 
   onClick = () => {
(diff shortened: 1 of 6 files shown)


src/panel/direction/RouteSummary.jsx
```diff
@@ -49,35 +49,23 @@ export default class RouteSummary extends React.Component {
...
+      <Flex className="itinerary_leg_details-buttons">
+        <Button
+          className="itinerary_leg_detailsBtn u-mr-8"
+          onClick={this.onClickDetails}
+          icon="icon_list"
+        >
           {_('Details', 'direction')}
         </Button>
+
         <ShareMenu url={window.location.toString()} scrollableParent=".panel-content">
```
bbecquet, line 61: You can simplify your life and this component by just removing this one, as the new design finally gets rid of the ambiguous Share action on itinerary alternatives.

G-Ray: for now we will keep this button, in another changes set, we will move it below the direction form :)

bbecquet (approved):

PR #866, opened 2020-10-21
DirectionPanel: history.back on close

G-Ray:
Any downside doing this ?

## Description
In DirectionPanel, call `history.back()` when closing.

## Why
New UX:
- If the user comes from a POI, when closing the Direction Panel, the user should return to the poi.
- If the user comes from Home, when closing the Direction Panel, the user should return to the home (current behavior).

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -224,7 +224,9 @@ export default class DirectionPanel extends React.Component {
       this.setState({ activePreviewRoute: null });
     } else {
       Telemetry.add(Telemetry.ITINERARY_CLOSE);
-      window.app.navigateTo('/');
+      this.props.poi
+        ? window.history.back() // Go back to the poi panel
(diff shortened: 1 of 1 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -224,7 +224,7 @@ export default class DirectionPanel extends React.Component {
       this.setState({ activePreviewRoute: null });
     } else {
       Telemetry.add(Telemetry.ITINERARY_CLOSE);
-      window.app.navigateTo('/');
+      window.history.back();
```
bbecquet, line 227: Doing that unconditionnaly will exit the app if we arrived directly on the direction panel. We should have a mechanism to return to the home instead in such case.
The common pattern for this is to compare `document.referrer`, which can be empty, filled with the current domain name or another one when coming from the another site. But it's not updated for a SPA, so maybe we're screwed :/

bbecquet (approved):

PR #869, opened 2020-10-22
Add tag on OSM edits (fix #460)

### LINKED ISSUE #460: Add some hashtag to edits coming from QwantMaps
I think it would be nice to auto add some hashtags to changesets coming from QwantMaps.

In order to do this we could append hashtags to the `editUrl` and ID will add those hashtags to the changeset, when uploading.
https://github.com/QwantResearch/erdapfel/blob/6dc61116207661a725229cf430ba3c6bc0cf81f2/src/adapters/poi/idunn_poi.js#L38

According to the [ID API documentation](https://github.com/openstreetmap/iD/blob/efb361ad2a35a4dbc8b2c2e5bf017c0f6fb4d4a8/API.md#url-parameters) something like adding `&hashtags=QwantMaps` to the `editUrl` should do the job. Though I didn't get it working when manually entering such a link.

remi-dupre:
## Description

Based on an external suggestion (https://github.com/QwantResearch/erdapfel/issues/460), suggest the addition of a tag referring to QwantMaps when modifying a POI on OSM.

When using the link "modify", changing a POI on openstreetmap.org and validating the change, the confirm dialog will look like this:

Screenshot_2020-10-22_14-00-17

## Why

The main benefits I can see is that it could help with understanding how people are interested in updating POI informations when using https://qwant.com/maps.

### DIFF
--- a/src/components/OsmContribution.jsx
+++ b/src/components/OsmContribution.jsx
@@ -4,8 +4,9 @@ import Telemetry from 'src/libs/telemetry';
 
 const OsmContribution = ({ poi }) => {
   const [_osmKey, itemKind, itemId] = poi.id.split(':');
+  const editParams = `editor=id&${itemKind}=${itemId}&hashtags=QwantMaps`;
   const viewUrl = `https://www.openstreetmap.org/${itemKind}/${itemId}`;
-  const editUrl = `https://www.openstreetmap.org/edit?editor=id&${itemKind}=${itemId}`;
+  const editUrl = `https://www.openstreetmap.org/edit?${editParams}`;
 
   const sendTelemetryEvent = event => () => Telemetry.sendPoiEvent(poi, event);
 


src/components/OsmContribution.jsx
```diff
@@ -5,7 +5,7 @@ import Telemetry from 'src/libs/telemetry';
 const OsmContribution = ({ poi }) => {
   const [_osmKey, itemKind, itemId] = poi.id.split(':');
   const viewUrl = `https://www.openstreetmap.org/${itemKind}/${itemId}`;
-  const editUrl = `https://www.openstreetmap.org/edit?editor=id&${itemKind}=${itemId}`;
+  const editUrl = `https://www.openstreetmap.org/edit?editor=id&${itemKind}=${itemId}&hashtags=QwantMaps`;
```
bbecquet, line 8: Oups, line too long for the linter :)
You could use this util function if you want https://github.com/QwantResearch/erdapfel/blob/master/src/libs/url_utils.js#L64

remi-dupre: Indeed. I didn't go for the helper: as one of the keys values is defined by a variable (`itemKind`), it may be slightly less concise to define a dictionary with query parameters. Unless I miss a Javascript feature or just some coding convention.

bbecquet: If you mean dynamic object key names, you can do it indeed, by using [ ] :)
```js
{
   editor: 'id',
   [itemKind]: 'itemId',
}
```

bbecquet (approved):

PR #870, opened 2020-10-22
fix: No input value was set for latlon items without address

amatissart:
## Description
Extend and use `getInputValue` from "libs/suggest.js" to standardize the value used in input fields when an item is selected.

## Why
When dragging a direction point (origin or destination) to a latlon position where no address was found, the input field stayed in an empty state.

### DIFF
--- a/src/libs/suggest.js
+++ b/src/libs/suggest.js
@@ -44,9 +44,6 @@ export const selectItem = (selectedItem, { query, replaceUrl = false, fromQueryP
 };
 
 export const getInputValue = item => {
-  if (item instanceof Poi) {
-    return item.name;
-  }
   if (item instanceof Category) {
     return item.getInputValue();
   }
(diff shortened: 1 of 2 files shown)


src/libs/suggest.js
```diff
@@ -44,9 +44,6 @@ export const selectItem = (selectedItem, { query, replaceUrl = false, fromQueryP
 };
 
 export const getInputValue = item => {
-  if (item instanceof Poi) {
```
amatissart, line 47: Technically this check `item instanceof Poi` could be preserved. However it makes testing more cumbersome as some tests trigger simulated events using serialized POIs. And avoid using `instanceof` is probably preferable.

bbecquet (approved):
I think we displayed the postcode/city previously.
But we can merge we the street only as a bug fix, and see later how to leverage address util functions to normalize the string with the city if needed.

PR #875, opened 2020-10-23
Show direction title only when needed on mobile

G-Ray:
## Description
On mobile, in order to maximize useful space, we can hide the title when the panel is opened

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -330,29 +330,34 @@ export default class DirectionPanel extends React.Component {
       openMobilePreview={this.openMobilePreview}
     />;
 
+    const isFormCompleted = origin && destination;
+    const isResultDisplayed = !activePreviewRoute && isFormCompleted;
+
     return <DeviceContext.Consumer>
       {isMobile => isMobile
         ? <Fragment>
           {!activePreviewRoute && <div className="direction-panel" ref={this.setMarginTop}>
-            <Flex
+            {!isFormCompleted && <Flex
               className="direction-panel-header"
               alignItems="center"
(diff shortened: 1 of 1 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -334,13 +334,14 @@ export default class DirectionPanel extends React.Component {
       {isMobile => isMobile
         ? <Fragment>
           {!activePreviewRoute && <div className="direction-panel" ref={this.setMarginTop}>
-            <Flex
+            {(routes.length === 0 && !isLoading) && <Flex
```
bbecquet, line 337: It would be more robust to use the same display condition as the result panel, i.e. having filled up `origin` and `destination`. That way the change will happen in the same render pass.
It would also hide the title in case on an error, which is also more consistent: when you go from a 'car' result to 'public transport' and the latter doesn't have results, you don't need to see the title again.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -353,6 +354,7 @@ export default class DirectionPanel extends React.Component {
               fitContent={['default']}
               marginTop={this.marginTop}
               minimizedTitle={_('Unfold to show the results', 'direction')}
+              onClose={this.onClose}
```
bbecquet, line 357: We should check with the design but I think we shouldn't use the same `onClose`. Here what we want is probably to clear the current result but go back to the form.

src/panel/direction/DirectionPanel.jsx
G-Ray: you r right, thanks :+1:

bbecquet (approved):

PR #877, opened 2020-10-23
Fix roadmap scrolling view on mobile

bbecquet:
## Description
On Android browsers, the last item of the scrollable roadmap was unreachable.
The reason was the use of `100vh` as height for the container, which ignores the height of the browser address bar.

localhost_3000_routes__origin=latlon_48 89186_2 27712 destination=latlon_47 99849_1 69436 mode=driving pt=true

### DIFF
--- a/src/scss/includes/mobileRouteDetails.scss
+++ b/src/scss/includes/mobileRouteDetails.scss
@@ -11,7 +11,7 @@
   top: 0;
   left: 0;
   width: 100%;
-  height: 100vh;
+  height: 100%;
   z-index: 3;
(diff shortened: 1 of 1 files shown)


src/scss/includes/mobileRouteDetails.scss
```diff
@@ -47,7 +47,6 @@
   }
 
   .itinerary_roadmap {
-    height: 100%;
```
bbecquet, line 50: This one is useless as as a flex child it will fill the room available.

G-Ray (approved):

PR #879, opened 2020-10-23
Restore the layout of public transport roadmaps

bbecquet:
## Description
Some specific elements of the public transport roadmaps were lost during the redesign. This PR restore them.

*I'll probably do some CSS clean-up after this fix because diving again in this complicated part made me realized the mess it's still*

## Screenshots
|Before|After|
|---|---|
|localhost_3000_routes__origin=latlon_48 86631_2 37200 destination=pj_08721866@NOMAD_S mode=publicTransport pt=true (1)|localhost_3000_routes__origin=latlon_48 86631_2 37200 destination=pj_08721866@NOMAD_S mode=publicTransport pt=true|

### DIFF
--- a/src/panel/direction/RoadMapItem.jsx
+++ b/src/panel/direction/RoadMapItem.jsx
@@ -3,19 +3,22 @@ import classnames from 'classnames';
 
 import { Divider, Flex } from 'src/components/ui';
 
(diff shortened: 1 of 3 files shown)


G-Ray (approved):
It looks nice :+1:

PR #885, opened 2020-10-26
[Clean-up] Simplify SuggestDropdown

bbecquet:
## Description
Remove obsolete styling function of the SuggestsDropdown component, because:
 - we always pass an `outputNode` to our `<Suggest>` instances, so the `isAttachedToInput` mode was legacy
 - the `paddingBottom` rule override as no effect on the current design
 - it's possible and better to change both these points with CSS only, if needed.

As a result, we can remove two props and reduce coupling with the DOM.

## Why
First step of simplifying the `<Suggest>` component. The plan is to make it usable with a React `<TopBar>` implementation.

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -120,8 +120,6 @@ const Suggest = ({
   const SuggestsDropdownElement = () =>
     <SuggestsDropdown
       className={className}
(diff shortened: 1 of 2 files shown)


G-Ray:
Great simplification as we always use an `outputNode` !

G-Ray (approved):

PR #886, opened 2020-10-27
Fix Firefox Android geoloc hanging

bbecquet:
## Description
Firefox Android has serious problems with the geoloc calls from the direction panel (like when using "my position" from the suggest).
As far as we understand, some early calls hang indefinitely, and subsequent calls get blocked. As a result, the UI seems stuck.

It seems the `maximumAge` option creates more problems than it solves, so remove it.
And add an explicit `timeout` so that in case of a call hanging, at least the UI is unstuck.


amatissart (approved):
Even if timeouts may occur on Firefox after repeating calls to `getCurrentPosition`, it's still better than the previous behavior.

PR #888, opened 2020-10-28
QMAPS-1761 embeddable mode 

xem:
## Description
no ui / no map, marker interactions / nopadding / no bottom ui offset on mobile
when url contains &no_ui=1

## Why
To be iframeable on Qwant Search

## Screenshots
image
image


src/adapters/scene.js
```diff
@@ -446,6 +459,9 @@ Scene.prototype.translateUIControl = function(selector, bottom) {
 };
 
 Scene.prototype.moveMobileBottomUI = function(bottom = 0) {
+  if (window.no_ui) {
```
G-Ray, line 462: Can we directly return here ?

src/adapters/scene.js
```diff
@@ -83,6 +83,15 @@ Scene.prototype.initMapBox = async function(locationHash) {
...
   this.popup.init(this.mb);
 
+  // iframe: no interactions
+  if (window.no_ui) {
+    this.mb['scrollZoom'].disable();
+    this.mb['dragPan'].disable();
+    this.mb['doubleClickZoom'].disable();
+    this.mb['dragRotate'].disable();
+    this.mb['touchZoomRotate'].disable();
+  }
```
amatissart, line 93: Not tested, but it could be cleaner to simply add `interactive: false` in the `Map` options:
https://docs.mapbox.com/mapbox-gl-js/api/map/#map-parameters

views/index.ejs
```diff
@@ -57,7 +57,15 @@
       </form>
       <div class="search_form__result"></div>
     </div>
-    <script>window.times = {init : Date.now()}</script>
+    <script>
+        window.times = {init : Date.now()}
+
+        // iframe: hide UI immediately to avoid content flashing while the JS bundle loads
+        if (location.search.includes('ui=0')) {
+            window.no_ui = true;
```
G-Ray, line 65: maybe we should keep the same name consistent with the url query param ?
`window.ui = false`

src/adapters/scene.js
```diff
@@ -77,21 +77,12 @@ Scene.prototype.initMapBox = async function(locationHash) {
     style: getStyle(),
     hash: false,
     maxZoom: 20,
+    interactive: window.no_ui ? false : true,
```
G-Ray, line 80: :+1:

src/adapters/scene_category.js
```diff
@@ -93,8 +93,14 @@ export default class SceneCategory {
   }
 
   handleLayerMarkerClick = e => {
+    // iframe: markers are not clickable
+    if (window.no_ui) {
```
bbecquet, line 97: Same remark as previously. Not attaching the listeners in the first place would make more sense.

bbecquet (commented):
The `getMapCenterOffset` also has to be modified in ` src/panel/layouts.js`, so the single POI view is centered.
Now it looks like this ([url](http://localhost:3000/place/pj:08770654@Kyoto_Sushi?no_ui=1))
localhost_3000_place_pj_08770654@Kyoto_Sushi_no_ui=1

src/scss/includes/app.scss
```diff
@@ -101,3 +101,7 @@ noscript {
     font-size: 24px;
   }
 }
+
+.no_ui .panel_container {
```
G-Ray, line 105: Can we define all the hidden classes inside à `no_ui.scss` file for instance ?

views/index.ejs
```diff
@@ -57,7 +57,15 @@
       </form>
       <div class="search_form__result"></div>
     </div>
-    <script>window.times = {init : Date.now()}</script>
+    <script>
+        window.times = {init : Date.now()}
+
+        // iframe: hide UI immediately to avoid content flashing while the JS bundle loads
+        if (location.search.includes('no_ui=1')) {
```
amatissart, line 64: [`.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility) is not available on IE11.  
Moreover, using such a lookup may be fragile, and could be exploited (what about `?q=nagano_ui=1`).

I think that it would be cleaner and more robust to check this paramater on server side, using [`req.query`](https://expressjs.com/fr/api.html#req.query )provided by Express.

xem: Simpler / more accurate JS version: search.indexOf("&no_ui=1") != -1
Do you think it's not enough?
If not, where would you put such code if we did it in Express?
@amatissart

src/scss/includes/app.scss
xem: sure

src/adapters/scene.js
```diff
@@ -131,6 +131,10 @@ Scene.prototype.initMapBox = async function(locationHash) {
     // which are thrown *after* two separate click events are thrown
     this.clickDelayHandler = null;
     this.mb.on('click', e => {
+      // iframe: disable clicks
+      if (window.no_ui) {
```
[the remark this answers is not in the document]
xem: I agree, good idea!

views/index.ejs
amatissart: Then it would mean the `no_ui` cannot be used as the first query param in the URL (`?no_ui=1&...`), whereas most applications would expect that the order should not matter. And there are possibly other encoding issues I am not aware of. That's why I would still prefer to use a well-tested implementation from existing libraries for such URL manipulation.

In Express, we could add a `noUi` variable to the context used by the .ejs template:
```diff
diff --git a/bin/app.js b/bin/app.js
index ce6451ad..d753ad28 100644
--- a/bin/app.js
+++ b/bin/app.js
@@ -116,7 +116,7 @@ function App(config) {
       appConfig = JSON.parse(JSON.stringify(config));
       appConfig.burgerMenu.enabled = false;
     }
-    res.render('index', { config: appConfig });
+    res.render('index', { config: appConfig, noUi: req.query.no_ui === '1' });
   });
 
   if (config.server.acceptPostedLogs) {
```
Then the JS condition can be replaced with .ejs syntax:
```
<% if (noUi) { %>
// your script
<% } %>
```

xem: ah very nice indeed, thanks

views/index.ejs
```diff
@@ -57,7 +57,18 @@
...
       <div class="search_form__result"></div>
     </div>
-    <script>window.times = {init : Date.now()}</script>
+    <script>
+      window.times = {init : Date.now()}
+    </script>
+    <% if (noUi) { %>
+    <script>
+      // iframe: hide UI immediately to avoid content flashing while the JS bundle loads
+      if (location.search.indexOf('&no_ui=1') != -1) {
```
amatissart, line 66: this `if` clause can be removed now that the check is done on server side.

xem: oops, I overlooked that, sorry, fixed!

src/panel/layouts.js
```diff
@@ -54,8 +61,9 @@ export function getVisibleBbox(mb) {
   return bbox;
 }
 
-export function getMapCenterOffset({ isMobile }) {
-  return isMobile ? [0, 0] : [(DESKTOP_PANEL_WIDTH + ADDITIONAL_PADDING) / 2, 0];
+// Disable offset on mobile and in iframe mode
```
bbecquet, line 64: It's strange to have this comment here whereas the code deciding that is elsewhere, in `scene.js`. It would be consistent with the other functions to receive `isIframe` here alongside `isMobile` as param.

src/adapters/scene.js
```diff
@@ -130,31 +130,35 @@ Scene.prototype.initMapBox = async function(locationHash) {
...
-      // as a subsequent double click isn't a problem
-      if (pois[0]) {
-        this.clickOnMap(e.lngLat, pois[0]);
-        return;
-      }
-      this.clickDelayHandler = setTimeout(() => {
-        // for touch UX we have to make sure a double tap zoom hasn't been made in the meantime
-        if (Date.now() - this.lastDoubleTapTimeStamp < this.DOUBLE_TAP_DELAY_MS) {
+
+    // iframe: disable clicks on map
```
G-Ray, line 134: early return to avoid touching existing code parts ?

src/panel/layouts.js
```diff
@@ -31,20 +35,23 @@ export function getMapPaddings({ isMobile, isDirectionsActive }) {
...
-    // On mobile, compute a bbox that excludes the header's height
-    ne_canvas.y += 65;
-  } else {
-    // On desktop, compute a bbox that excludes the left panel's width
-    sw_canvas.x += DESKTOP_PANEL_WIDTH + ADDITIONAL_PADDING / 2;
+  // iframe: no offset
+  if (!isIframe) {
+    if (isMobileDevice()) {
+      // On mobile, compute a bbox that excludes the header's height
+      ne_canvas.y += 65;
```
G-Ray, line 50: can we define an constant like for desktop ?

src/scss/includes/no_ui.scss
```diff
@@ -0,0 +1,18 @@
+// Iframe mode: hide all UI elements except map, markers and attribution
+.no_ui {
```
G-Ray, line 2: :+1:

src/panel/layouts.js
xem: Yes, I did that then thought it was a bit absurd to test both isMobile && isIframe for the same outcome ( [0,0] ).
But ok!

src/panel/layouts.js
xem: to replace the 65? yes, though it may be off-topic for this PR, no? (it was already like that)

src/adapters/scene.js
xem: an early return would cancel all the following code including mouse/finger interactions (that's OK) but also the final lines:
```
    window.execOnMapLoaded = f => f();
    fire('map_loaded');
```
I think we can't cut these.

src/panel/layouts.js
G-Ray: ah, yes ok

bbecquet (approved):

PR #889, opened 2020-10-29
Display itinerary roadmap start/end points with POI type icons

bbecquet:
## Description
 - Introduce a new `PlaceIcon` component to illustrate place types in a design-compliant way (it would be possible to re-use it when redesigning the geocoding suggestions)
 - Use this component in itinerary roadmaps in place of generic origin/destination pictos

*The PR diff is big and intimidating because I had to change a lot of small parts and I split some existing stuff into smaller parts, but they should make maintenance and further dev easier. And globally I don't do anything complicated, promised!* :smile:

## Screenshots
|Before|After|
|---|---|
|Capture d’écran 2020-10-29 à 16 09 19|Capture d’écran 2020-10-29 à 16 09 40|
|Capture d’écran 2020-10-29 à 16 13 49|Capture d’écran 2020-10-29 à 16 13 19|


src/components/PlaceIcon.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
G-Ray, line 1: :+1:

G-Ray (approved):

PR #890, opened 2020-10-30
Display share button only if isResultDisplayed

G-Ray:


tests/integration/tests/direction.js
```diff
@@ -67,11 +67,14 @@ test('simple search', async () => {
...
   responseHandler.addPreparedResponse(mockMapBox, /\/30\.0000000,5\.0000000;30\.0000000,5\.0000000/);
   await page.goto(`${APP_URL}/${ROUTES_PATH}`);
+  expect(await isHidden(page, '.direction-panel-share-button')).toBeTruthy();
   await page.waitForSelector('#direction-input_origin');
   await page.type('#direction-input_origin', 'direction');
   await page.keyboard.press('Enter');
   await page.type('#direction-input_destination', 'direction');
   await page.keyboard.press('Enter');
+  expect(await exists(page, '.direction-panel-share-button')).toBeTruthy();
+  await page.waitForSelector('.direction-panel-share-button', { visible: true });
```
bbecquet, line 77: You could just `expect` that the return of that is truthy instead of having two separate lines.

bbecquet (approved):

PR #894, opened 2020-10-30
Add labels on alternative route paths

bbecquet:
## Description
Introduce a custom lib to compute good placement for labels on routes. This is something I did some years ago, a bit reworked. It works on GeoJSON linestrings and make use of the [Turf](https://turfjs.org/) library for some operations.

The idea is to place labels on parts of the alternatives that are unique to each. So, most of the algorithm deals with detecting these unique segments. The implementation uses a count of the occurence of each coordinate accross all lines. The segments are lists of contiguous coordinates that are found only once. Doing it in place is far more complicated, as there are many edge cases (conceptually, this is a graph solving problem, but maybe it's not worth diving in graph algos for that).

We also optimize the orientation of labels so that they don't overlap too much on the route line: we display on the left or right side if the line is locally somewhat vertical, and above or below if it near horizontal.

Left/right and above/below are decided by making the labels try to "get away" from each others, using the bearing of the other closest one as an indication of the overall direction.

Most of the time the current solution is good, but as always with cartography, there are some not-ideal cases.

---

I use this lib to display labels on our alternatives as div Mapbox markers.
The CSS is straightforward, except for the part implementing the tooltips pointed end, where I used the [triangle border trick](https://css-tricks.com/snippets/css/css-triangle/).

## Screenshots
Capture d’écran de 2020-11-03 12-15-09
Capture d’écran de 2020-11-03 12-13-08


amatissart (commented):
Looks good to me.  
I made a few comments mainly about the manipulation of floats.

src/libs/route_labeler.js
```diff
@@ -0,0 +1,128 @@
...
+  let end = coordinates.length - 1;
+
+  // a distinct segment is a part of a line where coordinates
+  // appear only once accross all the features
+  for (let index = 0; index < coordinates.length; index++) {
+    const coord = coordinates[index];
+    if (start === 0 && coordCounts.get(asKey(coord)) === 1) {
+      start = index;
+    }
+    if (start !== 0 && end === coordinates.length - 1 && coordCounts.get(asKey(coord)) !== 1) {
```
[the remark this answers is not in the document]
bbecquet: Indeed! Thank you for reading the algo so thoroughly :)

src/libs/route_labeler.js
```diff
@@ -0,0 +1,128 @@
...
+function getAxis(ls) {
+  const segmentCoords = getCoords(ls);
+  const segmentBearing = bearing(segmentCoords[0], segmentCoords[segmentCoords.length - 1]);
+  if (Math.abs(segmentBearing) < 45 || Math.abs(segmentBearing) > 135) {
+    return 'vertical';
+  }
+  return 'horizontal';
+}
+
+const asKey = coord => `${coord[0]},${coord[1]}`;
```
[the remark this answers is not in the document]
bbecquet: No, `coord`, is the raw coordinate coming from the GeoJSON, we don't know its precision in advance.

I don't think there would be any issues relative to float manipulation, as I compare and stringify these raw values without doing any math operations on them, but you're right, let's make that more robust.

amatissart (approved):

src/scss/includes/routeLabel.scss
```diff
@@ -0,0 +1,118 @@
...
+
+    &.active::after {
+      border-bottom-color: white;
+    }
+  }
+
+  &--right {
+    left: -6px;
+
+    &::after {
```
G-Ray, line 106: can we factorize these properties using a local `mixin` :lipstick:  ?

bbecquet: I tried for some time, I think it's feasible but this is really not easy, as Sass is limited in its management of variables and you can't do it with only one var.
IMO it would be a good improvement for later :)

src/adapters/scene_direction.js
```diff
@@ -4,22 +4,46 @@ import { normalizeToFeatureCollection } from 'src/libs/geojson';
...
 import { fire, listen } from 'src/libs/customEvents';
+import { getLabelPositions } from 'src/libs/route_labeler';
 
 const createMarker = (lngLat, className = '', options = {}) => {
   const element = document.createElement('div');
   element.className = className;
   return new Marker({ ...options, element }).setLngLat(lngLat);
 };
 
+const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
```
G-Ray, line 26: this function would best fit in `route_labeler.js`, what do you think ?

src/libs/route_labeler.js
```diff
@@ -0,0 +1,131 @@
...
+// whose coordinates don't overlap with another feature
+export function findDistinctSegments(linestrings) {
+  if (linestrings.length < 2) {
+    return linestrings;
+  }
+  // extract raw coordinates
+  const featuresCoords = linestrings.map(ls => dropRepeatedCoords(coordAll(ls)));
+  // count occurences of each coordinate accross all features
+  const coordCounts = new Map();
+  [].concat(...featuresCoords).forEach(coord => {
```
G-Ray, line 62: [...featuresCoords] :innocent: ?

bbecquet: Nope, not the same, what I want is a `flat` operation, not a copy :)

src/libs/route_labeler.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+  return result;
+}
+
+// Reduce possibilities of collision by chosing anchors so that labels repulse each other
+function optimizeAnchors(positions) {
+  return positions.map((position, index) => {
+    const others = positions.slice();
+    others.splice(index, 1);
```
G-Ray, line 89: we can directly do
```js
const others = positions.splice(index, 1)
```
without manipulating any array in place.

src/libs/route_labeler.js
```diff
@@ -0,0 +1,131 @@
...
+      result.push(list[i]);
+    }
+  }
+  return result;
+}
+
+// Reduce possibilities of collision by chosing anchors so that labels repulse each other
+function optimizeAnchors(positions) {
+  return positions.map((position, index) => {
+    const others = positions.slice();
```
G-Ray, line 88: we could avoid manipulating any array in place with 
```js
const others = [ 
  ...positions.slice(0, index),
  ...positions.slice(index + 1, positions.length)
]
```

src/libs/route_labeler.js
G-Ray: ah yes sorry, I misread. I guess we can't use`Array.prototype.flat()` because of IE ?

bbecquet: That's it.

src/adapters/scene_direction.js
bbecquet: I don't know, I wrote this function of the model of `createMarker` above and the idea was to separate the work of the lib on raw stuffs from the MapBox dependency and design-dependant markup.

src/libs/route_labeler.js
```diff
@@ -0,0 +1,133 @@
...
+  return 'horizontal';
+}
+
+const asKey = coord => `${coord[0].toFixed(6)},${coord[1].toFixed(6)}`;
+
+function distinctSegment(coordinates, coordCounts) {
+  let start = 0;
+  let end = coordinates.length - 1;
+
+  // a distinct segment is a part of a line where coordinates
```
G-Ray, line 36: Not sure, but is this the same as : 
```js
const start = coordinates.findIndex((coord) => (coordCounts.get(asKey(coord)) === 1)) )
const end = coordinates.findIndex((coord) => (coordCounts.get(asKey(coord)) !== 1)) )
```

src/libs/route_labeler.js
```diff
@@ -0,0 +1,133 @@
...
+  // appear only once accross all the features
+  for (let index = 0; index < coordinates.length; index++) {
```
G-Ray, line 38:
```js
const start = coordinates.findIndex((coord) => (coordCounts.get(asKey(coord)) === 1)) )
const end =  ...coordinates.slice(start, coordinates.length).findIndex((coord) => (coordCounts.get(asKey(coord)) !== 1)) )
```

src/libs/route_labeler.js
```diff
@@ -0,0 +1,124 @@
...
+const toSimpleLinestring = geoJson => lineString(coordAll(geoJson));
+
+const TOLERANCE = 0.000001;
+const floatEquals = (f1, f2) => Math.abs(f1 - f2) < TOLERANCE;
+
+const coordEquals = (c1 = [], c2 = []) => floatEquals(c1[0], c2[0]) && floatEquals(c1[1], c2[1]);
+
+function dropRepeatedCoords(list) {
+  const result = [];
+  for (let i = 0; i < list.length; i++) {
```
G-Ray, line 68:
```js
const result = list.filter((item) => 
  !coordEquals(result[result.length - 1], item))
```

G-Ray (approved):
:clap:

PR #895, opened 2020-11-02
buildQueryString: decode URI + test

G-Ray:
## Description
When building a query string with `buildQueryString`,  decode URI components to get human readable urls, like in the actual Maps release.

## Why
previsouly, buildQueryString would return an encoded representation of a string like `admin:osm:relation:71525@Paris`.

### DIFF
--- a/src/libs/url_utils.js
+++ b/src/libs/url_utils.js
@@ -65,3 +65,10 @@ export function buildQueryString(queriesObject) {
   const params = new URLSearchParams(removeNullEntries(queriesObject)).toString();
   return params ? `?${params}` : '';
 }
+
+export function buildRelativeUrl({
+  pathname = window.location.pathname,
+  search = window.location.search,
+}) {
+  return joinPath([pathname, search]);
+}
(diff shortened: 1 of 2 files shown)


bbecquet (commented):
As `buildQueryString` is meant to build url-safe query strings, the encoding done by `URLSearchParams` is a feature. In what case do you want decoded values?

G-Ray:
> As `buildQueryString` is meant to build url-safe query strings, the encoding done by `URLSearchParams` is a feature. In what case do you want decoded values?

In order to keep the actual behavior.
Take for instance this url: https://www.qwant.com/maps/routes/?origin=admin:osm:relation:7444@Paris&mode=driving#map=11.98/48.8591710/2.3129699

urlSearchParams.toString() would return `origin=admin%3Aosm%3Arelation%3A7444%40Paris&mode=driving%23map%3D11.98%2F48.8591710%2F2.3129699"`

bbecquet:
> In order to keep the actual behavior.

> would return `origin=admin%3Aosm%3Arelation%3A7444%40Paris&mode=driving%23map%3D11.98%2F48.8591710%2F2.3129699"`

Sorry, I still don't see the problem and what behavior wouldn't work with that.
`buildQueryString` is used to build URL parts, so it's is job to encode strings safely. If you want to rely on the URL values later, I understand you'd like have to decode them, but it should be done when reading it, like in `parseQueryString`.
But I guess I'm missing the point.

PR #896, opened 2020-11-02
DirectionPanel: Improve url construction

G-Ray:
## Description
- In DirectionPanel, use functions from `src/libs/url_utils`.

 From now, urls params will not be overrided by DirectionPanel as we use `parseQueryString`. This is needed for future work, where we need to keep a `selected` query params which will be handled by `RouteResult`

Urls should now be built with encoded params. We need to be sure this has no negative impact accross the app.

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -22,6 +22,7 @@ import { getInputValue } from 'src/libs/suggest';
 import { geolocationPermissions, getGeolocationPermission } from 'src/libs/geolocation';
 import { openPendingDirectionModal } from 'src/modals/GeolocationModal';
 import ShareMenu from 'src/components/ui/ShareMenu';
+import { parseQueryString, buildQueryString } from 'src/libs/url_utils';
 
 const MARGIN_TOP_OFFSET = 64; // reserve space to display map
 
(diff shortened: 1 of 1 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -222,18 +223,18 @@ export default class DirectionPanel extends React.Component {
...
+      mode: this.state.vehicle,
+      origin: this.state.origin ? poiToUrl(this.state.origin) : null,
+      destination: this.state.destination ? poiToUrl(this.state.destination) : null,
+      pt: this.props.isPublicTransportActive ? 'true' : null,
+    };
+
+    const search = buildQueryString(queryObject);
+    const relativeUrl = buildRelativeUrl({ search });
+
+    window.app.navigateTo(relativeUrl, {}, {
```
amatissart, line 237: `buildRelativeUrl` uses the current pathname, possibly prefixed with `/maps/`, while `navigateTo` does not support this prefix and expects the application route only (eg `/routes` without prefix)

bbecquet: I think we can do without `buildRelativeUrl`. It is ok for the component to now what hardcoded route it needs.

bbecquet (approved):

PR #898, opened 2020-11-03
Direction: handle selected state as search param

G-Ray:
## Description
- Add `updateQueryString(queriesObject)` to url_utils.js : it applies a diff to the current search params.
- Handle selected route as a search param (`selected`). We don't need any state for this in `RouteResult` anymore.
- Sanitize selected search param in `DirectionPanel`. It would be even better to update the url if the user entered an invalid value like `selected=randomstring`, but we think it would be overcomplex for now.

## Why
Ongoing work to properly handle routing in Direction Panel

## Screenshots
image
(note the selected search params)

Peek 2020-11-03 16-53
Here, we can see the last selected route is restored from history.

### DIFF
--- a/src/panel/direction/RouteResult.jsx
+++ b/src/panel/direction/RouteResult.jsx
@@ -18,7 +18,7 @@ export default class RouteResult extends React.Component {
     isLoading: PropTypes.bool,
     error: PropTypes.number,
     openMobilePreview: PropTypes.func.isRequired,
-    activeRouteId: PropTypes.number.isRequired,
+    activeRouteId: PropTypes.number,
   }
 
   static defaultProps = {
@@ -35,6 +35,12 @@ export default class RouteResult extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.props.routes.length !== prevProps.routes.length) {
+      this.updateUrl({ selected: this.props.activeRouteId });
+    }
+  }
+
   selectRoute = routeId => {
     if (routeId === this.props.activeRouteId) {
       return;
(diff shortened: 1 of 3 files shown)


src/panel/direction/RouteResult.jsx
```diff
@@ -34,36 +34,32 @@ export default class RouteResult extends React.Component {
...
   }
 
-  componentDidUpdate(prevProps) {
-    if (this.props.routes.length !== prevProps.routes.length) {
-      this.setState({ activeRouteId: 0 });
-    }
-  }
-
   selectRoute = routeId => {
-    if (routeId === this.state.activeRouteId) {
```
bbecquet, line 44: Don't you think `activeRouteId` was a better name than `selected`?

G-Ray: the current route is `routes`, so "selected" is explicit enough to me. However I would choose another name if you think it's more explicit. Would you prefer `activeRouteId=0` @bbecquet @xem ?

bbecquet: I'm not worried about the clarity of the url, but the clarity of the code behind. In a component with an already complicated lifecycle, having a state value named simply `selected` isn't explicit enough to my taste.
If you want to keep `selected` in the url, it's possible to rename it on-the-fly when parsing the params in the route definition, while keeping something more clear as state/prop.

src/panel/direction/RouteResult.jsx
```diff
@@ -34,36 +35,32 @@ export default class RouteResult extends React.Component {
...
+    if (routeId === this.props.activeRouteId) {
       return;
     }
+
     Telemetry.add(Telemetry.ITINERARY_ROUTE_SELECT);
     fire('set_main_route', { routeId, fitView: true });
-    this.setState({ activeRouteId: routeId });
+
+    const search = updateQueryString({ selected: routeId });
+    window.app.navigateTo('routes/' + search, {}, { replace: true });
```
bbecquet, line 47: This method doesn't get called when first rendering the results, so the url isn't updated unless you chose another alternative.
I wonder if this can be a problem. It may signal some inconsistency that could prove problematic later. What do you think?

G-Ray: I'm in favor of trying to keeps things consistent. I updated the PR with an url update triggered in `didUpdate` lifecycle.

bbecquet (approved):

PR #899, opened 2020-11-03
Introduce spacing values from the Design System

bbecquet:
## Description
Declare standard spacing values used by the Qwant Design System and used everywhere accross the screens provided by the design team.
They are supposed to be used in place of hard-coded values for margins or paddings, and will help us define some standard layout components.
For now I replaced our existing margin utils (like `.u-mb-8`) by the full scope of sizes and made the substitution where they were used in the app.

What do you think @G-Ray and @xem?


xem:
great!
I would rename "xss" into 'xxs"

bbecquet:
@xem 

> I would rename "xss" into 'xxs"

Oops, indeed, that's a typo! Thanks for your keen eye :)

src/scss/includes/utils.scss
```diff
@@ -80,42 +80,19 @@
...
-  margin-right: 4px;
-}
-
-.u-mr-8 {
-  margin-right: 8px;
-}
-
-.u-mr-16 {
-  margin-right: 16px;
+@each $spacingName, $spacingValue in $spacings {
```
G-Ray, line 83: clever until we migrate our existing classes usages !

G-Ray (approved):

PR #900, opened 2020-11-04
QMAPS-1761 don't restore last location on load in iframe mode

xem:
## Description
- don't restore last location on load in iframe mode (&no_ui=1)
- don't save last location in iframe mode

can we test it on dev?

## Why
To avoid thys flyTo when loading the IA
fly

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -208,15 +208,15 @@ Scene.prototype.initMapBox = async function(locationHash) {
     this.mb.on('dragstart', () => { fire('map_user_interaction'); });
     this.mb.on('pitchstart', () => { fire('map_user_interaction'); });
 
-    if (!window.no_ui) {
-      this.mb.on('moveend', () => {
-        const { lng, lat } = this.mb.getCenter();
-        const zoom = this.mb.getZoom();
+    this.mb.on('moveend', () => {
+      const { lng, lat } = this.mb.getCenter();
+      const zoom = this.mb.getZoom();
+      if (!window.no_ui) {
         store.setLastLocation({ lng, lat, zoom });
-        window.app.updateHash(this.getLocationHash());
-        fire('map_moveend');
-      });
-    }
+      }
+      window.app.updateHash(this.getLocationHash());
+      fire('map_moveend');
+    });
 
     window.execOnMapLoaded = f => f();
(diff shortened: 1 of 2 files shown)


src/panel/poi/PoiPanel.jsx
```diff
@@ -192,7 +192,9 @@ export default class PoiPanel extends React.Component {
     const uri = `/places/${params}`;
 
     Telemetry.add(Telemetry.POI_BACKTOLIST);
-    fire('restore_location');
+    if (!window.no_ui) {
+      fire('restore_location');
+    }
```
amatissart, line 197: I am not sure to understand how this change is supposed to solve the problem you are describing.  
This function `backToList` is used only when clicking on the button at the top of the PoiPanel after an item has been selected from a list. Or am I missing something ?

amatissart:
> since pan/zoom is not available in this mode, the app can't save the last visited position,

The map position will be saved anyway if the query triggers a map move (e.g "restaurant à Nice" which requires the map to be centered on the detected zone)

src/panel/poi/PoiPanel.jsx
xem: Ah, let me check, there may be a mistake as there are 2 different "getLastLocation()" functions in the project =_=

bbecquet (approved):

src/adapters/scene.js
```diff
@@ -208,13 +208,15 @@ Scene.prototype.initMapBox = async function(locationHash) {
...
-      window.app.updateHash(this.getLocationHash());
-      fire('map_moveend');
-    });
+    if (!window.no_ui) {
+      this.mb.on('moveend', () => {
+        const { lng, lat } = this.mb.getCenter();
+        const zoom = this.mb.getZoom();
+        store.setLastLocation({ lng, lat, zoom });
+        window.app.updateHash(this.getLocationHash());
+        fire('map_moveend');
```
amatissart, line 217: This `map_moveend` event is used by the `CategoryPanel` to trigger the search for POIs in the current zone after the intention has been derived from the query and display the markers.  
So it should still be fired. As is no marker would be visible on the map.

amatissart (changes requested):

amatissart (approved):

PR #901, opened 2020-11-05
[Clean-up] Remove store abstraction

bbecquet:
## Description
Remove the abstraction level on store, which was used to expose local storage or Masq storage with the same API. Now `store.js` is the localStorage-based implementation.
Also uses a proper singleton implementation instead of using global vars.

*Another simplification would be to remove all these `async/await`, as the localstorage API is synchronous.*

## Why
Since the removal of Masq, the abstraction was useless. Let's remove some code complexity and shipped size.


src/adapters/store.js
```diff
@@ -1,93 +1,111 @@
...
 import { getKey } from 'src/libs/pois';
 import { fire } from 'src/libs/customEvents';
+import { isPoiCompliantKey } from 'src/libs/pois';
 
-export default class Store {
-  constructor() {
-    // get store from window if already initialized
-    if (window.__store) {
-      return window.__store;
+class Store {
```
G-Ray, line 8: :+1
What do you think about even removing the class and exporting only functions ?

bbecquet: Yep, even simpler than a singleton, but I wanted to keep this PR with minimal changes outside the store file itself.
Do you want me to make the change here?

G-Ray: would be great !

bbecquet: Sure, doing it!

G-Ray (approved):

G-Ray (approved):

PR #908, opened 2020-11-10
QMAPS-1833 - Reimplement modal on geolocate control trigger

G-Ray:
## Description
- Implement Geolocate modal logic in extended_geolocate_control (credits go to @amatissart)
- Update title, text, and padding
- Ask for location permissions even if user has dismissed the modal
- Store a boolean in LocalStorage to open the modal only once

## Why
New UX and text when the user clicks on the geolocate button.

## Screenshots
image

### DIFF
--- a/src/mapbox/extended_geolocate_control.js
+++ b/src/mapbox/extended_geolocate_control.js
@@ -1,5 +1,6 @@
 import * as Geolocation from '../libs/geolocation';
 import Telemetry from '../libs/telemetry';
+import { openPendingGeolocateModal } from 'src/modals/GeolocationModal';
 
 import { GeolocateControl } from 'mapbox-gl--ENV';
 
@@ -27,6 +28,13 @@ export default class ExtendedGeolocateControl extends GeolocateControl {
   }
 
   async trigger() {
+    const state = await Geolocation.getGeolocationPermission();
+    if ( state === Geolocation.geolocationPermissions.PROMPT ) {
+      const isModalAccepted = await openPendingGeolocateModal();
+      if (!isModalAccepted) {
+        return;
+      }
+    }
     super.trigger();
   }
 
--- a/src/mapbox/extended_geolocate_control.js
+++ b/src/mapbox/extended_geolocate_control.js
@@ -1,8 +1,10 @@
 import * as Geolocation from '../libs/geolocation';
 import Telemetry from '../libs/telemetry';
 import { openPendingGeolocateModal } from 'src/modals/GeolocationModal';
+import { get } from 'src/adapters/store';
 
 import { GeolocateControl } from 'mapbox-gl--ENV';
+import { set } from '../adapters/store';
 
 /**
 * Override default GeolocateControl
@@ -29,8 +31,10 @@ export default class ExtendedGeolocateControl extends GeolocateControl {
 
   async trigger() {
     const state = await Geolocation.getGeolocationPermission();
-    if (state === Geolocation.geolocationPermissions.PROMPT) {
+    if (state === Geolocation.geolocationPermissions.PROMPT &&
+        !get('has_geolocate_modal_opened_once')) {
       await openPendingGeolocateModal();
+      set('has_geolocate_modal_opened_once', true);
     }
     super.trigger();
   }
--- a/src/mapbox/extended_geolocate_control.js
+++ b/src/mapbox/extended_geolocate_control.js
@@ -29,11 +29,8 @@ export default class ExtendedGeolocateControl extends GeolocateControl {
 
   async trigger() {
     const state = await Geolocation.getGeolocationPermission();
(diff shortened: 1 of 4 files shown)


src/mapbox/extended_geolocate_control.js
```diff
@@ -27,6 +28,14 @@ export default class ExtendedGeolocateControl extends GeolocateControl {
   }
 
   async trigger() {
+    const state = await Geolocation.getGeolocationPermission();
+    if (state === Geolocation.geolocationPermissions.PROMPT &&
+        window.localStorage.getItem('hasGeolocateModalOpenedOnce') === null
```
G-Ray, line 33: is this key explicit enough ?

amatissart: That's probably ok. But it should be namespaced, to avoid confusion or conflict with other Qwant products.  
All interactions with localStorage are currently implemented in "store.js", and I think this logic should be moved there, where the prefix `qmaps_v1` is used in all keys.

G-Ray: You r right @amatissart , I refactored store.js a bit to help us reusing the get/set/del functions, and update this pr to use the store. Thanks

src/modals/GeolocationModal.jsx
```diff
@@ -10,15 +10,30 @@ let hasPermissionModalOpenedOnce = false;
...
+
+  const pendingText = _('We look at your location to show you where you are, and that\'s it! (See our {privacyPolicyLink}our privacy policy{closeTag})',
+    'geolocation', {
+      privacyPolicyLink: '<a target="_blank" rel="noopener noreferrer" href="https://about.qwant.com/legal/privacy">',
+      closeTag: '</a>',
+    });
+
   const statuses = {
     PENDING: {
+      title: _('At Qwant, your travels are your private life', 'geolocation'),
```
amatissart, line 29: What about ?
```suggestion
      title: _('At Qwant, your whereabouts are part of your privacy', 'geolocation'),
```

src/mapbox/extended_geolocate_control.js
```diff
@@ -1,7 +1,10 @@
 import * as Geolocation from '../libs/geolocation';
 import Telemetry from '../libs/telemetry';
+import { openPendingGeolocateModal } from 'src/modals/GeolocationModal';
+import { get } from 'src/adapters/store';
 
 import { GeolocateControl } from 'mapbox-gl--ENV';
+import { set } from '../adapters/store';
```
bbecquet, line 7: You can merge this line with the one above. Also, it could be useful to give these two functions some more explicit names locally.

src/modals/GeolocationModal.jsx
```diff
@@ -88,4 +103,20 @@ export async function openPendingDirectionModal() {
   });
 }
 
+export async function openPendingGeolocateModal() {
+  return new Promise(resolve => {
+    open(
+      'PENDING',
+      () => {
+        close();
+        resolve(false);
```
bbecquet, line 112: I think I saw in the designs that closing this modal here was equivalent to the `Continue` button. The principle is that the action was initiated by the user so they won't cancel it here.
You should check the spec for that.

src/scss/includes/modals/index.scss
```diff
@@ -32,7 +32,7 @@
 }
 
 .modal__maps__content {
-  padding: 0 12px;
+  padding: 0 $spacing-l;
```
bbecquet, line 35: Yeah :+1:

src/modals/GeolocationModal.jsx
```diff
@@ -10,15 +10,30 @@ let hasPermissionModalOpenedOnce = false;
...
   /* eslint-disable max-len */
+
+  const pendingOnDirectionsText = _('Always respecting your privacy.<br>As stated in {privacyPolicyLink}our privacy policy{closeTag}, we don\'t store your information because we don\'t want to know your whereabouts.',
+    'geolocation', {
+      privacyPolicyLink: '<a target="_blank" rel="noopener noreferrer" href="https://about.qwant.com/legal/privacy">',
+      closeTag: '</a>',
+    }
+  );
+
+  const pendingText = _('We look at your location to show you where you are, and that\'s it! (See our {privacyPolicyLink}our privacy policy{closeTag})',
```
bbecquet, line 21: The "our" word is duplicated before the link.

src/modals/GeolocationModal.jsx
G-Ray: yep, however, the only usage of `openPendingGeolocateModal` (for now) ignore the returned value (accepted if true, closed if false). So this is the current behavior :)

src/mapbox/extended_geolocate_control.js
G-Ray: thanks

bbecquet (approved):

PR #909, opened 2020-11-10
Make the PJ partnership footer visible on default mobile panel

bbecquet:
## Description
Ensure the PJ partnership footer is visible on the default mobile panel size.
Previously we encountered the now-classic bug of a `position: fixed` element (the footer) inside a `transform: translate(…)` element (the panel), which made impossible the footer not positionned relative to the viewport, but to the panel.
Like some other occurences of this problem, we use a React Portal so the footer isn't a DOM child of the Panel.

## Screenshots
|Before|After|
|---|---|
|Screenshot_20201110-162756_Chrome|Screenshot_20201110-162724_Chrome|

### DIFF
--- a/src/panel/category/CategoryPanel.jsx
+++ b/src/panel/category/CategoryPanel.jsx
@@ -5,6 +5,7 @@ import Panel from 'src/components/ui/Panel';
 import PoiItemList from './PoiItemList';
 import PoiItemListPlaceholder from './PoiItemListPlaceholder';
 import CategoryPanelError from './CategoryPanelError';
+import PJPartnershipFooter from './PJPartnershipFooter';
 import Telemetry from 'src/libs/telemetry';
 import nconf from '@qwant/nconf-getter';
 import IdunnPoi from 'src/adapters/poi/idunn_poi';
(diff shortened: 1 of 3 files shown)


src/panel/category/CategoryPanel.jsx
```diff
@@ -5,6 +5,7 @@ import Panel from 'src/components/ui/Panel';
 import PoiItemList from './PoiItemList';
 import PoiItemListPlaceholder from './PoiItemListPlaceholder';
 import CategoryPanelError from './CategoryPanelError';
+import PJPartnershipFooter from './PJPartnershipFooter';
```
G-Ray, line 8: :+1:

src/panel/category/PJPartnershipFooter.jsx
```diff
@@ -0,0 +1,24 @@
...
+  {_('Results in partnership with PagesJaunes', 'categories')}
+</div>;
+
+const PJPartnershipFooter = ({ isMobile }) => {
+  const portalContainer = useRef(document.createElement('div'));
+
+  useEffect(() => {
+    document.body.appendChild(portalContainer.current);
+    return () => {
+      document.body.removeChild(portalContainer.current);
```
G-Ray, line 15: :+1:

src/panel/category/PJPartnershipFooter.jsx
```diff
@@ -0,0 +1,24 @@
...
+    };
+  }, [isMobile]);
```
G-Ray, line 17: :100:

G-Ray (approved):

PR #911, opened 2020-11-12
Truncate POI name when displayed in list

bbecquet:
## Description
Truncate POI names when displayed in result lists, to uniformize the height of items.

## Screenshots
|Before|After|
|---|---|
|localhost_3000_places__type=restaurant(iPhone 6_7_8 Plus) (3)|localhost_3000_places__type=restaurant(iPhone 6_7_8 Plus) (2)|

### DIFF
--- a/src/components/PoiTitle.jsx
+++ b/src/components/PoiTitle.jsx
@@ -1,10 +1,11 @@
 /* global _ */
 import React from 'react';
+import classnames from 'classnames';
 import poiSubClass from 'src/mapbox/poi_subclass';
 import { capitalizeFirst } from 'src/libs/string';
 import Address from 'src/components/ui/Address';
 
-const PoiTitle = ({ poi, withAlternativeName }) => {
+const PoiTitle = ({ poi, withAlternativeName, inList }) => {
   const { name, localName, subClassName, address } = poi;
 
   // LatLon PoI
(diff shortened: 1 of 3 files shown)


src/scss/includes/utils.scss
```diff
@@ -14,6 +14,12 @@
   text-transform: uppercase;
 }
 
+.u-ellipsis {
```
G-Ray, line 17: :+1:

src/components/PoiTitle.jsx
```diff
@@ -37,7 +38,9 @@ const PoiTitle = ({ poi, withAlternativeName }) => {
 
   // Location / address
   return <div className="poiTitle">
-    <h2 className="poiTitle-main u-text--heading6">{title || subclass}</h2>
+    <h2 className={classnames('poiTitle-main u-text--heading6', { 'u-ellipsis': inList })}>
```
G-Ray, line 41: great, but shouldn't we apply the u-ellipsis class without inList condition ?

bbecquet: Nope, as we want the full name of the POI to be available when displaying the details of a single POI.
The ellipsis is here on lists so no POI takes too much vertical place compared to the others, for regularity and visual balance.

G-Ray (approved):

PR #912, opened 2020-11-12
store: refactor store.js with prefix handling

G-Ray:
## Description
- remove unnecessary async keyword
- Use prefix in get/set functions. This way, no user of store.js should set it explicitly.
- Add `del` function
- Add listKeys function

## Why
Simplify api. store can be used to store any value in LocalStorage with the correct maps/version prefix.

### DIFF
--- a/src/adapters/store.js
+++ b/src/adapters/store.js
@@ -1,7 +1,7 @@
 import Error from '../adapters/error';
 import { version } from '../../config/constants.yml';
 import { findIndexIgnoreCase } from '../libs/string';
-import { getKeyWithoutVersion } from 'src/libs/pois';
+import { getKey } from 'src/libs/pois';
 import { fire } from 'src/libs/customEvents';
 import { isPoiCompliantKey } from 'src/libs/pois';
 
@@ -77,7 +77,7 @@ export async function getFavoritesMatching(term) {
 
 export async function isInFavorites(poi) {
   try {
-    return Boolean(get(getKeyWithoutVersion(poi)));
+    return Boolean(get(getKey(poi)));
   } catch (e) {
     Error.sendOnce('store', 'has', 'error checking existing key', e);
   }
@@ -85,7 +85,7 @@ export async function isInFavorites(poi) {
 
 export async function addToFavorites(poi) {
   try {
-    set(getKeyWithoutVersion(poi), poi);
+    set(getKey(poi), poi);
     fire('poi_added_to_favs', poi);
   } catch (e) {
     Error.sendOnce('store', 'add', 'error adding poi', e);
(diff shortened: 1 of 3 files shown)


src/adapters/store.js
```diff
@@ -5,35 +5,59 @@ import { getKey } from 'src/libs/pois';
...
+        : acc;
+    }, []);
+}
+
 export async function getAllFavorites() {
-  let localStorageKeys = [];
+  let keys = [];
   try {
-    localStorageKeys = Object.keys(localStorage);
+    keys = listKeys();
```
G-Ray, line 52: I think we can safely remove the try/catch block here. Wdyt ?

bbecquet: Has long as you try to access local storage, it's a good idea to use try/catch. Local storage has some weird error cases, for example in some browsers in certain security settings it can exist on `window` while still creating errors.
Some ref https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/ 
I'm not sure the rest of the code isn't completely safe against that, but this try/catch is useful eand doesn't cost us anythin.

src/adapters/store.js
```diff
@@ -5,35 +5,59 @@ import { getKey } from 'src/libs/pois';
...
+}
+
+/**
+ * List keys without prefix.
+ * In case some keys are not prefixed, we don't return them
+ */
+function listKeys() {
+  return Object
+    .keys(localStorage || {})
+    .reduce((acc, k) => {
```
bbecquet, line 41: `reduce` is not easy to read for this use case. Using `filter` and `map` would be easier to understand as it corresponds more closely to the actual operations you do : 1. taking only keys with the prefix, 2. removing the prefix part.

src/adapters/store.js
```diff
@@ -5,35 +5,59 @@ import { getKey } from 'src/libs/pois';
...
+    const prefixedKey = `${prefix}${k}`;
+    localStorage.setItem(prefixedKey, JSON.stringify(v));
   } catch (e) {
     Error.sendOnce('local_store', 'set', 'error setting item', e);
   }
 }
 
+function del(k) {
+  const prefixedKey = `${prefix}${k}`;
+  localStorage.removeItem(prefixedKey);
```
bbecquet, line 31: You need try/catch here as well (see my other comment).

src/adapters/store.js
G-Ray: actually, it was my first implementation :smile: 
I rewrite it more properly, it's easier to read, thanks

bbecquet (dismissed):

bbecquet (approved):

PR #914, opened 2020-11-12
Add a one-line address on POI items in result lists 

bbecquet:
## Description
Display the POI address as the last-line of POI items in result lists.
~~According to the design, we only display the street and street number, so I added a new option in `<Address>` to omit the city part.~~ Let's keep the city part for now.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran 2020-11-12 à 18 30 48|Capture d’écran 2020-11-12 à 18 31 08|

### DIFF
--- a/src/components/PoiItem.jsx
+++ b/src/components/PoiItem.jsx
@@ -4,6 +4,7 @@ import OpeningHour from 'src/components/OpeningHour';
 import OsmSchedule from 'src/adapters/osm_schedule';
 import ReviewScore from 'src/components/ReviewScore';
 import PoiTitleImage from 'src/panel/poi/PoiTitleImage';
+import Address from 'src/components/ui/Address';
 import classnames from 'classnames';
 import poiSubClass from 'src/mapbox/poi_subclass';
 import { capitalizeFirst } from 'src/libs/string';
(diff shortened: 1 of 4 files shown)


src/components/PoiItem.jsx
```diff
@@ -37,6 +38,9 @@ const PoiItem = React.memo(({ poi,
           <OpeningHour schedule={new OsmSchedule(poi.blocksByType.opening_hours)} />
         </div>}
       </div>
+      {inList && <div className="poiItem-address u-text--subtitle u-ellipsis">
+        <Address address={poi.address} inline omitCountry omitCity />
```
amatissart, line 42: A bit sad to use `omitCity` here :crying_cat_face:  
Is it due to the lack of space ?

bbecquet: This is what the designs show, but I agree with you, it would be better with it. I think it's a matter of space and to avoid visual clutter. But IMO we should display it. Maybe I can merge smth with it first, and discuss it afterwards?

bbecquet: I'll remove my last commit for now.

G-Ray (approved):

src/components/PoiItem.jsx
amatissart: Ok, thanks

amatissart (approved):

PR #915, opened 2020-11-13
Create standalone RoutesList component

G-Ray:
## Description
- Create standalone `RoutesList` component, with placeholder

## Why
Welcomed simplification to reorder selected routes on mobile

### DIFF
--- a/src/panel/direction/RoutesList.jsx
+++ b/src/panel/direction/RoutesList.jsx
@@ -1,0 +1,66 @@
+import React from 'react';
+
+import { Item, ItemList } from 'src/components/ui/ItemList';
+import PlaceholderText from 'src/components/ui/PlaceholderText';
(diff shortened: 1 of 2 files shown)


src/panel/direction/RoutesList.jsx
```diff
@@ -0,0 +1,66 @@
...
+        <Item
+          key={route.id}>
+          <Route
+            id={route.id}
+            route={route}
+            origin={origin}
+            destination={destination}
+            vehicle={vehicle}
+            isActive={route.id === activeRouteId}
+            showDetails={route.id === activeRouteId && activeDetails}
```
G-Ray, line 33: Please note we use route.id as identifier instead of index. This is more robust and required for reordering routes on mobile.

bbecquet (approved):

PR #916, opened 2020-11-13
RoutesList: On mobile, move active route to top

G-Ray:
## Description
On mobile, move active route to the top

Remaining task should be to resize the panel, using `fitContent`, in order to only show the selected route.

## Why
UX changes

## Screenshots
Peek 2020-11-13 17-28

On desktop, behavior has not changed.

### DIFF
--- a/src/panel/direction/RoutesList.jsx
+++ b/src/panel/direction/RoutesList.jsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { useContext } from 'react';
 
+import { DeviceContext } from 'src/libs/device';
 import { Item, ItemList } from 'src/components/ui/ItemList';
 import PlaceholderText from 'src/components/ui/PlaceholderText';
 import Route from './Route';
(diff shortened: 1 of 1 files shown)


src/panel/direction/RoutesList.jsx
```diff
@@ -63,4 +69,19 @@ const Placeholder = () => {
   );
 };
 
+const moveRouteToTop = (routes, id) => {
```
bbecquet, line 72: This seems really complicated for just that, and find/findIndex still don't exist on IE 11 :P
What about using sort?
```js
return routes
    .slice() // clone the array as sort operates on-place
    .sort((a, b) => a.id === id ? -1 : (b.id === id ? : 1 : 0));
```

src/panel/direction/RoutesList.jsx
```diff
@@ -63,4 +69,17 @@ const Placeholder = () => {
...
+  if (!id) {
```
bbecquet, line 73: I think it's useless with the current implementation, but it doesn't hurt.

bbecquet (approved):

PR #920, opened 2020-11-17
Manage filtered POI labels and icons on 2 separate MapBox-GL layers

bbecquet:
## Description
Split the drawing of the filtered POIs in two separate MapBox-GL layers, one for the pins, the other dedicated to the labels, so layout algorithms work better.
As both layers use the same GeoJSON data source and the dynamic states (active/selected) are managed on it, the duplication is minimal. But we still need to duplicate visibility toggling and event listener attachments.

## Why
The one-layer implementation had label layout issues, principaly many overlaps of labels on icons. It looked dirty and was bad for user readability.
It seems MapBox-GL overlap avoidance algorithms don't work well within the same layer, when we force it to display all icons. As the removal of labels on other layers below always worked correctly, we used the same principle for these labels.
Thanks @amatissart for the great and simple idea :)

## Screenshots
|Before|After|
|---|---|
|localhost_3000_places__type=restaurant bbox=0 1817103%2C49 6916535%2C0 2261389%2C49 7177631|localhost_3000_places__type=restaurant bbox=2 4152807%2C48 848743299999995%2C2 4828110999999997%2C48 878747399999995 (1)|
|IMG_0008|IMG_0007|
|localhost_3000_places__type=restaurant (5)|localhost_3000_places__type=restaurant (4)|

### DIFF
--- a/src/adapters/pois_styles.js
+++ b/src/adapters/pois_styles.js
@@ -1,5 +1,5 @@
 
-export const getFilteredPoisStyle = ({ withName = true } = {}) => ({
+export const getFilteredPoisPinStyle = () => ({
   type: 'symbol',
   layout: {
     'icon-image': ['concat', 'pin-', ['get', 'iconName']],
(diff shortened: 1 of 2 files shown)


amatissart (approved):
Nice!

G-Ray:
Great results ! The tiny duplication is really not an issue to me, and just have to iterate over `this.layers`.
:world_map: :label: :red_circle:

PR #922, opened 2020-11-17
fit map view onTransitionEnd

G-Ray:
## Description
- Move activeRouteId handling into DirectionPanel
- fit view with currently selected route once animation has finished, from maximized to default size.
- Accept onTransitionEnd props in Panel

## Why
Usability

## Screenshots
Peek 2020-11-20 11-20

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -115,7 +115,7 @@ export default class DirectionPanel extends React.Component {
       });
     }
 
-    if (this.props.selected !== prevProps.selected) {
+    if (this.props.selected !== prevProps.selected && this.state.routes.length > 0) {
       fire('set_main_route', { routeId: this.sanitizeSelected(), fitView: !isMobileDevice() });
     }
   }
@@ -200,6 +200,7 @@ export default class DirectionPanel extends React.Component {
         return;
       }
       if (directionResponse && directionResponse.error === 0) {
+
         // Valid, non-empty response
         const routes = directionResponse.data.routes
           .sort((routeA, routeB) => routeA.duration - routeB.duration)
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -63,7 +63,6 @@ export default class DirectionPanel extends React.Component {
       originInputText: '',
       destinationInputText: '',
       marginTop: 0,
-      activeRouteId: 0, //this.sanitizeSelected(),
     };
 
     this.restorePoints(props);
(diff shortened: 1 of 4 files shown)


bbecquet (commented):
It would be good to find a way to fit the view only when going from maximized to default.
When using the feature, the constant zoom out/in is distracting for the eye. There's actually no zoom out in the design video when maximizing.
By having it only called when going back to default, it would serve two purposes:
 - fit view if the panel for the newly selected alternative is different in height from the previous one (for example with a "via" string taking two lines)
 - re fit the view if the user moved the map before chosing a new route.

And in the more frequent case, with the coming behaviour (fitting all the alternatives), it would not be called at all.

G-Ray:
> It would be good to find a way to fit the view only when going from maximized to default.
> When using the feature, the constant zoom out/in is distracting for the eye. There's actually no zoom out in the design video when maximizing.
> By having it only called when going back to default, it would serve two purposes:
> 
> * fit view if the panel for the newly selected alternative is different in height from the previous one (for example with a "via" string taking two lines)
> * re fit the view if the user moved the map before chosing a new route.
> 
> And in the more frequent case, with the coming behaviour (fitting all the alternatives), it would not be called at all.

Thanks for your suggestion @bbecquet , from now, the "map refit" should be triggered only when transitioning from `maximized` to `default`. Let's see how it goes with your other change set :rocket:

bbecquet (commented):
This is an improvement but there is still a double fit, resulting in a useless map zoom out/in animation.
When you click on a route, the `fit_map` in `setMainRoute` of `scene_direction.js` is done while the panel is still maximized, triggering a zoom out… which is interrupted and turned into a zoom in on the selected alternative when the panel transition to default is complete.
IMO the first one shouldn't happen in mobile. In the most common case (the user didn't pan/zoom the map) we should see only a minimal change from the view fitting the previously selected route to the new one.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -410,8 +409,8 @@ export default class DirectionPanel extends React.Component {
                 </ShareMenu>,
               ]}
               onTransitionEnd={(prevSize, size) => {
-                if (prevSize === 'maximized' && size === 'default' && activeRouteId >= 0) {
-                  fire('set_main_route', { routeId: activeRouteId, fitView: true });
+                if (prevSize === 'maximized' && size === 'default' && this.sanitizeSelected() >= 0) {
```
bbecquet, line 412: To avoid calling `sanitizeSelected` so many times in the render (and simplify you diff), you could declare `activeRouteId` once at the top.

bbecquet (approved):

PR #927, opened 2020-11-18
Add eslint-plugin-react-hooks plugin

G-Ray:
## Description
eslint-plugin-react-hook would help us catch possible errors while implementing hooks.
It is an "official" plugin (https://reactjs.org/docs/hooks-rules.html)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,8 @@
     },
     "extends": [
       "eslint:recommended",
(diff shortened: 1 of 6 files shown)


bbecquet (commented):
Good idea! Any help on hooks with explanations is welcome :) 
But when run on the current master, there are already 4 warnings. We should fix them in another commit.

bbecquet (approved):

PR #933, opened 2020-11-23
[Duplicate state] Don't store 'isActive' status in route

bbecquet:
## Description
Little clean-up to avoid managing the "active" status (aka. "selected") of a route alternative in two places: as a `activeRouteId` prop passed across components, and in the route struct itself as a `isActive` bool.
Just keep the external prop.

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -199,11 +199,11 @@ export default class DirectionPanel extends React.Component {
         // Valid, non-empty response
         const routes = directionResponse.data.routes
           .sort((routeA, routeB) => routeA.duration - routeB.duration)
-          .map((route, i) => ({ ...route, isActive: i === 0, id: i }));
+          .map((route, i) => ({ ...route, id: i }));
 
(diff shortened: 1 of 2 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -199,11 +199,11 @@ export default class DirectionPanel extends React.Component {
         // Valid, non-empty response
         const routes = directionResponse.data.routes
           .sort((routeA, routeB) => routeA.duration - routeB.duration)
-          .map((route, i) => ({ ...route, isActive: i === 0, id: i }));
+          .map((route, i) => ({ ...route, id: i }));
 
         this.setState({ isLoading: false, error: 0, routes });
         window.execOnMapLoaded(() => {
-          fire('set_routes', { routes, vehicle });
+          fire('set_routes', { routes, vehicle, activeRouteId: 0 });
```
bbecquet, line 206: There is a known bug her, as `activeRouteId` shouldn't always be 0 but take the `selected` url param if it exists, but this will be solved elsewhere :)

G-Ray (approved):

PR #935, opened 2020-11-23
Direction:  use search param to details status

G-Ray:
## Description
- Use `details` search param in place of state to know if route details should be displayed.
- Render MobileRouteDetails in DirectionPanel. A React Portal is not required anymore.

## Why
Simpler state management, and ability to use browser's history to navigate between app states.

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -22,7 +22,8 @@ import { getInputValue } from 'src/libs/suggest';
 import { geolocationPermissions, getGeolocationPermission } from 'src/libs/geolocation';
 import { openPendingDirectionModal } from 'src/modals/GeolocationModal';
 import ShareMenu from 'src/components/ui/ShareMenu';
-import { parseQueryString, buildQueryString } from 'src/libs/url_utils';
+import { parseQueryString, buildQueryString, updateQueryString } from 'src/libs/url_utils';
+import MobileRouteDetails from './MobileRouteDetails';
 
 const MARGIN_TOP_OFFSET = 64; // reserve space to display map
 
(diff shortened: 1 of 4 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -322,6 +323,12 @@ export default class DirectionPanel extends React.Component {
     return handler(e);
   }
 
+  toggleDetails() {
+    const isDetailsInQuery = this.props.details === 'true';
+    const search = updateQueryString({ details: !isDetailsInQuery });
+    window.app.navigateTo('routes/' + search, {}, { replace: false });
```
G-Ray, line 329: Is it ok to set `replace: false` option each time we toggle details ?

bbecquet (commented):
Now that this details view has its own history state, closing it with the close button should do a history back action.
Otherwise, navigating backwards after closing the view with the close button will restore it, which is not logical. It should be restored with a forward instead.

src/panel/direction/DirectionPanel.jsx
```diff
@@ -117,6 +118,8 @@ export default class DirectionPanel extends React.Component {
 
     if (this.props.selected !== prevProps.selected && this.state.routes.length > 0) {
       fire('set_main_route', { routeId: this.sanitizeSelected(), fitView: !isMobileDevice() });
+      const search = updateQueryString({ details: false });
+      window.app.navigateTo('routes/' + search, {}, { replace: false });
```
bbecquet, line 122: It should also be restored to `false` when a new route is computed (for example by dragging the points) after the details of the current one have been displayed.

G-Ray: Ok we will resolve history behavior in a separate PR as discussed

bbecquet (approved):

PR #936, opened 2020-11-23
[Tests] Add integration tests for category/POI list

bbecquet:
## Description
Add a new integration test suite for the POI list (aka. "category panel") feature, which wasn't tested at all for now (!).
For now let's start small by testing some "easy" things:
 - how the category panel is opened or closed
 - how the main input content is changed when the panel is active

In the next step, I'll try to test parts of the data logic, but this will probably require complex mocking of API and map, so let's do it separately.

### DIFF
--- a/src/panel/category/CategoryPanel.jsx
+++ b/src/panel/category/CategoryPanel.jsx
@@ -73,7 +73,7 @@ export default class CategoryPanel extends React.Component {
       return;
     }
 
-    if (window.map.mb.isMoving()) {
+    if (window.map.mb.isMoving && window.map.mb.isMoving()) {
       /*
         Do not trigger API search and zoom change when the map
         is already moving, to avoid flickering.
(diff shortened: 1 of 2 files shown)


src/panel/category/CategoryPanel.jsx
```diff
@@ -73,7 +73,7 @@ export default class CategoryPanel extends React.Component {
       return;
     }
 
-    if (window.map.mb.isMoving()) {
+    if (window.map.mb.isMoving && window.map.mb.isMoving()) {
```
bbecquet, line 76: This test is required for now because `isMoving` isn't included in the Mapbox-GL-JS mock.

amatissart: Side note: this syntax could also be used, but I am not sure I like it:
```suggestion
    if (window.map.mb.isMoving?.()) {
```

bbecquet: Yeah, I hesitated, not sure I'm a fan either, but it's probably a question of getting used to it.
I think the best way is still to avoid cases were we need that :)

amatissart (approved):

PR #937, opened 2020-11-24
Minimize all panels on user interaction on the map

bbecquet:
## Description
Automatically minimize all the mobile panels when the user interacts with the map, like we already do with the POI panel.

## Why
Give the user the maximum screen space when they want to interact with the map.

### DIFF
--- a/src/panel/PanelManager.jsx
+++ b/src/panel/PanelManager.jsx
@@ -51,7 +51,7 @@ export default class PanelManager extends React.Component {
     window.times.appRendered = Date.now();
 
     listen('map_user_interaction', () => {
-      if (this.state.ActivePanel === PoiPanel && isMobileDevice()) {
+      if (isMobileDevice()) {
         this.setState({ panelSize: 'minimized' });
       }
     });


G-Ray (dismissed):
:+1: Easy to review and nice UX improvement

bbecquet:
I've amended the commit to remove the exception for the DirectionPanel, after the design confirmed it's ok to have the same behavior for the route result panel.

src/panel/PanelManager.jsx
```diff
@@ -51,7 +51,7 @@ export default class PanelManager extends React.Component {
     window.times.appRendered = Date.now();
 
     listen('map_user_interaction', () => {
-      if (this.state.ActivePanel === PoiPanel && isMobileDevice()) {
+      if (isMobileDevice()) {
```
G-Ray, line 54: I guess it's the wanted behavior even in direction panel ?

bbecquet: Yep, that's what I explained in the comment :)

G-Ray: ok, I missed it when re-reviewing

G-Ray (approved):

PR #940, opened 2020-11-24
[Fix] Properly return to the previous POI when closing the DirectionPanel with the button

bbecquet:
## Description
Make sure the current history state is re-used when we update the url to reflect changes in the DirectionPanel.

## Why
The behavior of the panel close button depends on the existence of a POI object passed to DirectionPanel through the history state and the router.
If we pass an empty state, we lose the `poi` prop the first time we update the url, so the panel believes it wasn't initialized with a POI and consider the close action to be a return to the Service panel.

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -245,7 +245,7 @@ export default class DirectionPanel extends React.Component {
     const search = buildQueryString(queryObject);
     const relativeUrl = 'routes/' + search;
 
-    window.app.navigateTo(relativeUrl, {}, {
+    window.app.navigateTo(relativeUrl, window.history.state, {
       replace: true,
     });
   }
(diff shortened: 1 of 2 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -245,7 +245,7 @@ export default class DirectionPanel extends React.Component {
     const search = buildQueryString(queryObject);
     const relativeUrl = 'routes/' + search;
 
-    window.app.navigateTo(relativeUrl, {}, {
+    window.app.navigateTo(relativeUrl, window.history.state, {
```
amatissart, line 248: A similar call exists in `RouteResult`, after an alternative route has been selected
https://github.com/QwantResearch/erdapfel/blob/master/src/panel/direction/RouteResult.jsx#L55

Don't you think it should it be updated too ?

bbecquet: You're right, thanks a lot.
I forgot about the other one… I hope we can remove it soon and make these imperative url changes go through the same method, but for now let's fix it also.

amatissart (approved):

G-Ray (approved):

PR #944, opened 2020-11-26
Add shortcuts to main categories on all types of POI panel 

bbecquet:
## Description
Instead of displaying the full list of categories at the bottom of the "address" POI panel, we now display the first four only but on all types of POI panels (address, POI from OSM or PJ, mobile or desktop).

Adding the categories themselves was the easy part thanks to the `CategoryList` component, but other things had to be adjusted to fit the design.
As part of these changes, the padding of the desktop POI panel was finally reduced to match the design. Until now it was the only one still having a large horizontal padding value. Now it should look more consistent with other parts of the application.

## Screenshots

|Before|After|
|---|---|
|Capture d’écran 2020-11-26 à 14 36 14|Capture d’écran 2020-11-26 à 14 35 49|
|Capture d’écran 2020-11-26 à 14 40 42|Capture d’écran 2020-11-26 à 14 40 49|
|Capture d’écran 2020-11-26 à 14 43 35|Capture d’écran 2020-11-26 à 14 43 25|

### DIFF
--- a/src/panel/poi/PoiPanel.jsx
+++ b/src/panel/poi/PoiPanel.jsx
@@ -17,7 +17,7 @@ import { addToFavorites, removeFromFavorites, isInFavorites } from 'src/adapters
 import PoiItem from 'src/components/PoiItem';
 import { isNullOrEmpty } from 'src/libs/object';
 import { DeviceContext } from 'src/libs/device';
-import { Flex, Panel, PanelNav, CloseButton } from 'src/components/ui';
+import { Flex, Panel, PanelNav, CloseButton, Divider } from 'src/components/ui';
 
 const covid19Enabled = (nconf.get().covid19 || {}).enabled;
 
@@ -265,7 +265,7 @@ export default class PoiPanel extends React.Component {
               />
               <CloseButton onClick={isMobile ? backAction : this.closeAction} position="topRight" />
             </Flex>
-            <div className="u-mb-xs">
+            <div className="u-mb-l">
               <ActionButtons
                 poi={poi}
                 isDirectionActive={this.isDirectionActive}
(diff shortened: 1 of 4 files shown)


src/scss/includes/panels/poi_panel.scss
```diff
@@ -13,8 +13,13 @@ $BLOCK_ICON_FONT_SIZE: 16px;
 
 .poi_panel__content {
   animation: appear 600ms forwards;
-  padding: 20px 30px 10px 30px;
+  padding: 20px 14px 0;
```
G-Ray, line 16: we can use $spacing-xxs here too :)

src/panel/poi/PoiPanel.jsx
```diff
@@ -278,18 +278,17 @@ export default class PoiPanel extends React.Component {
...
               </div>
               }
-              {poi.id.match(/latlon:/) && <div className="service_panel__categories--poi">
-                <h3 className="u-text--smallTitle">
-                  {_('Search around this place', 'poi')}
-                </h3>
-                <CategoryList />
-              </div>}
               {isFromOSM(poi) && <OsmContribution poi={poi} />}
+              <Divider paddingTop={0} className="poi_panel__fullWidthDivider" />
```
G-Ray, line 287: Would it be possible to create two blocks with class `poi_panel__fullContent`, and use a divider between them ? This way we would not need to set negative margin on divider.

src/scss/includes/panels/poi_panel.scss
bbecquet: I think you mean `$spacing-l` to replace the `20px`.
I'll do it, but `14px` will remain as it isn't a standard spacing constant.

src/panel/poi/PoiPanel.jsx
bbecquet: I didn't want to do that in this PR before we choose a generic box layout system for the rest of the app, so the negative margin approach seemed simple.
And I don't want to touch `.poi_panel__fullContent` or duplicate it, as it's used for the mobile Panel to hide content dynamically. What I would need here is divs with only horizontal paddings.
If you prefer this solution I can still do it for this panel only, even if it touches little things outside.

G-Ray: seems fair, we will do this later

G-Ray (approved):

PR #945, opened 2020-11-26
Qmaps 1763: Display missing address in roadmap

G-Ray:
## Description
- Fetch address if it's missing when setting a new origin/destination point. In some cases, like when clicking on the map to set a point, address is undefined (type `MapPoint`).
- Decide to display latlon poi, as they should have an address object too.

## Why
better UX as it can be considered as a bug.

## Screenshots

Setting a destination point from the map :
Peek 2020-11-27 10-33

From a POI panel :
Peek 2020-11-27 10-34

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -131,7 +131,8 @@ export default class DirectionPanel extends React.Component {
   }
 
   async setTextInput(which, poi) {
-    if (poi.type === 'latlon') {
+    if (!poi.address || Object.keys(poi.address).length === 0) {
+      // fetch missing address
       poi.address = await address.fetch(poi);
     }
     this.setState({ [which + 'InputText']: getInputValue(poi) });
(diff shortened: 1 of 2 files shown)


src/panel/direction/DirectionPanel.jsx
```diff
@@ -131,7 +131,8 @@ export default class DirectionPanel extends React.Component {
   }
 
   async setTextInput(which, poi) {
-    if (poi.type === 'latlon') {
+    if (!poi.address || Object.keys(poi.address).length === 0) {
```
bbecquet, line 134: You can use the `isNullOrEmpty` util from `src/libs/object.js` to do that.

G-Ray: great ! I was looking for an equivalent to `lodash.isEmpty()` :)

src/panel/direction/DirectionPanel.jsx
```diff
@@ -131,7 +132,8 @@ export default class DirectionPanel extends React.Component {
   }
 
   async setTextInput(which, poi) {
-    if (poi.type === 'latlon') {
+    if (!poi.address || isNullOrEmpty(poi.address)) {
```
bbecquet, line 135: You don't need the `||`, it's managed by `isNullOrEmpty`.

bbecquet (commented):
To avoid useless repetition, we should omit the street part when dealing with an address/position. It would be consistent with what we do on the second line in the suggest when the first line is already the street.
Capture d’écran de 2020-11-27 11-34-56

bbecquet (approved):

PR #947, opened 2020-11-30
Simplify URL params management in DirectionPanel

bbecquet:
## Description
Following recent additions to the URL scheme controlling the behavior of the DirectionPanel, this PR proposes some simplifications in the manipulation of query params and routes by this component.

1. The immediate parsing of `details` and `selected` is done outside of the `DirectionPanel`. It receives these as props with the correct type and can only focus on the consistency checks
2. The number of places where we directly change URL is reduces, and it's done only in the top component. That way it's easier to track and keep consistent with the internal state of the panel.

What do you think @G-Ray?

### DIFF
--- a/src/libs/url_utils.js
+++ b/src/libs/url_utils.js
@@ -65,10 +65,3 @@ export function buildQueryString(queriesObject) {
   const params = new URLSearchParams(removeNullEntries(queriesObject)).toString();
   return params ? `?${params}` : '';
 }
-
-export function updateQueryString(queriesObject) {
-  return buildQueryString({
-    ...parseQueryString(window.location.search),
-    ...queriesObject,
-  });
-}
--- a/src/libs/url_utils.js
+++ b/src/libs/url_utils.js
@@ -65,3 +65,10 @@ export function buildQueryString(queriesObject) {
   const params = new URLSearchParams(removeNullEntries(queriesObject)).toString();
   return params ? `?${params}` : '';
 }
+
+export function updateQueryString(queriesObject) {
+  return buildQueryString({
+    ...parseQueryString(window.location.search),
+    ...queriesObject,
+  });
+}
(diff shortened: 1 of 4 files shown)


src/libs/url_utils.js
```diff
@@ -65,10 +65,3 @@ export function buildQueryString(queriesObject) {
   const params = new URLSearchParams(removeNullEntries(queriesObject)).toString();
   return params ? `?${params}` : '';
 }
-
-export function updateQueryString(queriesObject) {
```
G-Ray, line 69: Actually i liked this function :smile: 
Maybe we can use it in some places, like here https://github.com/QwantResearch/erdapfel/pull/947/files#diff-25b5c163df2044a30533a391fcc5c83ca9bdb174a2e1efad2302a2281e937d08R244 ?

bbecquet: I have nothing against this function but as it was used at the end in a single place and the code is straightforward with object destructuring, I didn't think it was that useful.
But yeah, I agree it could come handy again in the future, so many we can keep it.

src/panel/direction/RouteResult.jsx
```diff
@@ -30,25 +31,13 @@ export default class RouteResult extends React.Component {
...
     if (routeId === this.props.activeRouteId) {
       return;
     }
 
     Telemetry.add(Telemetry.ITINERARY_ROUTE_SELECT);
-
-    this.updateUrl({ selected: routeId });
-  }
-
-  updateUrl = queryObject => {
```
G-Ray, line 49: awesome !

src/libs/url_utils.js
bbecquet: I restored it :)

src/panel/PanelManager.jsx
```diff
@@ -149,9 +149,12 @@ export default class PanelManager extends React.Component {
         || parseQueryString(document.location.search)['pt'] === 'true';
 
       router.addRoute('Routes', '/routes(?:/?)(.*)', (routeParams, options) => {
+        const params = parseQueryString(routeParams);
+        params.details = params.details === 'true';
+        params.activeRouteId = Number(params.selected);
```
G-Ray, line 154: I think we should handle the case where this returns a `NaN`. Can we default to 0 ?

bbecquet: Done!

G-Ray (approved):

PR #948, opened 2020-12-01
Direction details: replace url conditionnally

G-Ray:
## Description
- replace url only when details view is closing. This way, the "back" button on mobile can be used to go back to the previous state, where details view was previously closed, meanwhile we avoid "pushing" indefinetely into the history.

However, the back button will have no effect if the user opens a url already containing `details=true` search param. Do you have any idea to improve this behavior ?

## Why
Improve UX

### DIFF
--- a/src/panel/direction/DirectionPanel.jsx
+++ b/src/panel/direction/DirectionPanel.jsx
@@ -327,7 +327,8 @@ export default class DirectionPanel extends React.Component {
   }
 
   toggleDetails() {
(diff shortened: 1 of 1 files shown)


bbecquet (commented):
Sadly there is an undesirable effect to this: everytime you open/close the details, the same url is pushed as a new history state on the stack. So you need as many "back" actions to exit the Direction panel, but each time there is no visual changes to it, so it looks like something is broken.

G-Ray:
Closing this because the actual behavior is the best one for now

PR #949, opened 2020-12-01
Refactor Suggest: callback when suggestions open/close

bbecquet:
## Description
A small refacto on the `Suggest` component to make it more generic.
Instead of taking a prop to hide the main UI panel when suggestions are open and call this very specific action directly in `Suggest`, take a callback which sends the suggestions visibility state, so any action can be implemented by the parent.

## Why
I will need to listen to this on the `Suggest` instances used by the `DirectionPanel`, so I need something more generic than `hidePanelOnOpen`.
This also reduces coupling across hi-level and lo-level components.


G-Ray (approved):
:+1:

PR #952, opened 2020-12-01
Fit_map on origin and destination click

G-Ray:
## Description
Trigger a `fit_map` event with origin or destination point.
Please note the zoom level is not the same as when we trigger a `zoom_step`. What do you think ?

## Why
Improve UX

## Screenshots
Peek 2020-12-01 18-08


bbecquet (approved):

bbecquet:
>Please note the zoom level is not the same as when we trigger a zoom_step. What do you think ?

Looks fine to me like this.
For now the zoom on a step is done on a line segment, so it will be necessarily different. We previously discussed how we should zoom on the maneuver point for steps, instead of the full route segment, and I hope we'll soon do it. We'll decide at this point if we should use the same zoom value.

PR #953, opened 2020-12-02
Qmaps 1772 step by step

xem:
## Description
Redesign mobile step-by-step mode for itineraries
- return button on top left
- scrollable list of cards at the bottom
- scroll-snap + custom style for current step (highlighted) and past steps (grey)
- no geolocation button (for now)
- cards have a fixed height by default (2 lines of text)
- if a third line of text is present; it fades to white, calling the user to click on the card to see more
- when clicking a card, all the cards expand to show all the text they contain and the 
 map's scale moves up
- when clicking again, they all get back to the original size, and the map's scale moves down
- if a card is extremely high (> 30vh), its content becomes scrollable
- further redesign will happen later!

## Screenshots
sbs2
sbs3

### DIFF
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -2,66 +2,71 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import RoadMapStep from './RoadMapStep';
 import { fire } from 'src/libs/customEvents';
-import { CloseButton } from 'src/components/ui';
 
 export default class MobileRoadMapPreview extends React.Component {
   static propTypes = {
     steps: PropTypes.array.isRequired,
     onClose: PropTypes.func.isRequired,
-  }
+  };
 
-  state = {
-    currentStep: 0,
-  }
+  state = { };
 
   componentDidMount() {
-    this.highlightCurrentStep();
-    fire('move_mobile_geolocation_button', 65);
+    this.steps = document.querySelector('.mobile-roadmap-preview-steps');
+    this.stepList = document.querySelectorAll('.itinerary_mobile_step');
+    this.currentStep = 0;
     fire('move_mobile_bottom_ui', 105);
+    this.highlightCurrentStep();
   }
 
   highlightCurrentStep = () => {
-    fire('zoom_step', this.props.steps[this.state.currentStep]);
-    const stepList = document.querySelectorAll('.itinerary_mobile_step');
-    stepList.forEach((item, index) => {
+    fire('zoom_step', this.props.steps[this.currentStep]);
+    this.stepList.forEach((item, index) => {
 
       // Reset the style of all the steps
       item.classList.remove('active');
       item.classList.remove('past');
 
       // Show previous steps as "past"
-      if (index < this.state.currentStep) {
+      if (index < this.currentStep) {
         item.classList.add('past');
       }
     });
 
     // Show current step as "active"
-    stepList[this.state.currentStep].classList.add('active');
-  }
+    this.stepList[this.currentStep].classList.add('active');
+  };
 
   scroll = () => {
-    const steps = document.querySelector('.mobile-roadmap-preview-steps');
+
     const currentStep = Math.floor(
       // Divide the step container's scrollLeft up to the middle of the screen with the size of a step
       // to determine which step is present at the middle of the screen
-      (steps.scrollLeft + window.innerWidth / 2) / (window.innerWidth - 70 + 12)
+      (this.steps.scrollLeft + window.innerWidth / 2) / (window.innerWidth - 70 + 12)
     );
 
     // If it has changed, save it and highlight it as the current step
-    if (this.state.currentStep !== currentStep) {
-      this.setState(
-        () => ({ currentStep }),
-        this.highlightCurrentStep
-      );
+    if (this.currentStep !== currentStep) {
+      this.currentStep = currentStep;
+      this.highlightCurrentStep();
     }
-  }
+  };
+
+  toggleSize = () => {
+    this.steps.classList.toggle('expanded');
+    fire('move_mobile_bottom_ui', this.steps.offsetHeight);
+  };
 
   render() {
     return <div className="itinerary_mobile_step_by_step">
       <div className="mobile-roadmap-preview-close" onClick={this.props.onClose}>
         <i className="icon-arrow-left"/>
       </div>
-      <div className="mobile-roadmap-preview-steps" onScroll={this.scroll}>
+      <div
+        className="mobile-roadmap-preview-steps"
+        onScroll={this.scroll}
+        onClick={this.toggleSize}
+      >
         {
           this.props.steps.map((step, index) =>
             <div key={index} className="itinerary_mobile_step">
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -24,30 +24,19 @@ const MobileRoadMapPreview = ({
     // If it has changed, save it and highlight it as the current step
     if (currentStep !== newStep) {
       setCurrentStep(newStep);
-      fire('zoom_step', steps[newStep]);
     }
   };
 
   const toggleSize = () => {
     setExpanded(!expanded);
-    fire('move_mobile_bottom_ui', stepsRef.current.offsetHeight);
   };
-
-  // On load, zoom on step 0, unexpanded
-  useEffect(
-    () => {
-      fire('move_mobile_bottom_ui', 105);
-      fire('zoom_step', steps[0]);
-    },
-    []
-  );
 
   // When expanded is modified, move the map's bottom UI
   useEffect(
     () => {
       fire('move_mobile_bottom_ui', stepsRef.current.offsetHeight);
     },
-    [expanded]
+    [expanded, stepsRef]
   );
 
   // When currentStep is modified, zoom on it
@@ -55,7 +44,7 @@ const MobileRoadMapPreview = ({
     () => {
       fire('zoom_step', steps[currentStep]);
     },
-    [currentStep]
+    [currentStep, steps]
   );
 
   return <div className="itinerary_mobile_step_by_step">
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -15,54 +15,61 @@ export default class MobileRoadMapPreview extends React.Component {
   }
 
   componentDidMount() {
-    this.zoomToCurrentStep();
+    this.highlightCurrentStep();
     fire('move_mobile_geolocation_button', 65);
-    document.body.classList.add('directions-stepByStep');
+    fire('move_mobile_bottom_ui', 105);
   }
 
-  componentWillUnmount() {
-    document.body.classList.remove('directions-stepByStep');
-  }
-
-  zoomToCurrentStep = () => {
+  highlightCurrentStep = () => {
     fire('zoom_step', this.props.steps[this.state.currentStep]);
+    const stepList = document.querySelectorAll('.itinerary_mobile_step');
+    stepList.forEach((item, index) => {
+
+      // Reset the style of all the steps
+      item.classList.remove('active');
+      item.classList.remove('past');
+
+      // Show previous steps as "past"
+      if (index < this.state.currentStep) {
+        item.classList.add('past');
+      }
+    });
+
+    // Show current step as "active"
+    stepList[this.state.currentStep].classList.add('active');
   }
 
-  previous = () => {
-    this.setState(
-      prevState => ({ currentStep: Math.max(prevState.currentStep - 1, 0) }),
-      this.zoomToCurrentStep
+  scroll = () => {
+    const steps = document.querySelector('.mobile-roadmap-preview-steps');
+    const currentStep = Math.floor(
+      // Divide the step container's scrollLeft up to the middle of the screen with the size of a step
+      // to determine which step is present at the middle of the screen
+      (steps.scrollLeft + window.innerWidth / 2) / (window.innerWidth - 70 + 12)
     );
-  }
 
-  next = () => {
-    this.setState(
-      prevState => ({ currentStep: Math.min(prevState.currentStep + 1, this.props.steps.length) }),
-      this.zoomToCurrentStep
-    );
+    // If it has changed, save it and highlight it as the current step
+    if (this.state.currentStep !== currentStep) {
+      this.setState(
+        () => ({ currentStep }),
+        this.highlightCurrentStep
+      );
+    }
   }
 
   render() {
-    const currentStep = this.props.steps[this.state.currentStep];
-
-    return <React.Fragment>
-      <div className="itinerary_mobile_step">
-        <CloseButton onClick={this.props.onClose} />
-        <RoadMapStep step={currentStep} />
+    return <div className="itinerary_mobile_step_by_step">
+      <div className="mobile-roadmap-preview-close" onClick={this.props.onClose}>
+        <i className="icon-arrow-left"/>
       </div>
-
-      <div className="itinerary_mobile_step_buttons">
-        <button
-          className="itinerary_mobile_step_button icon-chevron-left"
-          disabled={this.state.currentStep === 0}
-          onClick={this.previous}
-        />
-        <button
-          className="itinerary_mobile_step_button icon-chevron-right"
-          disabled={this.state.currentStep === this.props.steps.length - 1}
-          onClick={this.next}
-        />
+      <div className="mobile-roadmap-preview-steps" onScroll={this.scroll}>
+        {
+          this.props.steps.map((step, index) =>
+            <div key={index} className="itinerary_mobile_step">
+              <RoadMapStep step={step}/>
+            </div>)
+        }
+        <div className="itinerary_mobile_step spacer"></div>
       </div>
-    </React.Fragment>;
+    </div>;
   }
 }
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -1,81 +1,98 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import RoadMapStep from './RoadMapStep';
 import { fire } from 'src/libs/customEvents';
 import classnames from 'classnames';
 import { Button } from '../../components/ui';
 
-export default class MobileRoadMapPreview extends React.Component {
-  constructor(props) {
-    super(props);
-    this.stepsRef = React.createRef();
-  }
+const MobileRoadMapPreview = ({
+  steps,
+  onClose,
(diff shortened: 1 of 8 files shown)


xem:
à discuter: il faudra peut-être retoucher les paddings du flyTo quand la map se déplace de step en step, j'ai l'impression que parfois la card cache une partie du segment de route en cours (assez rare)

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
   static propTypes = {
     steps: PropTypes.array.isRequired,
     onClose: PropTypes.func.isRequired,
-  }
+  };
 
-  state = {
-    currentStep: 0,
-  }
+  state = { };
```
bbecquet, line 12: You can remove that if you don't need a state…

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
 
   componentDidMount() {
-    this.zoomToCurrentStep();
-    fire('move_mobile_geolocation_button', 65);
+    this.steps = document.querySelector('.mobile-roadmap-preview-steps');
+    this.stepList = document.querySelectorAll('.itinerary_mobile_step');
+    this.currentStep = 0;
```
bbecquet, line 17: … but actually, this is typically something that should be stored in the state :)
This perfectly matches the definition of it: an internal value storing a part of the current component state which, when it changes, should result in a visual change.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
+    fire('move_mobile_bottom_ui', 105);
+    this.highlightCurrentStep();
   }
 
-  zoomToCurrentStep = () => {
-    fire('zoom_step', this.props.steps[this.state.currentStep]);
-  }
+  highlightCurrentStep = () => {
```
bbecquet, line 22: This follows my previous comment: most of what you do imperatively in this method is the job of React.
By managing `currentStep` in the state, each change to it will trigger a re-render. In the `render` method, you'll have all the necessary pieces to assign the right `className` to each step DOM element while you render them, without the need for an external loop like this.

The main thing to keep in mind: in React, we don't render once then manipulate the DOM. Instead, we manipulate data state and the DOM always follows by reflecting changes.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
+    }
+  };
 
-    return <React.Fragment>
-      <div className="itinerary_mobile_step">
-        <CloseButton onClick={this.props.onClose} />
-        <RoadMapStep step={currentStep} />
-      </div>
+  toggleSize = () => {
+    this.steps.classList.toggle('expanded');
```
bbecquet, line 56: Same for this `expanded` status, this looks like a state value.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
-          onClick={this.previous}
-        />
-        <button
-          className="itinerary_mobile_step_button icon-chevron-right"
-          disabled={this.state.currentStep === this.props.steps.length - 1}
-          onClick={this.next}
-        />
+  render() {
+    return <div className="itinerary_mobile_step_by_step">
+      <div className="mobile-roadmap-preview-close" onClick={this.props.onClose}>
```
bbecquet, line 62: This could probably use the `<Button>` component, which creates a real `button` element.

bbecquet: I know it's consistent with the design, but this back button looks awfully small. Was it discussed?
Anyway, if not it's ok to move on with this size and come back on it later.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
+        onScroll={this.scroll}
+        onClick={this.toggleSize}
+      >
+        {
+          this.props.steps.map((step, index) =>
+            <div key={index} className="itinerary_mobile_step">
+              <RoadMapStep step={step}/>
+            </div>)
+        }
+        <div className="itinerary_mobile_step spacer"></div>
```
bbecquet, line 76: Isn't it something that could be added as an `:after` pseudo element?

src/panel/direction/RoadMapItem.jsx
```diff
@@ -7,7 +7,7 @@ const RoadMapItem = ({ children, icon, distance, className, line, ...rest }) =>
   <>
     <div className={classnames('itinerary_roadmap_item', className)} {...rest}>
       {line}
-      <Flex alignItems="center">
+      <Flex alignItems="flex-start">
```
bbecquet, line 10: This component is also used in the roadmaps displayed as list, so it changes the alignement of the icon in it too.

src/scss/includes/direction.scss
```diff
@@ -287,12 +287,11 @@ button.direction_shortcut {
 
   .itinerary_mobile_step {
```
bbecquet, line 288: This is the same class you overwrite below in a more specific selector. A lot of the values declared here are overwritten (ex: `position: fixed`) or become obsolete (ex : bottom/left/right).
I suggest you merge both declaration and maybe simplifying the nesting of selectors.

src/scss/includes/direction.scss
```diff
@@ -395,8 +368,115 @@ button.direction_shortcut {
...
+          background-color: $grey-lighter;
+
+          .itinerary_roadmap_instruction:after {
+            background: linear-gradient(#e8e8ed77, $grey-lighter);
+          }
+        }
+
+        &.active {
+          border: 1px solid $action-blue-dark;
+          box-shadow: 0 0 0 4px rgba(128, 172, 255, 0.2);
```
bbecquet, line 441: The shadow effect gets cut at the top. Probably a problem with the height of the list itself.
Capture d’écran de 2020-12-03 10-37-01

bbecquet (commented):
This works great :+1:
The code should be refactored to be more React though… and it will actually make things simpler :)

src/panel/direction/MobileRoadMapPreview.jsx
xem: Hi, thanks for the review, I actually started by doing this stuff in the proper "react" way, but when I did that, I had the impression that the snap-scroll seemed more janky, or buggy when the setState happened in the middle of a scroll. Maybe it was just a problem with my PC or my Chrome's mobile view, we could put the setStates back and verify it on real mobile devices if you want ...?

bbecquet: Hi!
That would be great to restore, yes :)

About the perf, there are common problems, not related to React in itself, when you plug DOM modifications directly to scroll events, as these are generated very quickly and can cause the DOM to "choke" with too many changes.

But I think if you keep you test in the scroll method to avoid doing a `this.setState({ currentStep })` when it's not needed, it should be fine. And if not we can still using throttling strategies but probably it won't be needed here.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,79 @@ import React from 'react';
...
 
-  state = {
-    currentStep: 0,
-  }
+  state = { };
 
   componentDidMount() {
-    this.zoomToCurrentStep();
-    fire('move_mobile_geolocation_button', 65);
+    this.steps = document.querySelector('.mobile-roadmap-preview-steps');
```
G-Ray, line 15: I think we could use ref instead

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -2,62 +2,80 @@ import React from 'react';
...
+  };
 
-    return <React.Fragment>
-      <div className="itinerary_mobile_step">
-        <CloseButton onClick={this.props.onClose} />
-        <RoadMapStep step={currentStep} />
-      </div>
+  toggleSize = () => {
+    this.setState( { expanded: !this.state.expanded } );
+    fire('move_mobile_bottom_ui', this.stepsRef.current.offsetHeight);
```
bbecquet, line 46: `setState` and the triggered `render` can be async, so here you're not guaranteed that the `offsetHeight` will be the new one.
This is what you see in action here, where things are desynchronized.
Peek 08-12-2020 12-07

To solve this, we must be sure before firing the event that the component has been updated after the `setState`. This is what the `componentDidUpdate` is for, so you can use it to check what state value has changed and trigger the right events while being sure the render pass has been made.

This is one solution, but this is the "old way". I think this is the good occasion to convert this component to a functional one, when such actions (actually side-effects) can be written much more naturally:
```js
const [expanded, setExpanded] = useState(false); // functional way of declaring state
// ...
useEffect(
    () => { /* ... */  }, // do what you want…
    [expanded]  // … everytime "expanded" changes
);  
```

This would require other changes to the component, but I think it's worth it.
Would you want to try?

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -1,63 +1,98 @@
...
+    // If it has changed, save it and highlight it as the current step
+    if (currentStep !== newStep) {
+      setCurrentStep(newStep);
+      fire('zoom_step', steps[newStep]);
+    }
+  };
+
+  const toggleSize = () => {
+    setExpanded(!expanded);
+    fire('move_mobile_bottom_ui', stepsRef.current.offsetHeight);
```
bbecquet, line 33: Same here, just focus on modifiying the state and let side effects do the work for you with a state that is sure up-to-date.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -1,63 +1,98 @@
...
+  };
+
+  // On load, zoom on step 0, unexpanded
+  useEffect(
```
bbecquet, line 37: You don't need three separate `useEffect`. Two should be enough, because you have two side effects:
 - an event firing when `expanded` value change (including on first render)
 - an event firing when `currentStep` value change (including on first render)
  
```js
useEffect(
    () => { fire('move_mobile_bottom_ui', stepsRef.current.offsetHeight); },  // with probably a check for stepsRef because it will be null on first render
    [expanded /*, ... */]
);

useEffect(
    () => { fire('zoom_step', steps[currentStep]); },
    [currentStep /*, ... */]
);
```

In functional components you don't think in terms of mount=>updated=>unmount, but in terms of which component "state" (actually a mix of its React state, its props, its refs) value triggers which side effect. 
And don't worry, it's perfectly normal to go through these steps after time spent on class components :smiley: 
It took me sooo much time to understand most of the time it's simpler like this, and I'm still not fluent.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -1,63 +1,98 @@
...
+    },
+    []
+  );
+
+  // When expanded is modified, move the map's bottom UI
+  useEffect(
+    () => {
+      fire('move_mobile_bottom_ui', stepsRef.current.offsetHeight);
+    },
+    [expanded]
```
bbecquet, line 50: You should see linter warnings about missing stuff here. Because anything used by the side effect has to be considered as possibly changing, and then declared in the array.
Here, you'll have to add `stepsRef`, as there is no guarantee another part of the component won't trigger a change of this ref, so you want to be sure this effect will be run again in this case.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -1,63 +1,98 @@
...
+  );
+
+  // When currentStep is modified, zoom on it
+  useEffect(
+    () => {
+      fire('zoom_step', steps[currentStep]);
+    },
+    [currentStep]
```
bbecquet, line 58: Same for `steps` here. And this is normal, if for any reason the `steps` prop changes, you'll want to trigger the effect.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -1,63 +1,98 @@
...
-        />
-      </div>
-    </React.Fragment>;
-  }
-}
+
+    // If it has changed, save it and highlight it as the current step
+    if (currentStep !== newStep) {
+      setCurrentStep(newStep);
+      fire('zoom_step', steps[newStep]);
```
bbecquet, line 27: This will be managed by the corresponding useEffect so it can be removed from here.

G-Ray:
It works really well !
However I have some hiccups when scrolling until the very last item. Am I the only one who experience this behavior ?
Peek 2020-12-09 17-47

src/scss/includes/direction.scss
```diff
@@ -395,8 +326,138 @@ button.direction_shortcut {
...
+
+    .mobile-roadmap-preview-close {
+      border: none;
+      padding: 0 0 0 8px;
+      margin: 0 0 0 11px;
+      background: $background;
+      width: 32px;
+      height: 32px;
+      border-radius: 50%;
+      box-shadow: 0 1px 4px 0 rgba(12, 12, 14, 0.2), 0 0 2px 0 rgba(12, 12, 14, 0.12);
```
G-Ray, line 344: you ca use $shadow which is defined in `colors.scss`

src/scss/includes/direction.scss
```diff
@@ -395,8 +326,138 @@ button.direction_shortcut {
...
   .autocomplete_suggestions {
     height: 100vh;
   }
+
+  // Step by step mobile view
+  .itinerary_mobile_step_by_step {
+
+    .mobile-roadmap-preview-close {
+      border: none;
+      padding: 0 0 0 8px;
```
G-Ray, line 338: I would suggest to use our new spacing values :)

src/scss/includes/direction.scss
```diff
@@ -395,8 +326,138 @@ button.direction_shortcut {
...
+        user-select: none;
+        flex: 0 0 calc(100vw - 70px);
+        margin: 0 6px;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        scroll-snap-align: start;
+        border: none;
+        border-radius: 12px;
+        box-shadow: 0 1px 4px 0 rgba(12, 12, 14, 0.2), 0 0 2px 0 rgba(12, 12, 14, 0.12);
```
G-Ray, line 401: $shadow

src/scss/includes/direction.scss
```diff
@@ -395,8 +326,138 @@ button.direction_shortcut {
...
+          background-color: $grey-lighter;
+
+          .itinerary_roadmap_instruction:after {
+            background: linear-gradient(#e8e8ed77, $grey-lighter);
+          }
+        }
+
+        &.active {
+          border: 1px solid $action-blue-dark;
+          box-shadow: 0 0 0 4px rgba(128, 172, 255, 0.2);
```
G-Ray, line 413: this value is defined as`$shadow-focus` in `colors.scss`

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -1,68 +1,87 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
G-Ray, line 1: :+1:

src/panel/direction/RoadMapItem.jsx
```diff
@@ -3,11 +3,11 @@ import classnames from 'classnames';
 
 import { Divider, Flex } from 'src/components/ui';
 
-const RoadMapItem = ({ children, icon, distance, className, line, ...rest }) =>
+const RoadMapItem = ({ children, icon, distance, className, line, flex, ...rest }) =>
```
bbecquet, line 6: `flex` isn't a good param name, we don't understand what it does. `alignItems` would be better.
And you could give it a default value equals to `center`, that way you don't need to modify other files not concerned by this PR and manage the special step-by-step case as a kind of exception (which it is).

xem:
> 
> 
> It works really well !
> However I have some hiccups when scrolling until the very last item. Am I the only one who experience this behavior ?
> Peek 2020-12-09 17-47

I also have random hiccups on Chrome PC + mobile simulator, but Benjamin said it was fine on real devices (to be confirmed maybe?)

bbecquet:
> > It works really well !
> > However I have some hiccups when scrolling until the very last item. Am I the only one who experience this behavior ?
> 
> I also have random hiccups on Chrome PC + mobile simulator, but Benjamin said it was fine on real devices (to be confirmed maybe?)

@xem @G-Ray 
Well, I see the same thing of the responsive view of Chrome dekstop, but on my real Android Chrome, I don't know, the whole app has always been laggy anyways :man_shrugging:
On iOS hovewer, it behaves really well.

bbecquet (approved):

amatissart:
I don't reproduce the hiccups you describe.  
But I noticed it's no longer possible to move or rotate the map when the step-by-step mode is on. The map position will stick to the position fitting the current step. Is it an expected behavior ?

xem:
@amatissart it is not, I'm looking into it (I suspect a kind of hidden infinite loop that provokes the scroll hiccups/lags and this... I just need to find where it is :) )

bbecquet:
> @amatissart it is not, I'm looking into it (I suspect a kind of hidden infinite loop that provokes the scroll hiccups/lags and this... I just need to find where it is :) )

@xem apparently the `zoom_step` event is fired when we touch the map, causing the view to always recenter on the current step. Let's move up the chain to understand why:
 1. according to the `useEffect` declaration, this can happen only when the `currentStep` state value or the `steps` prop changes
 2. as you took care to change `currentStep` only when needed, it's probably because the `steps` array changes when we click the map
 3. this can be confirmed by removing `steps` from the array of variables tracked for changes across render: if you remove it, it works
 4. in `DirectionPanel`, `steps` is passed as a dynamically-computed array each time we render `MobileRoadMapPreview`. It's logical that the array is considered different each time, because React will compare by ref, not content.
```jsx
<MobileRoadMapPreview
    steps={getAllSteps(activePreviewRoute)}
    onClose={this.onClose}
/>
```

So, to sum things up, we have to understand why moving the map causes a re-render of the `DirectionPanel`.
Looking quickly at it, it seems unnecessary and costly, so it may be a good target to bring optimization by removing it.
And if it proves useful at the end, we'll have to find a way to avoid recomputing the array each time.

bbecquet:
@xem @amatissart
I put a break point in the `DirectionPanel` `render` method and found the guilty state change in the callstack:

Capture d’écran de 2020-12-15 10-20-29

Conclusion: it has nothing to do with the direction feature itself, it's related to the recent change I made to minimize the mobile Panel on any map interaction. This triggers a `setState` which will re-render any component attached to the device context (in practice, all panels I think), even if the value itself hasn't changed (that's the confusing way `setState` works).

This can be fixed by avoiding the `setState` when `panelSize` already is `minimized`, or implementing a `shouldComponentUdpate`, or using a functional component for `PanelManager` so that state updates are atomic and don't result in a re-render if no change is made…

As it's not related to the step-by-step but would be an optim for the whole app, I propose we merge this PR and I do a separate  fix.

PR #955, opened 2020-12-02
Add button link on qwant logo

G-Ray:
## Description
Transform logo into a button which navigate to the homepage.

Please note in the animation below, there is no extra padding around the logo. Is it ok ?

## Why
UX.

## Screenshots
Peek 2020-12-02 18-05

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -43,6 +43,7 @@
           type="button"
           onclick="window.app.navigateTo('/');"
           class="search_form__logo__button"
+          title="<%= _('Qwant Maps Home', 'search bar') %>"
           <% if(config.app.versionFlag) { %>
             data-flag-text="<%= config.app.versionFlag %>"
           <% } %>
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -38,11 +38,15 @@
     <div id="react_root"></div>
     <div class="react_modal__container"></div>
     <div class="top_bar">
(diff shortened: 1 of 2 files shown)


views/index.ejs
```diff
@@ -38,11 +38,15 @@
     <div id="react_root"></div>
     <div class="react_modal__container"></div>
     <div class="top_bar">
-      <form onsubmit="return false" novalidate class="search_form"
-        <% if(config.app.versionFlag) { %>
-          data-flag-text="<%= config.app.versionFlag %>"
-        <% } %> >
-        <div class="search_form__logo"></div>
+      <form onsubmit="return false" novalidate class="search_form">
+        <button
```
bbecquet, line 42: Let's add a title to it like we did on other buttons. Especially as there is no inner content.
What about `title="<%= _('Qwant Maps home') %>"`? Or `'Back to home'` ?

G-Ray: wdyt about just "Qwant Maps" ?

bbecquet: The goal is to describe the action. "Qwant Maps" seems quite ambiguous. But we can ask the design team.

G-Ray: I asked the design team. For now, we decided to go with your proposition `Qwant Maps Home`

bbecquet (approved):

PR #956, opened 2020-12-03
direction: set font-size in placeholder to 14px

G-Ray:
## Description
Need to confirm if this is the expected value on desktop too


bbecquet (approved):
I'm approving in case there is no required change.

G-Ray:
It seems it's the expected value on desktop according the the mockups. I'm going to merge this one.

PR #959, opened 2020-12-04
Tweak search bar to show back button on focus only

G-Ray:
## Description
On mobile, display back button only when the search input is focus.

## Why
UX improvements

## Screenshots
Peek 2020-12-04 11-06

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -45,7 +45,7 @@
         <div class="search_form__logo"></div>
         <div id="react_menu__container"></div>
         <div class="search_form__wrapper empty">
-          <div class="search_form__return icon-arrow-left" onmousedown="clearSearch(event, true)"></div>
+          <div class="search_form__return icon-arrow-left"></div>
           <input id="search" type="search" class="search_form__input" spellcheck="false" required placeholder="<%= _('Search on Qwant Maps') %>" autocomplete="off">
           <button id="clear_button_mobile" class="search_form__clear icon-x" type="button" onmousedown="clearSearch(event)" title="<%= _('Clear', 'search bar') %>"></button>
           <input type="submit" value="" class="search_form__action" onclick="submitSearch()" title="<%= _('Search') %>">
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -322,14 +322,13 @@ input[type="search"] {
   }
 
   .search_form__return {
+    display: none;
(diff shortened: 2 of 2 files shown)


bbecquet:
Now that it's clearly linked to the focus and display of the suggestion dropdown, it's confusing that clicking on the arrow clears the entry. We should remove the `clearSearch()` action on this button, so its only purpose is to exit the suggest.
By the way, this is already what is done in the direction inputs: https://github.com/QwantResearch/erdapfel/blob/master/src/panel/direction/DirectionInput.jsx#L132-L135
So it will be a big plus for consistency.

G-Ray:
> Now that it's clearly linked to the focus and display of the suggestion dropdown, it's confusing that clicking on the arrow clears the entry. We should remove the `clearSearch()` action on this button, so its only purpose is to exit the suggest.
> By the way, this is already what is done in the direction inputs: https://github.com/QwantResearch/erdapfel/blob/master/src/panel/direction/DirectionInput.jsx#L132-L135
> So it will be a big plus for consistency.

Thanks, I removed the call to `clearSearch`.

views/index.ejs
```diff
@@ -45,7 +45,7 @@
         <div class="search_form__logo"></div>
         <div id="react_menu__container"></div>
         <div class="search_form__wrapper empty">
-          <div class="search_form__return icon-arrow-left" onmousedown="clearSearch(event, true)"></div>
+          <div class="search_form__return icon-arrow-left"></div>
```
amatissart, line 48: This situation with a `div` with no event attached to it caused issues on some mobile devices where the click on this pseudo-button was difficult to trigger. I would suggest to use a proper `<button>` here.

G-Ray: I suggest to do this in a separate PR. I'll try to refactor the search bar layout.

bbecquet: @amatissart @G-Ray
I agree it should be a real `<button>`, but I just tested on the iPhone and I don't see the problem re-appearing here with the `<div>`.
There is however a problem related to the very weird active area of the element, which is very misleading. Most often I focused the input while my finger seemed on the `<div>`.
Capture d’écran de 2020-12-04 16-58-13

I tested with a real `<button>` to be sure, but there is no improvement on this behavior, which already exists in prod.
We should definitely improve this element layout, but it's ok for me to do it in a separate PR.

bbecquet (approved):

PR #960, opened 2020-12-04
[Tests] Run integration suite on desktop and mobile devices

bbecquet:
## Description
I modified the integration test configuration so it accepts a new env var `TEST_DEVICE`, with values `mobile` or `desktop`.
This env var is used for 2 things:
 - to set the default viewport of all Puppeteer browser instances to a corresponding configuration 
 - to filter out some tests depending on the platform if they are not compatible with it.

This filter mechanism uses the file name:
 - `feature.js`: tests in it will be run on `mobile` and `desktop` configs
 - `feature.mobile.js`: tests in it will be run only on `mobile` config
 - `feature.desktop.js`: tests in it will be run only on `desktop` config

I moved some tests to `*.mobile.js` and `*.desktop.js` files to adapt to the new configs. But when possible it would probably be more maintainable to rewrite them so they are adapted to both device types.
It does the job, but in use it's a bit cumbersome to operate on different files to rewrite stuff. And as planned there is a lot of boilerplate code duplication (imports, before/after hooks, etc.). Maybe it's just the set up and it will be fine in practice.

What do you think?

## Why
Increase the coverage and reliability of our tests by forcing tests on mobile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,9 @@
   "scripts": {
     "bench": "node ./bench/performances.js",
     "unit-test": "jest --config=./tests/unit.js",
-    "integration-test": "jest --config=./tests/integration.js",
+    "integration-test-mobile": "TEST_DEVICE=mobile jest --config=./tests/integration.js",
+    "integration-test-desktop": "TEST_DEVICE=desktop jest --config=./tests/integration.js",
+    "integration-test": "npm run integration-test-mobile && npm run integration-test-desktop",
     "test": "npm run unit-test && npm run integration-test",
     "start": "node bin/index.js",
     "build": "webpack --config build/webpack.config.js",
--- a/tests/integration/device.js
+++ b/tests/integration/device.js
@@ -1,0 +1,26 @@
+
+const viewports = {
+  mobile: {
+    width: 400,
+    height: 800,
+    hasTouch: true,
+  },
+};
+
+const ignorePatterns = {
+  mobile: ['.desktop.'],
+  desktop: ['.mobile.'],
+};
+
+function getPupeeterViewport(deviceName) {
+  return viewports[deviceName] || null;
+}
+
+function getIgnorePatterns(deviceName) {
+  return ignorePatterns[deviceName] || [];
+}
+
+module.exports = {
+  getPupeeterViewport,
+  getIgnorePatterns,
+};
--- a/tests/integration/device.js
+++ b/tests/integration/device.js
@@ -6,6 +6,10 @@ const viewports = {
     hasTouch: true,
     isMobile: true,
   },
+  desktop: {
+    width: 800,
+    height: 600,
+  },
 };
 
 const ignorePatterns = {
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -6,7 +6,9 @@
   "scripts": {
     "bench": "node ./bench/performances.js",
     "unit-test": "jest --config=./tests/unit.js",
-    "integration-test": "jest --config=./tests/integration.js",
+    "integration-test-mobile": "TEST_DEVICE=mobile jest --config=./tests/integration.js",
+    "integration-test-desktop": "TEST_DEVICE=desktop jest --config=./tests/integration.js",
+    "integration-test": "npm run integration-test-mobile && npm run integration-test-desktop",
```
G-Ray, line 11: I guess we could event run the mobile and desktop tests in parallel ?

tests/integration/device.js
```diff
@@ -0,0 +1,26 @@
+
+const viewports = {
+  mobile: {
```
G-Ray, line 3: Maybe we should set an explicit viewport size for desktop too ?

tests/integration/device.js
```diff
@@ -0,0 +1,26 @@
...
+    width: 400,
+    height: 800,
+    hasTouch: true,
```
G-Ray, line 6: have you tried the `isMobile` option ? This might have a serious impact on our rendering IIUC https://github.com/puppeteer/puppeteer/blob/v5.5.0/docs/api.md#puppeteerconnectoptions

tests/integration/device.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+};
+
+const ignorePatterns = {
```
G-Ray, line 10: :+1:

G-Ray:
Great ! I think the file naming with our `desktop` or `mobile` suffixes is great. The added boilerplate is sad yes... It might be possible to factorize the boilerplate across a group of tests ?

I like to name group related npm scripts with a comma. This have some use cases like running multiple script in parallel or in sequence with `https://www.npmjs.com/package/npm-run-all` for instance.
We could use:
`integration-test:desktop`, `integration-test:mobile` and `integration-test` to run both.

package.json
bbecquet: Maybe, but we have to be careful not breaking the behavior of the CI with return values.
Let's see later if this brings something.

tests/integration/device.js
bbecquet: Yep, better be explicit. I'll add it.

tests/integration/device.js
bbecquet: I don't think this will change anything for now but clearly I should have set it and forgot. Thanks!

bbecquet:
> I like to name group related npm scripts with a comma.

I didn't use it on purpose. Normally the comma stuff has a precise meaning, which is to add some specificity to a more generic script. For example by adding some CLI parameters.
We are not totally in this case here. The `integration-test` script runs 2 instances of the others, it's not a less-specific version.

G-Ray (approved):

G-Ray (approved):
I think we can merge this :) :clap:

PR #965, opened 2020-12-07
Improve Search form layout

G-Ray:
## Description
- Avoid absolute positioning of elements in Search form
- Height fixes (height of 48px vs 46px previously)
- Use spacing variables
- Use flex layout
- Colors fixes
- Centering fixes, thanks to flexbox

## Why
- Search layout was a bit hard to maintain, due to absolute positioning of element.
- Some element were not perfectly centered due to absolute positioning.

## Screenshots
desktop
mobile

### DIFF
--- a/src/scss/includes/menu.scss
+++ b/src/scss/includes/menu.scss
@@ -1,10 +1,6 @@
 
 #react_menu__container {
-  align-self: stretch;
   width: 40px;
-  position: absolute;
-  right: 0;
-  top: 24px;
 }
 
 .no-burger #react_menu__container {
--- a/src/scss/includes/menu.scss
+++ b/src/scss/includes/menu.scss
@@ -1,6 +1,8 @@
 
 #react_menu__container {
-  width: 40px;
+  width: $spacing-xxl-4;
+  height: $spacing-xxl-4;
+  flex-shrink: 0;
 }
 
 .no-burger #react_menu__container {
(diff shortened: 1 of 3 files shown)


bbecquet (commented):
This is great clean-up!
I just noticed a visual and interactive area change on the right button, itinerary on desktop and burger menu on mobile.
On your new implementation, the right area is larger than before, a bit too large compared to the designs it seems, and a the same the button it contains is smaller, reducing it usability.
|Master branch|This branch|
|---|---|
|Capture d’écran de 2020-12-08 15-03-08|Capture d’écran de 2020-12-08 15-02-59|
|Capture d’écran de 2020-12-08 15-01-31|Capture d’écran de 2020-12-08 15-01-42|





I know we fine-tuned this area so it's as small as possible while the button in it is as big as possible. Would it be possible to restore that while keeping most of your improvements?

G-Ray:
I updated the PR with buttons size of `48px`.

bbecquet:
> I updated the PR with buttons size of `48px`. Theys should also be squared now.

It seems you forgot to adjust some right margin on mobile though: 
Capture d’écran de 2020-12-10 10-39-25

Otherwise everything seems great and cleaner :+1:

G-Ray:
> > I updated the PR with buttons size of `48px`. Theys should also be squared now.
> 
> It seems you forgot to adjust some right margin on mobile though:
> Capture d’écran de 2020-12-10 10-39-25
> 
> Otherwise everything seems great and cleaner

thanks, should be fixed by 53b115bac1054aa05c5e23c174dacc745a22012c

bbecquet (approved):

PR #966, opened 2020-12-07
[Clean-up] Simplify favorite POI code

bbecquet:
## Description
Simplify 2 things related to how the POI favorite Y/N status is handled by the POI panel:
 - now that the store is solely based on local storage, it is synchronous so we don't need extra code to wait for the status ;
 - we don't need two separates events to signal POI added and POI removed from the store, one with a boolean status is enough and brings simplification.

## Why
Code simplification.

### DIFF
--- a/src/panel/poi/PoiPanel.jsx
+++ b/src/panel/poi/PoiPanel.jsx
@@ -21,15 +21,6 @@ import { Flex, Panel, PanelNav, CloseButton, Divider } from 'src/components/ui';
 
 const covid19Enabled = (nconf.get().covid19 || {}).enabled;
 
-async function isPoiFavorite(poi) {
-  try {
-    const storePoi = await isInFavorites(poi);
-    return !!storePoi;
-  } catch (e) {
-    return false;
-  }
-}
-
 export default class PoiPanel extends React.Component {
   static propTypes = {
     poiId: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


src/panel/poi/PoiPanel.jsx
```diff
@@ -57,15 +48,9 @@ export default class PoiPanel extends React.Component {
...
     !this.props.pois ? this.loadPoi() : this.loadPois();
 
-    this.storeAddHandler = listen('poi_added_to_favs', poi => {
-      if (poi === this.state.fullPoi) {
-        this.setState({ isPoiInFavorite: true });
-      }
-    });
-
-    this.storeRemoveHandler = listen('poi_removed_from_favs', poi => {
+    this.storeAddRemoveHandler = listen('poi_favorite_state_changed', (poi, isPoiInFavorite) => {
```
G-Ray, line 51: What about a function name like `storePoiChangeHandler`, somethink like this ?

bbecquet: Not sure I understand, can you write some code?

G-Ray: I just suggest to rename the handler: `storeAddRemoveHandler` -> `storePoiChangeHandler`. To reflect its event name `poi_favorite_state_changed` more closely. Wdyt ?

bbecquet: Alright. Thanks for the explanation!

bbecquet: Done :)

G-Ray (approved):
:+1:

PR #967, opened 2020-12-08
Limit POI panel description to 4 lines

bbecquet:
## Description
Limit the optional description block in POI panel to 4 lines maximum.

Uses the [`-webkit-line-clamp` property](https://css-tricks.com/almanac/properties/l/line-clamp/), which is non-standard but [well-supported by modern browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp#Browser_compatibility), even Firefox (tested on Android and desktop) and Safari (tested on iPhone and another Webkit-based desktop browser).
For such non-critical feature, it's ok and great to avoid a JS-based solution.

## Why
UI balance.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran 2020-12-08 à 14 48 22|Capture d’écran 2020-12-08 à 14 47 47|

### DIFF
--- a/src/scss/includes/mixins.scss
+++ b/src/scss/includes/mixins.scss
@@ -1,0 +1,6 @@
+@mixin lineClamp($lineCount) {
+  display: -webkit-box;
+  -webkit-line-clamp: $lineCount;
+  -webkit-box-orient: vertical;  
+  overflow: hidden;
+}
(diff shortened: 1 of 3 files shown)


G-Ray (approved):
Nice

PR #968, opened 2020-12-08
Qmaps 1713 favorite icon

G-Ray:
## Description
- Introduce Icons exports. Svg icons are included inline with `@svgr/webpack` loader. This way, we can pass props like `color` or `fill` props.
- Button component now accept two type of icons : a string or a node.
- Replace favorite icon with an heart icon, and colorize it.

## Why
UI/UX. First steps to introduce a new icon stack, without using fonts icons.

## Screenshots
Peek 2020-12-09 11-54


public/images/feather/heart.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="feather feather-heart"><path d="M20.84 4.61a5.5 5.5 0 0 0-7.78 0L12 5.67l-1.06-1.06a5.5 5.5 0 0 0-7.78 7.78l1.06 1.06L12 21.23l7.78-7.78 1.06-1.06a5.5 5.5 0 0 0 0-7.78z"></path></svg>
```
G-Ray, line 1: We can discuss about folder hierarchy, here, I created `/public/images/feather folder` for this icon only.

xem (commented):
I find this approach very clean and convenient, OK for me
Just a remark, wouldn't it be better to introduce the component Icon in a separate file than Button.jsx? Not all icons belong to buttons (ex: itinerary fields icons)

xem:
@G-Ray @bbecquet @amatissart regarding the folder's name and hierarchy, I suggest that we don't name it "feather", because it can contain icons coming from other sources or from our designers.
Also, there are cases where these icons are not used inside buttons but as css backgrounds. So, the folder must indeed be in /public.
How about  /public/icons/  instead of /public/images/feather ?
Let's debate :D

public/images/feather/heart.svg
amatissart: I have no strong opinion about the folder name. Mentioning "feather" may not be necessary unless we plan to use these icons specifically in the future.

I am a bit reluctant to store these icons in the "public" folder though. Other "public" files are compressed in the Docker image, to be served directly via an explicit HTTP request. But these icons work differently. They are bundled in the .js files at build time. So in my opinion, it would be clearer to put them in "src".
And that would help to keep in mind these files behave like other .js dependencies, eg they should be bundled separately and loaded lazily if they become too heavy.

G-Ray: I agree with all your points. Storing svg inside /public folder would only make sense if we need them to be publicly accessible.
So, what about storing all new svg icons inside `src/icons` ?

G-Ray:
> I find this approach very clean and convenient, OK for me
> Just a remark, wouldn't it be better to introduce the component Icon in a separate file than Button.jsx? Not all icons belong to buttons (ex: itinerary fields icons)

Icon is just a quick way for a Button to render the icon depending if it's an icon font (a string is passe as props), or any other react node. I don't think we need to expose this little util component.

How about /public/icons/ instead of /public/images/feather ?
@amatissart pointed out relevant considerations about storing icons inside the public folder. Let's continue the discussion here https://github.com/QwantResearch/erdapfel/pull/968#discussion_r538560003

public/images/feather/heart.svg
bbecquet: I agree we don't need to mention the origin set, `feather` here, in the folder, it may be more confusing than helpful.

About where icons are stored, I wouldn't mind if they where in `/public/`, ready to be used as real images if needed like @xem says. But there is no use case right now, at least in this PR, so ok to move on with `/src/icons` here.
And in anycase, thanks to the file responsible for the imports, it's completely possible to have a mix of imports of SVG files from `public` and `src` if we need that in the future.

src/components/ui/Icons.jsx
```diff
@@ -0,0 +1 @@
+export { default as Heart } from '../../../public/images/feather/heart.svg';
```
bbecquet, line 1: It would be more logical to name this file `icons.js`, as it isn't the definition of a React component.

src/panel/poi/ActionButtons.jsx
```diff
@@ -42,9 +47,17 @@ const ActionButtons = ({
...
       className="poi_panel__action__favorite"
-      onClick={toggleStorePoi}
-      icon={isPoiInFavorite ? 'icon_star-filled' : 'star'}
+      data-active={isPoiInFavorite}
       title={_('Favorites', 'poi panel')}
+      onClick={toggleStorePoi}
+      style={{ borderColor: favoriteColor }}
+      icon={
+        <Heart
+          width={16}
```
bbecquet, line 56: It would be nice to find a way to omit this explicit size, but let's move on and think about this later :+1:

src/panel/poi/ActionButtons.jsx
```diff
@@ -42,9 +47,17 @@ const ActionButtons = ({
...
+          color={favoriteColor}
+          fill={favoriteColor || 'white'}
```
bbecquet, line 58: It doesn't change anything visually here, but technically the `fill` in this case should be `none` or `transparent`.

src/components/ui/Icons.jsx
G-Ray: Ok I renamed the file to icon.js, but let it in same folder if it's ok ?

src/panel/poi/ActionButtons.jsx
G-Ray: Peek 2020-12-10 15-42
actually it does have a visual impact, thanks !

bbecquet (approved):

PR #969, opened 2020-12-09
Replace travis with a github workflow

amatissart:
## Why
OSS credits are no longer renewed on Travis free plan.


.tx/push.sh
```diff
@@ -6,6 +6,6 @@ npm run i18n
 # Remove msgstr from en.po so that it's considered as a .pot template file by Transifex
 msgfilter --keep-header -i language/message/en.po -o language/message/en.po true
 
-python3.5 -m pip install --user transifex-client==0.13.12
+sudo pip3 install transifex-client==0.13.12
```
G-Ray, line 9: Why do we need to run this cmd as root ?

amatissart: When installed as a non-root user, the `tx` command is present in `~/.local/bin` which is apparently not added to the PATH in the ubuntu image used by the workflow.  
So I preferred to stick to the documentation: https://docs.transifex.com/client/installing-the-client

G-Ray (approved):

bbecquet (approved):
:heart:

PR #971, opened 2020-12-10
Add an optional env name in the page title

bbecquet:
## Description
Small tweak to add an optional `envName` config value in the page title.
The idea is you add such name in your `.env` file so it's easier to differentiate local and prod (or plive, or dev, if we change the config), especially on mobile tab switching where the url is not displayed.

## Screenshots
|Before|After|
|---|---|
|IMG_0011|IMG_0012|
|Screenshot_20201210-173615_Chrome|Screenshot_20201210-173658_Chrome|

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -4,6 +4,7 @@
     <meta charset='utf-8'>
     <title><%= _('Qwant Maps') %>
       <% if(config.app.versionFlag) { %> - <%= config.app.versionFlag %><% } %>
+      <% if(config.envName) { %> - <%= config.envName %><% } %>
     </title>
     <meta name="description" content="<%= _('The map that respects your privacy') %>">
     <base href="<%= config.system.baseUrl %>" target="_blank">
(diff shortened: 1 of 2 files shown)


G-Ray:
Nice suggestion ! 
It makes me remind on thing : Shouldn't we add an `env.example` with dummy values to the repo ?

G-Ray (approved):

views/index.ejs
```diff
@@ -4,6 +4,7 @@
     <meta charset='utf-8'>
     <title><%= _('Qwant Maps') %>
       <% if(config.app.versionFlag) { %> - <%= config.app.versionFlag %><% } %>
+      <% if(config.envName) { %> - <%= config.envName %><% } %>
```
amatissart, line 7: I would recommend to add `envName` to default_config.yml in order to make all allowed keys explicit.

bbecquet: Done, thanks for the suggestion.

G-Ray (approved):

amatissart (approved):

PR #973, opened 2020-12-10
Add a drop shadow on top bar when the suggestions are scrolled

bbecquet:
## Description
Adds a drop shadow on input bars when the suggest list scrolls "under" it.
Do it on the main search bar and on the direction input, which are still 2 different components even if looking the same.

Define a new hook to ease such operations as this effect is to be re-used on other screens of the app.

**This doesn't work yet on iOS, as when the keyboard is open the whole body is scrolling. This is unrelated to this PR, I'll try to fix it separately.**

## Screenshots
Peek 14-12-2020 10-10

### DIFF
--- a/src/hooks/useScrollWatcher.js
+++ b/src/hooks/useScrollWatcher.js
@@ -14,8 +14,8 @@ function useScrollWatcher(ref, cb, throttlingTimeout = 0) {
     scrollingElement.addEventListener('scroll', onScroll);
 
     return () => {
-      cb(0);
       scrollingElement.removeEventListener('scroll', onScroll);
+      cb(0);
     };
   }, [ref, cb, throttlingTimeout]);
 }
--- a/src/hooks/useScrollWatcher.js
+++ b/src/hooks/useScrollWatcher.js
@@ -1,0 +1,23 @@
+import { useEffect } from 'react';
+import throttle from 'lodash.throttle';
+
+function useScrollWatcher(ref, cb, throttlingTimeout = 0) {
+  useEffect(() => {
+    if (!cb) {
+      return;
+    }
+    const scrollingElement = ref.current;
+    const onScroll = throttle(() => {
+      cb(scrollingElement.scrollTop);
+    }, throttlingTimeout);
+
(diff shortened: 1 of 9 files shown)


src/hooks/useScrollWatcher.js
```diff
@@ -0,0 +1,23 @@
+import { useEffect } from 'react';
+import throttle from 'lodash.throttle';
+
+function useScrollWatcher(ref, cb, throttlingTimeout = 0) {
```
G-Ray, line 4: :+1:

src/hooks/useScrollWatcher.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+    const scrollingElement = ref.current;
+    const onScroll = throttle(() => {
+      cb(scrollingElement.scrollTop);
+    }, throttlingTimeout);
+
+    scrollingElement.addEventListener('scroll', onScroll);
+
+    return () => {
+      cb(0);
```
G-Ray, line 17: Should it be safer to call `cb(0)` after having removed the listener ? I'm afraid there is a probability where a `onScroll` event could be fired between this line and next one.

amatissart (commented):
The implementation looks good to me, but it seems the shadow would actually be visible in very rare occasions, depending on the screen size.

On my Android device (5.8"), suggestions are never scrollable. Even if the last suggestion is not completely visible, no scroll is possible, and I need to collapse the keyboard to see the whole list.


(Similar behavior on both Chrome and Firefox).  
Is this behavior expected at this point ?  
Maybe it could be solved at the same time as the problem you describe on iOS ?

bbecquet:
@amatissart Thanks a lot for your feedback.
With my small screen Android device, I didn't notice the problem. On it, the keyboard hides half of the screen and I can scroll the list and trigger the shadow effect. But the list just scroll partly, and the last result is still hidden behind the keyboard.
So I guess like you suggested we have to fix this bigger problem before dealing with shadows. I'm putting this PR on hold.

xem:
@bbecquet may I suggest to try a pure CSS solution?
https://lea.verou.me/2012/04/background-attachment-local/

PR #976, opened 2020-12-15
Do not force minimize panel when unnecessary

bbecquet:
## Description
Improvement over https://github.com/QwantResearch/erdapfel/pull/937: avoid unnecessary re-renders by not calling `setState` when it's not needed.

Helps with the bug of the map not being movable on the step-by-step feature (see https://github.com/QwantResearch/erdapfel/pull/953#issuecomment-745209002), although the first time movement will still be imposssible… :thinking:

### DIFF
--- a/src/panel/PanelManager.jsx
+++ b/src/panel/PanelManager.jsx
@@ -63,7 +63,7 @@ export default class PanelManager extends React.Component {
     window.times.appRendered = Date.now();
 
     listen('map_user_interaction', () => {
-      if (isMobileDevice()) {
+      if (isMobileDevice() && this.state.panelSize !== 'minimized') {
         this.setState({ panelSize: 'minimized' });
       }
     });
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -44,7 +44,10 @@ const MobileRoadMapPreview = ({
     () => {
       fire('zoom_step', steps[currentStep]);
     },
-    [currentStep, steps]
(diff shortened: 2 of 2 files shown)


amatissart (dismissed):
:+1:

To solve the bug introduced in #953, would it make sense to trigger the map move only when ` currentStep` has changed ? 
As far as I understand, `steps` is re-computed for each render of the DirectionPanel (via `getAllSteps()`), but its content will not change anyway before the `MobileRoadMapPreview` is unmounted.

```diff
diff --git a/src/panel/direction/MobileRoadMapPreview.jsx b/src/panel/direction/MobileRoadMapPreview.jsx
index 8476aad7..027ae7de 100644
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -42,11 +42,11 @@ const MobileRoadMapPreview = ({
   // When currentStep is modified, zoom on it
   useEffect(
     () => {
       fire('zoom_step', steps[currentStep]);
     },
-    [currentStep, steps]
+    [currentStep]
   );
 
   return <div className="itinerary_mobile_step_by_step">
     <Button
       className="mobile-roadmap-preview-close"
```

bbecquet:
@amatissart 
Yes it would work in our case, but it will trigger a linter warning if we remove `steps` in the watched values array, as conceptually the `useEffect` content **should** depend on `steps`, to prevent UI not consistent with data state.
But I agree in this case we don't risk falling into an inconsistent case, the `steps` content will indeed be the same. I'll add a linter exception with explanation.

amatissart (approved):

PR #977, opened 2020-12-15
QMAPS-1873 add infobox on first visit for foreign browsers coming from Qwant Search IAs

xem:
## Description
Show this popup if browser language is not "fr" AND first visit (or popup not closed in previous visits), AND the user arrived from a Qwant Search IA

## Screenshots

image
image

### DIFF
--- a/src/adapters/store.js
+++ b/src/adapters/store.js
@@ -119,3 +119,21 @@ export async function setLastLocation(loc) {
     throw e;
   }
 }
+
+export function getBetaPopupClosed() {
+  try {
+    return get('beta_popup_closed');
+  } catch (e) {
+    Error.sendOnce('store', 'getBetaPopupClosed', 'error getting beta popup closed', e);
+    return null;
+  }
+}
+
+export function setBetaPopupClosed() {
+  try {
+    return set('beta_popup_closed', 1);
+  } catch (e) {
+    Error.sendOnce('store', 'setBetaPopupClosed', 'error setting beta popup closed', e);
+    throw e;
+  }
+}
--- a/src/adapters/store.js
+++ b/src/adapters/store.js
@@ -119,21 +119,3 @@ export async function setLastLocation(loc) {
     throw e;
   }
 }
-
-export function getBetaPopupClosed() {
-  try {
-    return get('beta_popup_closed');
-  } catch (e) {
-    Error.sendOnce('store', 'getBetaPopupClosed', 'error getting beta popup closed', e);
-    return null;
-  }
-}
-
-export function setBetaPopupClosed() {
-  try {
-    return set('beta_popup_closed', 1);
-  } catch (e) {
-    Error.sendOnce('store', 'setBetaPopupClosed', 'error setting beta popup closed', e);
-    throw e;
-  }
-}
--- a/src/components/BetaInfoBox.jsx
+++ b/src/components/BetaInfoBox.jsx
@@ -9,7 +9,7 @@ const initialQueryParams = parseQueryString(window.location.search);
 
 const BetaInfoBox = () => {
   const [closed, setClosed] = useState(get('beta_popup_closed'));
-  const [client] = useState(initialQueryParams['client']);
+  const client = initialQueryParams['client'];
   const closeBetaPopup = () => {
     set('beta_popup_closed', 1);
     setClosed(true);
--- a/src/components/BetaInfoBox.jsx
+++ b/src/components/BetaInfoBox.jsx
@@ -1,0 +1,44 @@
+/* global _ */
+import React, { useState } from 'react';
+import Alert from './ui/Alert';
+import { Button } from './ui';
+import { getBetaPopupClosed, setBetaPopupClosed } from 'src/adapters/store';
+import { parseQueryString } from '../libs/url_utils';
+
+const initialQueryParams = parseQueryString(window.location.search);
+
+const BetaInfoBox = () => {
+  const [closed, setClosed] = useState(getBetaPopupClosed());
+  const [client] = useState(initialQueryParams['client']);
+
+  const closeBetaPopup = () => {
+    setBetaPopupClosed();
+    setClosed(true);
+  };
+
+  const footer = <div className="alert-link">
+    <Button variant="tertiary" onClick={closeBetaPopup}>
+      {_('Do not display this message again')}
+    </Button>
+  </div>;
+
+  return (
+    // Show beta popup if browser language is not french, if popup hasn't already been closed, and if the user comes from an IA
+    !navigator.language.includes('fr')
+    && !closed
+    && client === 'search-ia-local'
+    &&
+    <Alert
+      className=""
+      title="Qwant Maps is in Beta!"
+      /* eslint-disable max-len */
+      description={_('This means that this version may have some bugs. We work very hard to improve Qwant Maps every day, while keeping your travels private.')}
+      /* eslint-enable max-len */
+      type="info"
+      onClose={closeBetaPopup}
+      footer={footer}
+    />
+  );
+};
+
+export default BetaInfoBox;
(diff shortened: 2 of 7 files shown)


src/components/ui/Alert.jsx
```diff
@@ -28,6 +29,11 @@ const Alert = ({
     <div className="alert-content">
       <span role="alert">{description}</span>
     </div>
+    <div className="alert-link">
+      <a onClick={onClose}>
+        { _('Do not display this message again') }
```
bbecquet, line 34: If we want to re-use the Alert component, this shouldn't be hardcoded but passed as optional prop.

src/components/ui/Alert.jsx
```diff
@@ -28,6 +29,11 @@ const Alert = ({
     <div className="alert-content">
       <span role="alert">{description}</span>
     </div>
+    <div className="alert-link">
+      <a onClick={onClose}>
```
bbecquet, line 33: If there is no href this isn't a link (and it won't display a pointer cursor).
However, this looks exactly like a `<Button variant="tertiary">`.

src/panel/PanelManager.jsx
```diff
@@ -31,13 +34,16 @@ export default class PanelManager extends React.Component {
       ActivePanel: ServicePanel,
       options: {},
       panelSize: 'default',
+      betaPopupClosed: getBetaPopupClosed(),
+      client: initialQueryParams['client'],
     };
   }
 
-  componentDidMount() {
+  async componentDidMount() {
```
bbecquet, line 42: This can remain sync.

src/panel/PanelManager.jsx
```diff
@@ -211,13 +217,35 @@ export default class PanelManager extends React.Component {
...
+    const { ActivePanel, options, panelSize, betaPopupClosed, client } = this.state;
 
     return <PanelContext.Provider value={{ size: panelSize, setSize: this.setPanelSize }} >
       <div className="panel_container">
         <ActivePanel {...options} />
       </div>
+
+      {
+        // Show beta popup if browser language is not french, if popup hasn't already been closed, and if the user comes from an IA
+        !navigator.language.includes('fr')
```
bbecquet, line 235: As the condition and state management is complex enough, why not encapsulating all this in a dedicated component? I suggest a name like `<BetaDisclaimer>`.
You could move everything related to it in it and render `null` when it's not needed. That way you could just render it without condition and wouldn't need to change anything in the props and state of `PanelManager`, as it's already tricky.
BTW, as this disclaimer isn't related to the panels at all, it would make even more sense to put it in `RootComponent` instead of `PanelManager`.

src/adapters/store.js
```diff
@@ -119,3 +119,21 @@ export async function setLastLocation(loc) {
     throw e;
   }
 }
+
+export function getBetaPopupClosed() {
```
G-Ray, line 123: I would move these function into `BetaInfoBox.jsx`.
IMO `store.js` should only expose a limited set of functions in order to abstract the way we store our persisted data (through localstorage).

If, at a later stage, we need to reuse these functions, then we will move theme into a new lib file like `betaPopup.js` for instance.

What do you think ?

src/components/BetaInfoBox.jsx
```diff
@@ -0,0 +1,44 @@
...
+    </Button>
+  </div>;
+
+  return (
+    // Show beta popup if browser language is not french, if popup hasn't already been closed, and if the user comes from an IA
+    !navigator.language.includes('fr')
+    && !closed
+    && client === 'search-ia-local'
+    &&
+    <Alert
```
G-Ray, line 31: :+1: Glad to see this component is reused after Masq :smiling_face_with_three_hearts: 

className should be optional, we can remove this prop

src/components/BetaInfoBox.jsx
```diff
@@ -0,0 +1,44 @@
...
+      className=""
+      title="Qwant Maps is in Beta!"
```
amatissart, line 33: A translation is missing here

src/components/BetaInfoBox.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const footer = <div className="alert-link">
+    <Button variant="tertiary" onClick={closeBetaPopup}>
+      {_('Do not display this message again')}
+    </Button>
+  </div>;
+
+  return (
+    // Show beta popup if browser language is not french, if popup hasn't already been closed, and if the user comes from an IA
+    !navigator.language.includes('fr')
```
amatissart, line 27: For consistency, I would prefer to use `window.getLang()` which is already defined by the application

src/panel/PanelManager.jsx
```diff
@@ -17,8 +17,8 @@ import { PanelContext } from 'src/libs/panelContext.js';
 import NoResultPanel from 'src/panel/NoResultPanel';
 
 const directionConf = nconf.get().direction;
-
 const directSearchRouteName = 'Direct search query';
+const initialQueryParams = parseQueryString(window.location.search);
```
bbecquet, line 21: There is no more reason to put that outside `componentDidMount`. You can simply revert this file to its master version.

src/components/BetaInfoBox.jsx
```diff
@@ -0,0 +1,44 @@
...
+    &&
+    <Alert
+      className=""
+      title="Qwant Maps is in Beta!"
+      /* eslint-disable max-len */
+      description={_('This means that this version may have some bugs. We work very hard to improve Qwant Maps every day, while keeping your travels private.')}
+      /* eslint-enable max-len */
+      type="info"
+      onClose={closeBetaPopup}
+      footer={footer}
```
bbecquet, line 39: No big deal, but you can also render the footer directly here, this is a common practice to pass JSX as prop.

src/components/BetaInfoBox.jsx
bbecquet: Also, we should still avoid `includes` on arrays and strings until we drop the support for IE11.

BTW, I've opened an issue to track the things we'll be able to do then, feel free to suggest others I forgot :) https://github.com/QwantResearch/erdapfel/issues/978

src/components/BetaInfoBox.jsx
```diff
@@ -0,0 +1,44 @@
+/* global _ */
+import React, { useState } from 'react';
+import Alert from './ui/Alert';
+import { Button } from './ui';
+import { getBetaPopupClosed, setBetaPopupClosed } from 'src/adapters/store';
+import { parseQueryString } from '../libs/url_utils';
+
+const initialQueryParams = parseQueryString(window.location.search);
+
+const BetaInfoBox = () => {
```
bbecquet, line 10: Cool! This is easier to maintain with this form, especially as a functional component.
Thanks for making this change :+1:

src/adapters/store.js
xem: Sure, I did it like that to maintain the existing organization of store.js (containing all sorts of getters and setters for other components), but I can make something more generic, promise-less, and use it in my Alert.

src/components/BetaInfoBox.jsx
```diff
@@ -0,0 +1,43 @@
...
+import Alert from './ui/Alert';
+import { Button } from './ui';
+import { get, set } from 'src/adapters/store';
+import { parseQueryString } from '../libs/url_utils';
+
+const initialQueryParams = parseQueryString(window.location.search);
+
+const BetaInfoBox = () => {
+  const [closed, setClosed] = useState(get('beta_popup_closed'));
+  const [client] = useState(initialQueryParams['client']);
```
bbecquet, line 12: Just a remark for your next functional components: as you don't allow it to be changed from the component itself (there is no setter), it doesn't need to use a state for that. You could just extract the value as a normal const in the function.

bbecquet (approved):

bbecquet (approved):

PR #979, opened 2020-12-16
Qmaps 1895 - fix state loop with a window.app.navigateBack method

G-Ray:
## Description
- Suggest an api to go back to previous state:
`back` would try to use history.back if there is a previous state, and would replace the state as a fallback method.
the suggested method' signature is :
```js
back({ relativeUrl = '/', state = {} }) {
```

## Why
Short term : Avoid current issue where the app state goes into an infinite loop when toggling direction details.
mid-term : Use a new api to normalize the way we handle going back across our app.

### DIFF
--- a/src/panel/app_panel.js
+++ b/src/panel/app_panel.js
@@ -89,7 +89,7 @@ export default class App {
    * @param {Object} fallback.relativeUrl - The relativeUrl to fallback to if history.back is unavailable
    * @param {Object} fallback.state - The state to fallback to if history.back is unavailable
    */
-  back({ relativeUrl = '/', state = {} }) {
+  navigateBack({ relativeUrl = '/', state = {} }) {
     if (history.state !== null) {
       window.history.back();
     } else {
--- a/src/panel/app_panel.js
+++ b/src/panel/app_panel.js
@@ -84,9 +84,9 @@ export default class App {
 
   /**
    * Go to the previous application state using history.back()
-   * If no previous state is available, use replaceState as a fallback.
+   * If we determine history.back would exit the application, use replaceState as a fallback.
    * @param {Object} fallback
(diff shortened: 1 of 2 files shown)


src/panel/app_panel.js
```diff
@@ -81,6 +81,20 @@ export default class App {
...
 
+  /**
+   * Go to the previous application state using history.back()
+   * If no previous state is available, use replaceState as a fallback.
+   * @param {Object} fallback
+   * @param {Object} fallback.relativeUrl - The relativeUrl to fallback to if history.back is unavailable
+   * @param {Object} fallback.state - The state to fallback to if history.back is unavailable
+   */
+  back({ relativeUrl = '/', state = {} }) {
+    if (history.state !== null) {
```
G-Ray, line 93: This is rather fragile.
This can only ensure the app has been loaded into a new tab without previous history.

bbecquet: > This can only ensure the app has been loaded into a new tab without previous history.

This doesn't ensure that, because we don't have access to previous history entries for security reason. It would be the same if coming from another website which has a link to an "internal" state of Qwant Maps.

It works because our internal router always put at least an empty object as state, so a `null` means we didn't use `navigateTo` to create the current history state.

This is fragile indeed because it relies on the fact that we don't add anything to the state when using `updateUrl` while loading the results AND the state keeps its `null` value, ignoring the default of `{ }` specified in the `navigateTo` arguments because it works only for `undefined`.

On the other hand, it works for our current bug :)
But maybe we could find a way to make it more fragile… Maybe by always setting a history state value in `navigateTo` used in `replace` mode, and check it instead of `null`.

bbecquet (approved):
Thanks @G-Ray :+1: This fixes the very annoying bug, so let's merge as-is and we'll try to make it more robust later.

PR #980, opened 2020-12-16
Direction: center route on label click, on mobile

G-Ray:
## Description
On mobile, when clicking on a route label, the map was not recentered on the itinerary.

## Why
Consistency between mobile/desktop, and UX


xem (approved):

bbecquet:
If I remember correctly, the design instructions are now to just force fit only on the first display, and to not disturb the map moves made by the user after.
I'll try to find a ref for that.

bbecquet (approved):

PR #982, opened 2020-12-18
Fix the parsing of line/outline route colors

bbecquet:
## Description
Fix the assignment of line colors on route GeoJSON features, managing the case where hex colors are missing the `#` for some providers.


G-Ray (commented):
great

amatissart (approved):

PR #987, opened 2020-12-22
SuggestItem: use magnifier icon as fallback for categories without icon

amatissart:
## Why
Some intentions may be related to categories without explicit icon

## Screenshots

|Before|After|
|-|-|
|image|image|


bbecquet (approved):
Maybe one day I'll remember this special case -_-

PR #988, opened 2020-12-22
Simplify display of favorites in the suggest dropdown

bbecquet:
## Description
Remove the named section "Favorites" in the suggestion dropdown, which simplifies the suggest code as it was kind of a hack.

## Why
Design.
Now that favorite suggestions have their own icon, we don't need to distinguish them even more by a dedicated section.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2020-12-22 17-43-40|Capture d’écran de 2020-12-22 17-43-16|

### DIFF
--- a/src/components/ui/SuggestItem.jsx
+++ b/src/components/ui/SuggestItem.jsx
@@ -1,5 +1,6 @@
 /* global _ */
 import React from 'react';
+import classnames from 'classnames';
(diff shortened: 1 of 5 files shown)


G-Ray (approved):
the (my) code was quite atrocious, thanks :)

PR #990, opened 2020-12-23
QMAPS-1920 various CSS adjustments around the input fields (PC/mobile)

xem:
## Description
- Align the input fields and icons better, and avoid overlaps
- Fix some colors and sizes too, according to the mockups

## Screenshots

Mobile:
Pas focus
image
Focus champ principal / itinéraire
image

PC:
Itinéraire:
image

Home:
image
image

### DIFF
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -23,7 +23,7 @@
 .search_form__input {
   min-width: 0;
   width: calc(100% - 30px);
-  height: 46px;
+  height: $spacing-xxl-4;
   font-size: 16px;
   font-weight: normal;
   color: $grey-black;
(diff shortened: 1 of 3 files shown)


src/scss/includes/search_form.scss
```diff
@@ -23,7 +23,7 @@
 .search_form__input {
   min-width: 0;
   width: calc(100% - 30px);
-  height: $spacing-xxl-4;
+  height: 46px;
```
G-Ray, line 26: height seems to be 48px to me, at least on desktop mockups, as it's a value from design system

xem: sorry indeed, I forgor to reset my OS zoom before touching that, it's what made me think it's misaligned

bbecquet (approved):

PR #991, opened 2020-12-28
mobile direction shortcut: set poi state if any

G-Ray:
## Description
On mobile, when redirecting to `/routes`, send poi in state if it exists. This is the case when the sortcut is cliked from a minimized POI panel.
This way, destination input can be automatically filled.

## Why
UX


xem (approved):

amatissart:
I am not sure about this change. Reading `window.history.state` in this context looks fragile.  
Maybe we could add a test case for now, to ensure that a regression will not go unnoticed ?

G-Ray:
This PR is superseeded by https://github.com/QwantResearch/erdapfel/pull/999

PR #992, opened 2020-12-29
Qmaps 1729 proximity panel

G-Ray:
## Description
- position panel as relative to the top bar
- Toggle service panel visibility as long as no query value has been typed

I'm not sure about the UX. When a user has multiple favorites, the service panel might be displayed too "low".

## Why
UX

## Screenshots
Peek 2020-12-30 15-02

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -16,7 +16,7 @@ const Suggest = ({
   withGeoloc,
   onSelect = selectItem,
   onClear,
-  onToggleSuggestions,
+  onChange,
   className,
 }) => {
   const [items, setItems] = useState([]);
@@ -29,12 +29,6 @@ const Suggest = ({
     setIsOpen(false);
     setItems([]);
   };
-
-  useEffect(() => {
-    if (onToggleSuggestions) {
-      onToggleSuggestions(isOpen);
-    }
-  }, [isOpen, onToggleSuggestions]);
 
   useEffect(() => {
     let currentQuery = null;
@@ -82,6 +76,10 @@ const Suggest = ({
       fetchItems(value);
       setIsOpen(true);
       setLastQuery(value);
+
+      if (onChange) {
+        onChange(value);
+      }
     };
 
     const handleKeyDown = async event => {
@@ -111,7 +109,7 @@ const Suggest = ({
       inputNode.removeEventListener('input', handleInput);
       inputNode.removeEventListener('keydown', handleKeyDown);
     };
-  }, [inputNode, onClear, withCategories, withGeoloc, isMobile]);
+  }, [inputNode, onClear, withCategories, withGeoloc, isMobile, onChange]);
 
   if (!isOpen) {
     return null;
@@ -150,7 +148,6 @@ Suggest.propTypes = {
   withGeoloc: bool,
   onSelect: func,
   onClear: func,
-  onToggleSuggestions: func,
   className: string,
 };
 
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -19,6 +19,7 @@ const Suggest = ({
   onChange,
   className,
   onClose,
+  onOpen,
 }) => {
   const [items, setItems] = useState([]);
   const [isOpen, setIsOpen] = useState(false);
@@ -35,7 +36,11 @@ const Suggest = ({
     if (onClose && !isOpen) {
       onClose();
     }
-  }, [isOpen, onClose]);
+
+    if (onOpen && isOpen) {
+      onOpen();
+    }
+  }, [isOpen, onClose, onOpen]);
 
   useEffect(() => {
     let currentQuery = null;
@@ -155,6 +160,7 @@ Suggest.propTypes = {
   withGeoloc: bool,
   onSelect: func,
   onClear: func,
+  onOpen: func,
   onClose: func,
   onChange: func,
   className: string,
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -18,6 +18,7 @@ const Suggest = ({
   onClear,
   onChange,
   className,
+  onClose,
 }) => {
   const [items, setItems] = useState([]);
   const [isOpen, setIsOpen] = useState(false);
@@ -29,6 +30,12 @@ const Suggest = ({
     setIsOpen(false);
     setItems([]);
   };
+
+  useEffect(() => {
+    if (onClose && !isOpen) {
+      onClose();
+    }
+  }, [isOpen, onClose]);
 
   useEffect(() => {
     let currentQuery = null;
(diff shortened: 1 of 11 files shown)


src/panel/service/ServicePanelDesktop.jsx
```diff
@@ -16,12 +16,12 @@ const ServicePanelDesktop = () => {
...
       />
     </Panel>
-    <div className="u-center">
+    <div className="u-center" style={{ width: 400 }}>
       <Button
         className="service_panel__category_toggle"
         variant="tertiary"
         icon={collapsed ? 'icon_chevron-down' : 'icon_chevron-up'}
-        onClick={() => setCollapsed(!collapsed)}
+        onMouseDown={() => setCollapsed(!collapsed)}
```
G-Ray, line 24: onClick can not triggered if the user click on this button while the top search bar is focus.

bbecquet:
It has a side effect on the direction feature:
localhost_3000_

It uses a CSS exception to hide the top bar in this mode (still the same old story, the top bar is not in the React tree, etc.), so your change is probably incompatible with it.

src/components/ui/Suggest.jsx
```diff
@@ -150,7 +148,6 @@ Suggest.propTypes = {
   withGeoloc: bool,
   onSelect: func,
   onClear: func,
-  onToggleSuggestions: func,
```
bbecquet, line 153: Nit: It should be replaced by `onChange`.

bbecquet (approved):

amatissart (dismissed):
The Panel is no longer visible on mobile after a query is being typed, and the "back" button is pressed:

https://user-images.githubusercontent.com/4726554/103897817-5b5d1000-50f4-11eb-8112-cef99fdf14bc.mp4

G-Ray:
Remaining issues should be fixed. can you please confirm ?

bbecquet (dismissed):
Looks fine to me.

src/panel/PanelManager.jsx
```diff
@@ -214,10 +216,25 @@ export default class PanelManager extends React.Component {
...
     const { ActivePanel, options, panelSize } = this.state;
 
-    return <PanelContext.Provider value={{ size: panelSize, setSize: this.setPanelSize }} >
-      <div className="panel_container">
-        <ActivePanel {...options} />
-      </div>
-    </PanelContext.Provider>;
+    return <div>
+      <Suggest
+        inputNode={this.props.searchBarInputNode}
```
bbecquet, line 221: Can you add this to propTypes please?

src/panel/PanelManager.jsx
```diff
@@ -214,10 +216,25 @@ export default class PanelManager extends React.Component {
...
+        outputNode={document.querySelector('.search_form__result')}
+        withCategories
+        onChange={query => {
+          if (ActivePanel === ServicePanel) {
+            togglePanelVisibility(query.length === 0);
```
bbecquet, line 226: Now that everything is managed in this component with access to all elements, we should be able to get rid of this fragile hiding mechanism and rely on a state value to conditionnaly render the Panel.

src/panel/PanelManager.jsx
```diff
@@ -214,27 +216,41 @@ export default class PanelManager extends React.Component {
...
+        inputNode={searchBarInputNode}
+        outputNode={searchBarOutputNode}
         withCategories
         onChange={query => {
           if (ActivePanel === ServicePanel) {
-            togglePanelVisibility(query.length === 0);
-          } else {
-            togglePanelVisibility(false);
+            this.setState({ isPanelVisible: query.length === 0 });
+          } else if (isPanelVisible) {
```
bbecquet, line 230: Do we really need all these explicit checks on the previous `isPanelVisible` value?
Unless there is a blatant perf problem or side effect, it would be easier to read and understand if you only consider the condition on the panel and let React do its job with state updates.
```js
this.setState({ isPanelVisible: ActivePanel === ServicePanel && query.length === 0 });
```
Same for the other callbacks.

bbecquet: If we encounter perf problems with `setState` calls that trigger too many re-renders, one way to solve it is to switch to a functional component, so state comparisons are on atomic values and won't trigger useless renders.

src/panel/RootComponent.jsx
```diff
@@ -33,7 +33,11 @@ const RootComponent = ({
   });
 
   return <DeviceContext.Provider value={isMobile}>
-    <PanelManager router={router} searchBarInputNode={document.getElementById('search')}/>
+    <PanelManager
+      router={router}
+      searchBarInputNode={document.getElementById('search')}
+      searchBarOutputNode={document.querySelector('.search_form__result')}
```
bbecquet, line 39: :+1:

src/panel/PanelManager.jsx
G-Ray: yes it should works, I was not really focus on optimizing things, instead I tried to make things work at first :zany_face: 
Thank you

src/panel/PanelManager.jsx
```diff
@@ -212,12 +216,39 @@ export default class PanelManager extends React.Component {
...
+    return <div>
+      <Suggest
+        inputNode={searchBarInputNode}
+        outputNode={searchBarOutputNode}
+        withCategories
+        onChange={query => {
+          this.setState({
+            isPanelVisible: ActivePanel === ServicePanel && query.length === 0,
+          });
+        }}
```
amatissart, line 231: As is, the `<Suggest>` component will be re-rendered for each input: its `currentQuery` will not be persisted and the successive queries aren't properly aborted.
These weird race conditions may also explain why the CI is failing.

Maybe we should check the previous state in `setState`, to check if the state really needs to be changed ?
Or maybe `React.memo` should be used in Suggest.jsx ?

src/panel/PanelManager.jsx
```diff
@@ -212,12 +216,41 @@ export default class PanelManager extends React.Component {
...
-      </div>
-    </PanelContext.Provider>;
+    return <div>
+      <Suggest
+        inputNode={searchBarInputNode}
+        outputNode={searchBarOutputNode}
+        withCategories
+        onChange={query => {
+          if (ActivePanel === ServicePanel) {
+            this.setState({ isPanelVisible: query.length === 0 });
```
amatissart, line 229: The race conditions are still present. `<Suggest>` will suffer from unnecessary renders on successive input events while the active panel is ServicePanel.

amatissart:
I pushed some changes on the mechanism used to hide the current panel:
* `onChange` is now defined as a method of `PanelManager` to avoid unwanted `useEffect` reload in Suggest
* the `ActivePanel` is no longer unmounted/remounted, `display: none` is used instead. This was probably why some tests were failing: remounting the PoiPanel could trigger a new server request to load the panel content and conflict with an item selection from suggest results.
* css rules about `.panel-hidden` are deleted

bbecquet (approved):

PR #994, opened 2021-01-04
[IE11] Fix 3 JS errors that break nearly everything

bbecquet:
## Description
Fix 3 Javascript errors specific to IE11, that make the app completely broken on this browser and were easy to fix.

There are many other problems, mostly layout and behaviour of scrollable elements, that make the app barely usable at all. But they are trickier and probably time-consuming to fix.
I don't think if it's worth it and IMO we should quickly decide to drop the support of this browser, in favor of providing a better experience to users of modern browsers.


xem (approved):
nice catches!

amatissart (approved):

PR #995, opened 2021-01-04
Poi: floating buttons for direction + geolocate

G-Ray:
## Description
*What this PR does. If needed, explain choices you made, complex parts, architectural changes, etc.*

## Why

## Screenshots


G-Ray:
Let's put this on hold for now, as I don't see that much benefits compared to new maintenance cost of having to support 2 types of buttons: mapbox ones and react one

PR #997, opened 2021-01-05
Conversion of TopBar to React

bbecquet:
## Description
Introduce a new `<TopBar>` component:
 - replace the static HTML markup of the top bar
 - improve the `Suggest` I/O and state management, removing some hacks and sources of bugs
 - encapsulate layout manipulations that were done imperatively outside (focused, filled) 
 - remove completely the `SearchInput` global
 - simplify `Menu` rendering 

The `TopBar` itself was quite straightforward to move to React. Most of this PR content deals with the second point, the refacto of `Suggest`, which continues work done in https://github.com/Qwant/erdapfel/pull/1079.
The most important part is that the input field itself is now rendered by the `Suggest` component, but to let the parent control how it renders, I use a render function. This seems a bit complicated at first but is actually better than manipulating a DOM ref all the way. And it makes the life cycles easier to control deterministically.
I also chose to separate more clearly the three different values that the input field can take:
 - the user input
 - the value set when the app panel changes (ex: POI or category name)
 - the value of the suggestion which is currently highlighted with Up/Down keyboard navigation 

Previously, confusion between these three values made the different cases hard to follow and debug.

As the `DirectionInput` also uses `Suggest`, I had to adapt things in it too.

I tried to do things step-by-step, but could not separate this. And this seemed like the better option to do that when moving the input field in React, as the main reason we used a ref before was to ensure compatibility with the static input.

The downside is this makes a very big PR :disappointed: Sorry… *I propose we do a guided review if you think it's useful.*

BTW, I did not push some markup and CSS simplifications that could be also done, because they can wait.


src/components/TopBar.jsx
```diff
@@ -0,0 +1,148 @@
...
+      onToggleSuggestions={suggestionsOpened => { togglePanelVisibility(!suggestionsOpened); }}
+    />}
+  </>;
+};
+
+
+const TopBarWithRef =
+  React.forwardRef((props, ref) => <TopBar {...props} inputRef={ref} />);
+
+TopBarWithRef.displayName = 'DirectionInput';
```
amatissart, line 146: copy paste error, probably

src/components/TopBar.jsx
```diff
@@ -0,0 +1,183 @@
...
+    const query = inputRef.current.value;
+    const results = await fetchSuggests(query, {
+      withCategories: true,
+      useFocus: true,
+    });
+    onSelectSuggestion(results[0], {
+      query,
+      replaceUrl: true,
+    });
+  };
```
amatissart, line 73: There are some inconsistencies in the params passed to `onSelectSuggestion` and `selectItem`. 
According to line 56, the query should not be nested in a object, but the usage of `<Suggest onSubmit=...>` should be checked too.

As a consequence, an object may be used as the input value:
image

src/components/ui/Suggest.jsx
```diff
@@ -63,78 +66,80 @@ const Suggest = ({
...
-      const { value } = e.target;
+  useEffect(() => {
+    setHighlighted(null);
+    if (value !== null) {
       fetchItems(value);
       setIsOpen(true);
-      setLastQuery(value);
-    };
+    }
+  }, [fetchItems, value]);
```
amatissart, line 82: These 2 `useEffect` do not handle correctly the case when a empty input is focused. A list of favorites should appear as suggestions and it's no longer the case (probably related to the condition `inputNode.value === '' ` that has been removed).

bbecquet: I replaced these 2 effects by a single one, which makes things actually simpler. I hope I haven't overlooked yet another edge-case…

src/components/TopBar.jsx
```diff
@@ -0,0 +1,175 @@
...
+      // KeyboardEvent.key is either the printed character representation or a standard value for specials keys
+      // See https://developer.mozilla.org/fr/docs/Web/API/KeyboardEvent/key/Key_Values
+      if (
+        e.key.length === 1 &&
+        !e.ctrlKey &&
+        !e.metaKey &&
+        document.activeElement?.tagName !== 'INPUT'
+      ) {
+        setUserInputValue(inputRef.current.value + e.key);
+        inputRef.current.focus();
```
amatissart, line 34: It works but I am not sure to understand why calling `setUserInputValue()` is necessary.  
Isn't the input value supposed to change (caught by `onChange`) after the input has been focused ?

bbecquet: This is the point which is driving me mad…
The focus is done and yes this is supposed to work like you say, but the first `onChange` isn't triggered and the first letter is "lost". Honestly I've not figured exactly why yet, after a long debugging time spent on it…
I think it's related to the management of the user input value in `PanelManager`, as in one of the previous commits it was managed as local state in `TopBar` and it worked fine.

Feel free to spend time on it, maybe you'll understand what I wasn't able to, and I would be very interested in another PoV.

But if this is the last remaining hack I guess it's not too bad compared to the previous implementation.

xem (approved):
wow, that's a big, but very clean rewrite!
Works great according to my local tests.
as for the "hack", I agree that it's not too bad if it's just that.

amatissart (approved):

src/components/TopBar.jsx
amatissart: I am still not sure about what is going on here. Has this behavior ever worked with a "controlled" input? It looks like 
React manages a custom lifecycle for keystrokes in this case.

An alternative solution would be to re-dispatch the "keydown" event after the input has been focused. Then, the dependency to `setUserInputValue` could be removed:
```suggestion
        inputRef.current.focus();
        inputRef.current.dispatchEvent(new KeyboardEvent(e.type, e));
```
It may be more consistent with the initial implementation, but I am not sure it is any better.

PR #998, opened 2021-01-06
QMAPS fix dots ans line alignment in itinerary panel Desktop

xem:
## Description
fix dots and line alignment in itinerary panel Desktop
without breaking them on mobile of course

## Screenshots
image

### DIFF
--- a/src/scss/includes/direction-form.scss
+++ b/src/scss/includes/direction-form.scss
@@ -68,7 +68,7 @@
       background-repeat: no-repeat;
 
       @media (min-width: 641px) {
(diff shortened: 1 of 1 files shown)


amatissart:
The divider margin also needs to be adjusted. 
For some reason, on your screenshort it's too long, while on my machine it's too short:
image

xem:
Nice catch @amatissart , fixing it. It was too long on my screenshot because I zoomed my browser to make a bigger capture, otherwise it's too short here too :)

amatissart (approved):

PR #999, opened 2021-01-06
Poi: trigger + handle directionShortcutClick event

G-Ray:
## Description
- On the Direction shortcut button (managed by mapbox), listen for `set_direction_shortcut_callback`.
- Use set_direction_shortcut_callback on PoiPanel.

## Why
Add missing destination when clicking the "shortcut" direction button from a Poi.

## Screenshots
Peek 2021-01-06 14-54

### DIFF
--- a/src/mapbox/extended_nav_control.js
+++ b/src/mapbox/extended_nav_control.js
@@ -2,12 +2,19 @@ import ExtendedScaleControl from './extended_scale_control';
 import ExtendedAttributionControl from './extended_attribution_control';
 import GeolocControl from './extended_geolocate_control';
 import Telemetry from 'src/libs/telemetry';
+import { listen, unListen } from '../libs/customEvents';
 
 export default class ExtendedControl {
   constructor() {
     this._container = document.createElement('div');
     this.topButtonGroup = document.createElement('div');
     this.bottomButtonGroup = document.createElement('div');
+
+    // Store a callback to trigger when direction shortcut is clicked.
+    // if no callback is registered, then the default action will be executed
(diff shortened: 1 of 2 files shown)


xem (approved):
elegant

src/mapbox/extended_nav_control.js
```diff
@@ -38,7 +39,7 @@ export default class ExtendedControl {
       'direction',
       () => {
         Telemetry.add(Telemetry.MAP_ITINERARY);
-        window.app.navigateTo('/routes');
+        fire('geolocateControlClick');
```
amatissart, line 42: As a consequence, the button has no effect when it's used on the ServicePanel.

As @bbecquet had suggested, I think it would be cleaner to fire the event this other way around, to send a `poi` value here.

src/panel/poi/PoiPanel.jsx
```diff
@@ -44,6 +44,7 @@ export default class PoiPanel extends React.Component {
 
   componentDidMount() {
     fire('mobile_direction_button_visibility', false);
+    this.geolocateControlClickHandler = listen('geolocateControlClick', this.openDirection);
```
amatissart, line 47: What about 
```suggestion
    fire('set_current_poi', poi)
```
to notify other elements/buttons about the current poi ?

Note that this event would also be necessary in `componentDidUpdate`

amatissart (changes requested):

amatissart (approved):

bbecquet (approved):

PR #1000, opened 2021-01-06
QMAPS-1902 (poc) make steps scroll more rigid

xem:
## Description
Try to hijack css snap scroll to avoid scrolling many itinerary mobile's steps at once

## Screenshots
scroll

### DIFF
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -16,14 +16,25 @@ const MobileRoadMapPreview = ({
   const scroll = () => {
 
     const newStep = Math.floor(
-      // Divide the step container's scrollLeft up to the middle of the screen with the size of a step
-      // to determine which step is present at the middle of the screen
-      (stepsRef.current.scrollLeft + window.innerWidth / 2) / (window.innerWidth - 70 + 12)
+      // Determine which step is fully visible on screen
+      stepsRef.current.scrollLeft / (window.innerWidth - 70 + 12)
     );
 
-    // If it has changed, save it and highlight it as the current step
+    // If it is different from the current step:
     if (currentStep !== newStep) {
+
+      // Save it and highlight it as the current step
       setCurrentStep(newStep);
+
+      if ((currentStep - 1) !== newStep) {
+        // Stop scroll inertia
+        stepsRef.current.style.overflow = 'hidden';
+        stepsRef.current.scrollLeft = stepsRef.current.scrollLeft;
+        setTimeout(() => {
+          // Ref is not available in setTimeout, use querySelector instead
+          document.querySelector('.mobile-roadmap-preview-steps').style.overflow = 'auto';
+        }, 100);
+      }
     }
   };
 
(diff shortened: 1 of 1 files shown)


src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -16,14 +16,25 @@ const MobileRoadMapPreview = ({
...
 
-    // If it has changed, save it and highlight it as the current step
+    // If it is different from the current step:
     if (currentStep !== newStep) {
+
+      // Save it and highlight it as the current step
       setCurrentStep(newStep);
+
+      // Stop scroll inertia (except when going to previous step, which happens at the start of the scroll)
+      if ((currentStep - 1) !== newStep) {
```
bbecquet, line 30: Maybe this condition can be changed to work for both sides, as for now it adds inertia only in the 'next' direction.

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -16,14 +16,25 @@ const MobileRoadMapPreview = ({
...
+        stepsRef.current.style.overflow = 'hidden';
+        stepsRef.current.scrollLeft = stepsRef.current.scrollLeft;
```
bbecquet, line 32: This is really strange as it's just a self assign, but when I remove it I have a kind of glitch, like a jump at the end of the transition, so it's probably useful in a way but we have to understand it better.
What was the idea here?

bbecquet (commented):
At last I've tested it properly on iOS. This is very promising as it succeeds in enforcing the stops in one direction, but there are still some bugs.

Not noted in the PR comments, I also couldn't reach completely the last step: the last card wasn't highlighted, so there is probably also a problem with the `scrollLeft` value in the last step case.

Maybe we can extract the principle of your proposition, and find another implementation?
@xem if you had to summarize how this hack works in a few words, what would it be?

xem:
@bbecquet tbh I've been throwing random ideas at my browser until something finally have an effect. :)

> @xem if you had to summarize how this hack works in a few words, what would it be?

> it's just a self assign (...) What was the idea here?

- The idea is to let CSS overflow:auto + snap scroll continue to work natively (instead of doing everything using heavy JS)
- Every time a scroll event is detected, I do a quick computation to detect if, by any chance, a step card is fully visible on screen. If this card is different from the currently highlighted card, then the UI is updated to consider the corresponding step as the current one.
- The problem is that, as we're depending on native CSS scroll, snap, snap hardness and inertia, and since these behaviors are very dependent on browsers implementations, we can't be sure that the scroll has stopped on this step. If the inertia is too big or the snap too loose, the scroll can continue towards the next step(s), and that's what we want to avoid (or make harder) in this issue.
- So I wondered whet I could do to force this possible extra scroll from happening: first, I instantly make the container non-scrollable (via overflow:hidden), before re-making it scrollable a moment later (with a setTimeout). I saw that this technique wasn't enough to cancel the scroll inertia in all cases. In some cases. The browser may remember the inertia from when the container was scrollable, and continue scrolling even it after making it un-scrollable and re-scrollable (crazy huh?).
- I knew what value of scrollLeft I wanted to be set to the container, though. It's the value at the moment where the card is fully visible. It's the scrollLeft value at the exact moment we detect a new card must be highlighted. So, I tried the idea to make the parent un-scrollable, then assign a fixed scrollLeft value to it instead of letting it be automatic, then let the parent be scrollable again after a little pause. The self-assign seems stupid indeed, but I guess it's what the parent needs. A scrollLeft equal to itself instead of auto. On my mobile silulator, it seemed to kinda work, so here we are.

PR #1006, opened 2021-01-13
QMAPS-1919 update CSS for mobile step by step's arrow

xem:
## Description
update CSS for mobile step by step's arrow

## Screenshots
image

### DIFF
--- a/src/panel/direction/MobileRoadMapPreview.jsx
+++ b/src/panel/direction/MobileRoadMapPreview.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import RoadMapStep from './RoadMapStep';
 import { fire } from 'src/libs/customEvents';
 import classnames from 'classnames';
-import { Button } from '../../components/ui';
+import { FloatingButton } from '../../components/ui';
 
 const MobileRoadMapPreview = ({
   steps,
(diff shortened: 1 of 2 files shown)


src/scss/includes/direction.scss
```diff
@@ -335,14 +335,19 @@ button.direction_shortcut {
 
     .mobile-roadmap-preview-close {
       border: none;
-      padding: 0 0 0 $spacing-xs;
```
bbecquet, line 338: I think now we should be able to use our standard `FloatingButton` component for that, and remove all this specific CSS.

src/scss/includes/direction.scss
```diff
@@ -333,16 +333,8 @@ button.direction_shortcut {
...
-      margin: 0 0 0 11px;
-      background: $background;
-      width: 32px;
-      height: 32px;
-      border-radius: 50%;
-      box-shadow: $shadow;
-      color: $grey-black;
+    .floatingButton {
+      margin-left: 11px;
```
G-Ray, line 337: shouldn't we use a spacing value like `$spacing-s (12px)` ? Mockups are sometimes not exact I guess, in which case I think it's safe to round the value to our spacing value ?

src/panel/direction/MobileRoadMapPreview.jsx
```diff
@@ -51,8 +51,7 @@ const MobileRoadMapPreview = ({
   );
 
   return <div className="itinerary_mobile_step_by_step">
-    <Button
-      className="mobile-roadmap-preview-close"
+    <FloatingButton
```
G-Ray, line 54: :+1:

G-Ray (approved):

PR #1007, opened 2021-01-13
QMAPS-1914 change CSS in itinerary list

xem:
## Description
Change display of details and close buttons on mobile in the itinerary list.
And various fixes to match the design better.

## Screenshots
image

### DIFF
--- a/src/panel/direction/RouteSummary.jsx
+++ b/src/panel/direction/RouteSummary.jsx
@@ -37,6 +37,7 @@ export default class RouteSummary extends React.Component {
     return <Flex
       className="itinerary_leg_summary"
       justifyContent="space-between"
+      alignItems="flex-end"
       onClick={this.onClick}
     >
       <RouteSummaryInfo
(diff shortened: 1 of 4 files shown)


src/panel/direction/RouteSummary.jsx
```diff
@@ -37,6 +37,7 @@ export default class RouteSummary extends React.Component {
     return <Flex
       className="itinerary_leg_summary"
       justifyContent="space-between"
+      alignItems="flex-end"
```
bbecquet, line 40: This results in a proptype console warning, as this value isn't explicitely declared yet in our `<Flex>` component.

src/scss/includes/direction-panel.scss
```diff
@@ -54,13 +54,16 @@
...
-.direction-panel-mobile {
   &.maximized, &.panel--holding {
     .itemList > div:not(:first-child) {
       display: block;
+      margin-top: 12px;
     }
   }
+
+  .panel-close {
+    top: 15px;
```
bbecquet, line 66: This isn't taken into account as the selector is less specific than the existing one.

src/scss/includes/direction.scss
```diff
@@ -76,11 +78,12 @@
 }
 
 .itinerary_leg_summary {
-  padding: $spacing-l;
+  padding: 0 $spacing-s;
```
bbecquet, line 81: Why this change? As a result, the button is too close to the bottom of the screen and two consecutive summaries are too close when the panel is deployed.

Capture d’écran de 2021-01-15 10-13-10

Capture d’écran de 2021-01-15 10-13-18

G-Ray:
Looks good to me

G-Ray (approved):

PR #1008, opened 2021-01-14
Define a hover style for POI labels

bbecquet:
## Description
Change the label color of map tiles POI to blue on "hover".
To do so, modify the style of POI layers dynamically to add a condition based on Mapbox-GL [`feature-state`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#feature-state) to the existing `text-color` rule.
(Sadly some other parameters like the font weight can't be changed with this mechanism.)

For now I add that to existing init code about POI interactivity in `scene.js`, but I think we could structure this part better. Like extracting all this interactivity code to another module so it's easier to manage and simplify `scene.js`.

## Screenshots
Peek 14-01-2021 15-27


amatissart (approved):
Nice!

PR #1011, opened 2021-01-15
Better map fit for routes with labels

bbecquet:
## Description
A proposition to improve how the map view is changed to fit routes, by taking overlay labels into accounts, so they are not (or at least less) cut by the screen borders.

It seemed like an easy problem at first, solvable by always enforcing big paddings, but this had drawbacks:
 - paddings on all sides reduced a lot the amount of map space usable, which is critical on mobile devices.
 - we could use the direction of the labels to add paddings only on some sides, but nothing ensured the labels exceeded the route bbox in the first place. Moreover, our padding code is already horrible with a lot of intertwined functions, and this would have made it even more complex.

As an alternative, I proposed to enlarge the real routes bbox by shifting the label anchor points (real lng/lat) by some amount in the direction they are displayed.
That way, we only expand the bbox in the useful directions, and "inner" labels have less chance to uselessly expand the bbox. The amount by which the points are shifted is expressed in lng/lat degrees as an arbitrary ratio of the route bbox, so it's not dependant on the route length.

I made a small drawing to try to explain it better :smile: 

IMG_0014

The black rectangle is the route bbox. A,B,C are the anchor points of the labels, which are shifted in there respective directions, contributing to enlarge the original bbox to the green one. The B point will have no effect as it is internal enough.

## Screenshots
|Before|After|
|---|---|
|maps dev qwant ninja_maps_routes__mode=driving destination=latlon%3A48 94730%3A2 72787 pt=true origin=latlon%3A48 74473%3A2 38566 selected=0(iPhone 6_7_8)|localhost_3000_routes__mode=driving destination=latlon%3A48 91025%3A2 04315 pt=true origin=latlon%3A48 82937%3A2 07420 selected=0(iPhone 6_7_8)|
|maps dev qwant ninja_maps_routes__mode=driving destination=latlon%3A48 94730%3A2 72787 pt=true origin=latlon%3A48 74473%3A2 38566 selected=0(iPhone 6_7_8) (2)|localhost_3000_routes__mode=driving destination=latlon%3A48 73405%3A2 34890 pt=true origin=latlon%3A48 71043%3A2 50723 selected=0(iPhone 6_7_8)|

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -71,8 +71,8 @@ export default class SceneDirection {
     this.map = map;
     this.routes = [];
     this.routeMarkers = [];
-    this.labelPositions = [];
     this.routeLabels = [];
+    this.fullBbox = null;
     this.mapFeaturesByRoute = {};
 
     const iconsBaseUrl = nconf.get().system.baseUrl + 'statics/images/direction_icons';
@@ -184,8 +184,8 @@ export default class SceneDirection {
     });
     this.updateMarkers(mainRoute);
     this.updateRouteLabels(mainRoute);
-    if (fitView) {
-      fire('fit_map', this.getAllRoutesBBox());
+    if (fitView && this.fullBbox) {
+      fire('fit_map', this.fullBbox);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/adapters/scene_direction.js
```diff
@@ -38,11 +39,39 @@ const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
...
+    lngShift: (routesBbox.getEast() - routesBbox.getWest()) / (smallScreen ? 3 : 10),
+  };
+  const labelsBbbox = new LngLatBounds();
+  labelPositions
+    .map(shiftLabelPosition(shift))
+    .forEach(labelPosition => { labelsBbbox.extend(labelPosition); });
+  return labelsBbbox;
+};
+
+const shiftLabelPosition = ({ lngShift, latShift }) => ({ lngLat, anchor }) => {
```
G-Ray, line 55: :+1:

src/adapters/scene_direction.js
```diff
@@ -189,9 +218,9 @@ export default class SceneDirection {
   }
 
   displayLabels(routes, vehicle) {
-    const labelPositions = getLabelPositions(routes.map(route => route.geometry));
+    this.labelPositions = getLabelPositions(routes.map(route => route.geometry));
```
G-Ray, line 221: Is there any unwanted effect if we use `getLabelsBbbox` here to compute `labelPositions` ?

G-Ray:
Great improvement and implementation !
I'm just asking if we could compute this shifts directly in `displayRoutes` ?

src/adapters/scene_direction.js
bbecquet: It would be possible yes, but I think it would make the code more complex, having to compute the routes bbox here also or in advance and store it.
`fit_map` can happen outside of `displayRoutes`, so I took the same approach as before to store the mininum state and compute things on the fly. But I agree this file is quite a mess and having to store `labelPositions` is unsatisfying.

G-Ray: ok, in this case, could we avoid `this.labelPositions`, by calling `getLabelPositions` where needed instead (in `getAllRoutesBBox` ?

G-Ray: If it's quite straightforward go fot it :)

bbecquet: Nothing is straightforward here, but I did it :)

G-Ray (approved):

PR #1012, opened 2021-01-15
[React] Convert some easy components to functional 

bbecquet:
## Description
Convert these React class components to the functional syntax:
 - `<Website>`
 - `<OpeningHour>`
 - `<Menu>`
 - `<ServicePanelMobile>`
 - `<FavoritePoi>`
 
## Why
The functional component syntax is the one promoted by React, used by the Search team, and allows easier state management and composition of code with hooks.


xem (approved):
very good cleanup

PR #1014, opened 2021-01-19
Direction: on mobile, clear geoloc input on tap

G-Ray:
## Description
On mobile, clear the input field on tap if the field is of type geoloc, thus avoiding unwanted queries to suggestions.

## Why
Better UX

(I can't provide any GIF this time, as it only capture a blank screen.)

### DIFF
--- a/src/panel/direction/DirectionForm.jsx
+++ b/src/panel/direction/DirectionForm.jsx
@@ -76,6 +76,7 @@ export default class DirectionForm extends React.Component {
           <DirectionInput
             isLoading={isLoading}
             value={originInputText}
+            point={origin}
             pointType="origin"
             onChangePoint={(input, point) => this.onChangePoint('origin', input, point)}
             ref={this.originRef}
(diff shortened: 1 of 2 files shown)


src/panel/direction/DirectionForm.jsx
```diff
@@ -85,6 +86,7 @@ export default class DirectionForm extends React.Component {
           <DirectionInput
             isLoading={isLoading}
             value={destinationInputText}
+            point={destination}
```
G-Ray, line 89: I think we should pass point as props from now, and use its properties inside `DirectionInput`. wdyt ?

bbecquet: If I remember correctly, we passed flat properties and not full point objects because we needed to separate the current input string itself from the point data that may be attached to the field.
There is probably a best way to do that now, but let's not change everything in this PR, as it's one of the biggest source of bugs in the app :)

I have high hopes (maybe too high?) we can improve the whole `DirectionPanel`/`DirectionForm`/`DirectionInput` group when:
1. `<Suggest>` has been simplified further (with requires the top bar to be implemented in React)
2. we move these components to the functional form, which often highlights better ways to manage complex states.

bbecquet (approved):

PR #1016, opened 2021-01-25
Use icon mapping function from qwant-basic-gl-style

bbecquet:
## Description
Uses the POI-to-icon mapping function exposed by `@qwant/qwant-basic-gl-style` in https://github.com/Qwant/qwant-basic-gl-style/pull/111, instead of defining it in Erdapfel. The only exception is for the Geolocation case, which is technically managed as a POI in Erdapfel, but it wouldn't make sense in the shared library.

## Why
Share the icon mapping with Search.

### DIFF
--- a/src/adapters/icon_manager.js
+++ b/src/adapters/icon_manager.js
@@ -1,96 +1,27 @@
-import styleIcons from '@qwant/qwant-basic-gl-style/icons.yml';
+import { getPoiIcon } from '@qwant/qwant-basic-gl-style';
 import classnames from 'classnames';
 import { ACTION_BLUE_DARK } from 'src/libs/colors';
 
-const nameToClass = iconName => iconName.match(/^(.*?)-[0-9]{1,2}$/)[1];
-
 export default class IconManager {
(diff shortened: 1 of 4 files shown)


src/adapters/icon_manager.js
```diff
@@ -1,96 +1,27 @@
...
-      };
     }
-  }
-}
 
-export function createIcon(iconOptions) {
-  const icon = IconManager.get(iconOptions);
-
-  // Show a white circle instead of marker2 in map markers of PoI that have no class or subclass.
-  if (icon.iconClass === nameToClass(styleIcons.defaultIcon)) {
```
bbecquet, line 82: This function depended on the YML mapping, but during the simplification I realized we never use it for something other than creating this static marker with no special options.
Capture d’écran de 2021-01-25 11-10-39

G-Ray (dismissed):

G-Ray (commented):
Nice cleanup :clap:

G-Ray (approved):

PR #1019, opened 2021-01-26
Burger menu redesign

bbecquet:
## Description
Redesign the burger menu.
 - remove most of the menu content, which is considered confusing or outdated
 - new squared design of the button on map and new icons (burger, close circle and items)
 - new partly transparent overlay, instead of a nearly opaque one

For now, the mobile version won't match the final design. It would require a complex drawer component that we don't want to clone in Erdapfel before we can use the real Design System components. The compromise is to use a sliding panel, like on desktop and like the current prod version on mobile.

## Why
Modernization + simplification + matching the next Search design

## Screenshots
|Before|After|
|---|---|
|www qwant com_maps_|localhost_3000_favs_ (1)|

Peek 26-01-2021 15-48


src/panel/Menu.jsx
```diff
@@ -36,58 +37,47 @@ const Menu = () => {
...
-              <i className="menu__panel__action__icon icon-zap" />
-              <span>{_('How to contribute', 'menu')}</span>
-            </a>
-          </div>
-
-          {menuItems.map(menuItem => <MenuItem key={menuItem.sectionName} menuItem={menuItem} />)}
+            }}
+            icon={<Heart width={16} color={PINK_DARK} />}
+          >
+            {_('Favorites', 'menu')}
```
amatissart, line 65: The mockup specified "My favorites", which is consistent with the title used on the FavoritesPanel header.

bbecquet: Fixed.

src/components/ui/icons.js
```diff
@@ -1,2 +1,8 @@
 export { default as Heart } from '../../../public/images/feather/heart.svg';
 export { default as Magnifier } from '../../../public/images/magnifier-dark.svg';
+
+export { default as IconLightbulb } from '../../../public/images/remix/lightbulb-line.svg';
```
G-Ray, line 4: :+1:

src/scss/includes/menu.scss
```diff
@@ -160,20 +73,32 @@
...
-
-    &__line {
-      margin: 2px auto;
+    right: $spacing-s;
+    top: $spacing-xl;
+    width: 40px;
+    height: 40px;
+    border-radius: 8px;
+    box-shadow: $shadow;
+    color: $grey-semi-darkness;
```
G-Ray, line 82: `grey-semi-darkness` should be set on mobile too :lipstick:

src/panel/Menu.jsx
```diff
@@ -36,58 +37,47 @@ const Menu = () => {
...
-            >
-              <img
-                className="menu__panel__action__icon"
-                src="./statics/images/magnifier-dark.svg" alt=""
-              />
-              <span>{_('Search', 'menu')}</span>
-            </button>
-            {isDirectionActive &&
-              <button className="menu__panel__action" onClick={() => {
-                Telemetry.add(Telemetry.MENU_ITINERARY);
```
G-Ray, line 66: if we get rid of this specific event, we can remove the `menu_itinerary` event from the `telemetry` module

src/panel/Menu.jsx
```diff
@@ -21,9 +20,11 @@ const Menu = () => {
     };
   }, []);
 
-  const open = () => {
-    Telemetry.add(Telemetry.MENU_CLICK);
-    setIsOpen(true);
+  const toggle = () => {
+    if (!isOpen) {
+      Telemetry.add(Telemetry.MENU_CLICK);
```
G-Ray, line 25: Ok, so we can expect 2 times less registered events from now IIUC

src/panel/menu/MenuItem.jsx
```diff
@@ -1,27 +1,23 @@
...
-    {links.map(link => <a
-      key={link.uri}
-      className="menu__panel__link"
-      href={link.uri} rel="noopener noreferrer" target="_blank"
-    >
-      {_(link.name)}
-    </a>)}
-  </div>}
-</div>;
+const MenuItem = ({ icon, children, href, onClick, outsideLink }) =>
```
G-Ray, line 6: :+1:

src/panel/Menu.jsx
bbecquet: Yep, and nobody cares :)

G-Ray (approved):

PR #1027, opened 2021-02-19
Use OSM view/edit URLs provided by the API

bbecquet:
## Description
Follow-up to https://github.com/Qwant/idunn/pull/201
Now that Idunn provides OpenStreetMap.org view/edit URLs directly in the POI payload, use them instead of building them in the front by string manipulation.

## Why
Simplification + single source of truth


amatissart (commented):
It looks good to me, but this change depends on the latest version Idunn being deployed to production before merging.

amatissart (approved):
[Idunn v0.2.4](https://github.com/Qwant/idunn/releases/tag/v0.2.4) has now been released and deployed to production.

PR #1028, opened 2021-02-19
style: Fallback to "en" on unsupported "lang", and implement "layers" filter

amatissart:
Extends the existing "/style.json" endpoint, with validation of its optional parameters:

* `lang` now fallbacks to English for unsupported languages.
* `layers` is a new parameter that accept "**all**" or "**nopoi**".  
"all" layers are returned by default. "nopoi" excludes non-transport POIs.

Tests are defined in "tests/integration/tests/server.js" to validate this behavior.

### DIFF
--- a/bin/middlewares/map_style.js
+++ b/bin/middlewares/map_style.js
@@ -5,9 +5,15 @@ const styleConfigure = require('@qwant/mapbox_style_configure');
 const qwantStyle = require('@qwant/qwant-basic-gl-style/style.json');
 const path = require('path');
 const Uri = require('@qwant/uri');
+const { query, validationResult } = require('express-validator');
 
 
-module.exports = function(config, languages) {
+const LayersEnum = Object.freeze({
+  ALL: 'all',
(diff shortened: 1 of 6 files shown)


bin/middlewares/map_style.js
```diff
@@ -30,25 +36,16 @@ module.exports = function(config, languages) {
...
-        res.status(400);
-        res.send({ error: `Unsupported language code : ${langParameter}` });
-        return;
-      }
-    } else {
-      matchedLanguage = languages.defaultLanguage;
+  const styleCache = {};
+
+  const getStyle = function({ host, lang, nopoi }) {
+    const cacheKey = `${host};${lang};${nopoi}`;
```
bbecquet, line 42: I wondered if we had an already imported memoization util like for example [lodash.memoize](https://www.npmjs.com/package/lodash.memoize), to have a pure fonction and avoid writing the cache logic ourselves.
But apparently not and it's simple enough, so :+1:

bbecquet (approved):
Great!

PR #1029, opened 2021-02-23
Update pj logo

G-Ray:
image
:arrow_down: 
image


bbecquet:
We should remove this added left margin as visually the space is too large with the new logo https://github.com/Qwant/erdapfel/blob/master/src/scss/includes/panels/poi_panel.scss#L52-L54

bbecquet (approved):

PR #1030, opened 2021-02-26
Create a fallback page for unsupported browsers (esp. IE11) 

bbecquet:
## Description
Build the new page that will be used as a fallback for users with unsupported browsers, primiarily IE11 (but the design and the code is not limited to it).
In this PR, only build the page as static markup and CSS and declare a new route (`/unsupported`) to check the display this page. The real conditions to trigger this page and the actual form it takes will be addressed in a separate PR.

The only tricky part here is that I modified a bit the webpack config to reuse the existing CSS workflow to build a separate `unsupported.css` file.
Due to limitations in Webpack 4 about entry names, I had to change the name of the main app Sass file to `app.scss` instead of `main.scss`, so that the output is still `app.css` with the new config. IMO it makes even more sense like that than previously.

## Screenshots
Screenshot on a real IE11 on an old Win7 PC:
qwant_maps_unsupported


src/scss/unsupported.scss
```diff
@@ -0,0 +1,87 @@
+
```
G-Ray, line 1: Neat

G-Ray (approved):
:ok_hand:

views/unsupported.ejs
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="<%= language.code %>">
+  <head>
+    <meta charset="utf-8">
+    <title><%= _('Qwant Maps') %> - <%= _('Unsupported browser', 'unsupported') %></title>
+    <meta name="description" content="<%= _('The map that respects your privacy') %>">
+    <meta name='viewport' content="width=device-width,initial-scale=1"/>
+    <link href='./statics/css/unsupported.css' rel='stylesheet'/>
+    <link href='./mapstyle/iconfont/iconfont.css' rel='stylesheet'/>
```
amatissart, line 9: Is this `iconfont.css` really required ?

bbecquet: No indeed, thanks for spotting it :+1:

G-Ray (approved):

PR #1031, opened 2021-02-26
Back to Qwant button

G-Ray:
## Description
Display a "back to qwant" button on poi and category panels, depending on the presence of a `isFromQwant` query param.

Query param `client` can trigger this button withvalue 'search-ia-multi'

## Why
UX between phoenix and qwant

## Screenshots
desktop POI panel
mobile POI panel
desktop POI category
desktop POI category

I still need to figure out why the divider is missing in the category panel, in desktop mode

### DIFF
--- a/src/components/BackToQwantButton.jsx
+++ b/src/components/BackToQwantButton.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+import { Button } from 'src/components/ui';
+
+export const BackToQwantButton = ({ ...props }) => {
+
+  return (
+    <Button
+      key="back-to-qwant"
+      icon="arrow-left"
+      variant="tertiary"
+      onClick={() => window.history.back()}
+      style={{
+        backgroundColor: 'white',
+        boxShadow: '0 1px 4px 0 rgba(12, 12, 14, 0.2), 0 0 2px 0 rgba(12, 12, 14, 0.12)',
+      }}
+      {...props}
+    >
+      Retour sur Qwant
+    </Button>
+  );
+};
--- a/src/components/BackToQwantButton.jsx
+++ b/src/components/BackToQwantButton.jsx
@@ -1,9 +1,29 @@
 /* globals _ */
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import { Button } from 'src/components/ui';
 
-export const BackToQwantButton = ({ ...props }) => {
+export const BackToQwantButton = ({ isMobile }) => {
+
+  useEffect(() => {
+    let mapScaleElement;
+
+    window.execOnMapLoaded(() => {
+      if (!isMobile) {return;}
+      // Hide scale while the button is mounted as it would overlap
+      const elems = document.getElementsByClassName('map_control__scale_attribute_container');
+      if (elems.length > 0) {
+        mapScaleElement = elems[0];
+        mapScaleElement.style.visibility = 'hidden';
+      }
+    });
+
+    return () => {
+      if (mapScaleElement) {
+        mapScaleElement.style.visibility = 'visible';
+      }
+    };
+  }, [isMobile]);
 
   return (
     <Button
@@ -11,7 +31,6 @@ export const BackToQwantButton = ({ ...props }) => {
       icon="arrow-left"
       variant="tertiary"
       onClick={() => window.history.back()}
-      {...props}
     >
       {_('Back to Qwant.com')}
     </Button>
--- a/src/components/BackToQwantButton.jsx
+++ b/src/components/BackToQwantButton.jsx
@@ -3,26 +3,16 @@ import React, { useEffect } from 'react';
 
 import { Button } from 'src/components/ui';
 
+const hiddenAttributeClassName = 'map_control__scale_attribute_container--hidden';
+
 export const BackToQwantButton = ({ isMobile }) => {
 
   useEffect(() => {
-    let mapScaleElement;
+    if (!isMobile) {return;}
 
-    window.execOnMapLoaded(() => {
-      if (!isMobile) {return;}
-      // Hide scale while the button is mounted as it would overlap
-      const elems = document.getElementsByClassName('map_control__scale_attribute_container');
-      if (elems.length > 0) {
-        mapScaleElement = elems[0];
-        mapScaleElement.style.visibility = 'hidden';
-      }
-    });
-
-    return () => {
-      if (mapScaleElement) {
-        mapScaleElement.style.visibility = 'visible';
-      }
-    };
+    // Hide scale while the button is mounted as it would overlap
+    document.body.classList.add(hiddenAttributeClassName);
+    return () => document.body.classList.remove(hiddenAttributeClassName);
   }, [isMobile]);
 
   return (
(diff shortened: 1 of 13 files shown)


src/libs/url_utils.js
```diff
@@ -72,3 +72,9 @@ export function updateQueryString(queriesObject) {
     ...queriesObject,
   });
 }
+
+export function isFromQwant() {
+  const params = parseQueryString(window.location.search);
+  const qwantClients = ['search-ia-local', 'search-ia-multi', 'search-ia-address'];
```
amatissart, line 78: Unless I am mistaken, the "back to Qwant" button should be visible for "IA Multi" only. So I would suggest to rename this function to something like `shouldShowBackToQwant()` ?

Moreover, `search-ia-maps-multi` would be more explicit in my opinion. And `search-ia-local` will be obsolete soon, so no reason to mention it here.

src/libs/url_utils.js
```diff
@@ -72,3 +72,9 @@ export function updateQueryString(queriesObject) {
...
+  return params.client && qwantClients.includes(params.client);
```
amatissart, line 79: Beware of using `.includes()`, it would probably break the app on IE11.

src/libs/url_utils.js
G-Ray: The button should be displayed on a poi panel too from the ticket's description. I'm ok with the renaming though :)

src/components/ui/FloatingItems.jsx
```diff
@@ -1,7 +1,11 @@
 import React from 'react';
 
-const FloatingItems = ({ items }) =>
-  <div className="floatingItems">
+const FloatingItems = ({ items, position }) =>
+  <div
+    className="floatingItems"
+    style={position === 'left'
```
bbecquet, line 6: Why not manage this 12 as padding, as this seems to be what it is I think.
And I guess the transform is about putting things above the scale and i button? I guess it would be better visually to hide them (or lift them above the items) when `position === 'left'`.

src/components/ui/Panel.jsx
```diff
@@ -311,7 +314,10 @@ class Panel extends React.Component {
             }}
             {...(isMobile && resizable && this.getEventHandlers())}
           >
-            {floatingItems && <FloatingItems items={floatingItems} />}
+            {floatingItemsLeft && size !== 'maximized' &&
```
bbecquet, line 317: Why do we need the `size !== 'maximized'` test when it wasn't used for the right items?

src/panel/category/CategoryPanel.jsx
```diff
@@ -134,10 +137,28 @@ const CategoryPanel = ({ poiFilters = {}, bbox }) => {
     </>;
   }
 
+  const NavHeader = () => {
+    if (isMobile) {return null;}
+
+    if (shouldShowBackToQwant()) {
+      return (
+        <PanelNav>
+          <BackToQwantButton style={{}} key="back-to-qwant" />
```
bbecquet, line 146: You don't need these props.

src/panel/category/CategoryPanel.jsx
```diff
@@ -134,10 +137,28 @@ const CategoryPanel = ({ poiFilters = {}, bbox }) => {
     </>;
   }
 
+  const NavHeader = () => {
+    if (isMobile) {return null;}
```
bbecquet, line 141: I wonder if it's worth creating another component for that…
At least you can simplify it by moving the last `return null` into this early return test.

src/panel/poi/PoiPanel.jsx
```diff
@@ -221,6 +222,34 @@ export default class PoiPanel extends React.Component {
         ? this.backToFavorite
         : this.closeAction;
 
+    const NavHeader = ({ isMobile }) => {
+      if (isMobile) {return null;}
+
+      if (shouldShowBackToQwant()) {
+        return (
+          <PanelNav>
+            <BackToQwantButton style={{}} key="back-to-qwant" />
```
bbecquet, line 231: Once again, why the `key`? :)

src/components/BackToQwantButton.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+  return (
+    <Button
+      className="backToQwantButton"
+      icon="arrow-left"
+      variant="tertiary"
+      onClick={() => window.history.back()}
+      {...props}
+    >
+      Retour sur Qwant
```
amatissart, line 15: To translate, and according to mockups, the wording should be something like "Back to Qwant.com".

src/components/ui/Panel.jsx
G-Ray: image

I added this check in order to avoid this situation. This should not happen currently as we have no button visible in maximized state.

Maybe we should add this check for right floatingItems ?

src/panel/poi/PoiPanel.jsx
```diff
@@ -232,13 +261,10 @@ export default class PoiPanel extends React.Component {
...
-          renderHeader={!isMobile && backAction !== this.closeAction &&
-            <PanelNav
-              isMobile={isMobile}
-              onGoBack={backAction}
-              goBackText={_('Display all results')}
-            />
-          }
+          renderHeader={<NavHeader isMobile={isMobile} />}
+          floatingItemsLeft={isMobile && [
+            <BackToQwantButton key="back-to-qwant" />,
```
bbecquet, line 266: Same here, we should check `shouldShowBackToQwant`.
So at the end it will be in four different places, this will be hard to maintain and keep cohesive. Maybe we should try to find a way to factorize that and make it easy to apply to other Panels also? WDYT?

src/panel/category/CategoryPanel.jsx
```diff
@@ -134,10 +137,28 @@ const CategoryPanel = ({ poiFilters = {}, bbox }) => {
...
+    return null;
+  };
+
   return <Panel
     resizable
+    renderHeader={<NavHeader isMobile={isMobile} />}
     minimizedTitle={_('Unfold to show the results', 'categories')}
     className={classnames('category__panel', { 'panel--pj': dataSource === sources.pagesjaunes })}
+    floatingItemsLeft={[
+      isMobile && <BackToQwantButton key="back-to-qwant" />,
```
bbecquet, line 160: We have to use `shouldShowBackToQwant` here also, otherwise it's always present on mobile.

src/components/ui/Panel.jsx
bbecquet: > Maybe we should add this check for right floatingItems ?

Yes probably :+1:

src/panel/poi/PoiPanel.jsx
bbecquet: Beware, you used `shouldShowBackToQwant` as a boolean instead of a function.

src/components/ui/FloatingItems.jsx
G-Ray: Thanks, I added a commit to hide the scale element on the map while the button is displayed, on mobile.

bbecquet (approved):

PR #1033, opened 2021-03-01
Trigger the unsupported browser page when user uses IE11

bbecquet:
## Description
Follow-up to https://github.com/Qwant/erdapfel/pull/1030.
Now that we have set up a page for unsupported browsers, let's redirect to it when we detect it's an old IE.
I used an express middleware for that, so it's easy to make it grow and maintain without complexifying the default app route implementation.

### DIFF
--- a/bin/middlewares/unsupported_browser.js
+++ b/bin/middlewares/unsupported_browser.js
@@ -9,7 +9,7 @@ module.exports = function(config) {
 
   return function(req, res, next) {
     if (redirect && isUnsupported(req.headers['user-agent'])) {
-      res.redirect(302, req.baseUrl + '/unsupported');
+      res.redirect(302, config.system.baseUrl + 'unsupported');
     } else {
       next();
     }
--- a/bin/middlewares/unsupported_browser.js
+++ b/bin/middlewares/unsupported_browser.js
@@ -4,10 +4,14 @@ function isUnsupported(userAgent = '') {
   return userAgent.match('Trident'); // old IE detection
 }
 
-module.exports = function(req, res, next) {
-  if (isUnsupported(req.headers['user-agent'])) {
-    res.redirect(302, '/unsupported');
-  } else {
-    next();
-  }
+module.exports = function(config) {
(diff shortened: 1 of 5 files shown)


G-Ray (dismissed):

bin/middlewares/unsupported_browser.js
```diff
@@ -0,0 +1,13 @@
+/* globals module */
+
+function isUnsupported(userAgent = '') {
+  return userAgent.match('Trident'); // old IE detection
+}
+
+module.exports = function(req, res, next) {
+  if (isUnsupported(req.headers['user-agent'])) {
+    res.redirect(302, '/unsupported');
```
amatissart, line 9: I think that we should define a feature flag on this redirection, to be able to merge this PR independently from the moment when the end of support is decided.

bbecquet: Yes I should have done that in the first place… And it will be useful to the ensure translations of the page are complete.

bbecquet:
@amatissart I added a feature flag and a test case for the redirection.

bin/middlewares/unsupported_browser.js
```diff
@@ -0,0 +1,17 @@
...
+function isUnsupported(userAgent = '') {
+  return userAgent.match('Trident'); // old IE detection
+}
+
+module.exports = function(config) {
+  const redirect = config.server.unsupportedBrowsers.redirect;
+
+  return function(req, res, next) {
+    if (redirect && isUnsupported(req.headers['user-agent'])) {
+      res.redirect(302, req.baseUrl + '/unsupported');
```
amatissart, line 12: `req.baseUrl` is the base URL used by the parent router. In practice on deployed environments, Erdapfel is behind a proxy that is responsible for the routing of URLs among all maps services. So `req.baseUrl` would be empty in this case, even if the user needs to be redirected to `/maps/unsupported`.

`config.system.baseUrl` should probably be used instead of `req.baseUrl`:
```suggestion
      res.redirect(302, config.system.baseUrl + 'unsupported');
```
```

amatissart (approved):

PR #1034, opened 2021-03-02
Use npm 7 and define local_modules as workspaces

amatissart:
Upgrade to [npm 7](https://github.blog/2021-02-02-npm-7-is-now-generally-available/), and use [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to maintain packages defined in local_modules.

This new version of npm comes with some breaking changes about how "scripts" are executed, and defines a new package-lock.json format. So npm 7 should be used on local environments used by developers too.


package.json
```diff
@@ -16,7 +16,7 @@
     "i18n-pull": "tx pull --all --force",
     "watch": "webpack --config build/webpack.config.js -w --mode=development",
     "msgmerge": "node local_modules/merge-po/index",
-    "prepare": "node build/before_build.js",
+    "prepare": "if [ \"$NODE_ENV\" != production ]; then node build/before_build.js; fi",
```
bbecquet, line 19: Not sure to understand why it isn't useful in prod. Can you explain please?

amatissart: This "before_build" script generates and copies static files related to map style fonts and icons to the "public" directory. So it needs to be run only once, at build time.

With npm 6, "prepare" was triggered automatically after `npm install` or `npm ci`, but not after `npm install --production`. This fits with how the docker image is built: the production install is used to install the non-dev dependencies required by the server, after all build operations have been executed in a previous step.

It's unclear why, but this behavior changed with npm 7. "prepare" is executed after all `npm install` or `npm ci`, even if `--production` is passed.

I agree that it should be made more explicit though. As this change is specific to how the docker image is built, maybe we could use a custom env var `IGNORE_PREPARE`, instead of exploiting `NODE_ENV` ?

bbecquet: Thanks for the in-depth explanation :+1:

I agree that it would be clearer with a dedicated env var. We can even add a comment for it in the Dockerfile if it is not clear enough.

I guess another solution would be to stop using `prepare` for that, and use a custom script for it that we call explicitely. This change in v7 is really strange.

bbecquet (approved):

Dockerfile
```diff
@@ -22,7 +22,7 @@ COPY --chown=node local_modules $PROJECT_DIR/local_modules
 
 # Install with dev and build dependencies
 # 'npm prepare' is called after install
-RUN npm install
+RUN npm ci
```
G-Ray, line 25: :+1:

G-Ray (approved):

PR #1035, opened 2021-03-02
Transpile @qwant/qwant-basic-gl-style

bbecquet:
## Description
Fix IE11 JS loading by transpiling the `@qwant/qwant-basic-gl-style` dependency.

### DIFF
--- a/build/babel.config.js
+++ b/build/babel.config.js
@@ -38,9 +38,9 @@ module.exports = function(mode) {
   const conf = mode === 'production' ? production : development;
 
   return {
-    'presets': conf.presets.concat([ '@babel/preset-react' ]),
-    'ignore': [
-      'node_modules',
+    presets: conf.presets.concat([ '@babel/preset-react' ]),
+    ignore: [
+      /node_modules\/(?!@qwant\/qwant-basic-gl-style)/,
     ],
     plugins,
   };
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -123,9 +123,6 @@ const mainJsChunkConfig = buildMode => {
             options: babelConf(buildMode),
           },
         ],
-        exclude: [
-          /\/node_modules/,
-        ],
       }, {
         test: /\.svg$/,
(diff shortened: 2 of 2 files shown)


build/webpack.config.js
```diff
@@ -123,9 +123,6 @@ const mainJsChunkConfig = buildMode => {
             options: babelConf(buildMode),
           },
         ],
-        exclude: [
```
G-Ray, line 126: Shouldn't we let this one as this ?

build/babel.config.js
```diff
@@ -38,9 +38,9 @@ module.exports = function(mode) {
   const conf = mode === 'production' ? production : development;
 
   return {
-    'presets': conf.presets.concat([ '@babel/preset-react' ]),
-    'ignore': [
-      'node_modules',
+    presets: conf.presets.concat([ '@babel/preset-react' ]),
+    ignore: [
+      /node_modules\/(?!@qwant\/qwant-basic-gl-style)/,
```
amatissart, line 43: Do you think it would be a good idea to babelify all node module (except maybe mapbox-gl that provides a minified version already) ?

amatissart (approved):

build/webpack.config.js
bbecquet: No, actually it's duplicated with the `ignore` rule in `babel.config.js`. If we let it, we have to pass the exception here also.

build/babel.config.js
bbecquet: Yeah, why not.
The price to pay is slower build times and a warning message from babel on `react-dom` because it is too big to be properly optimized. But we can exclude that one too.
Do you propose I do it here?

amatissart: Well, if multiple exceptions are also required the other way around, I am ok to merge this PR as is.  
And this problem should be less critical once IE11 becomes unsupported.

bbecquet: Ok, I'm merging this then.

PR #1036, opened 2021-03-02
Move POIs subclasses names to a new "qwant-maps-common" package

amatissart:
* `@qwant/qwant-maps-common` is a separate package, defined as a workspace (see #1034)
  As a result, this package is built automatically on the initial `npm install`  
  It may also be published directly to NPM to be used by other projects
   
* This new packages exposes a single `getPlaceCategoryName` function.

* Related translations are moved a new resource "qwant-maps.common" in the same project "qwant-maps" on Transifex. 
  The workflow used to upload and update translations is unchanged.


local_modules/qwant-maps-common/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+Output files are generated in `./dist/`.
+
+
+### Publish a new version
+
+After you applied your modifications, and translations are up-to-date, you may publish a new release.
+
+* Update the version number in "package.json"
```
bbecquet, line 43: What about proposing to run `npm version` for that? https://docs.npmjs.com/cli/v7/commands/npm-version

bbecquet (approved):

local_modules/qwant-maps-common/README.md
amatissart: Thanks for the tip. I updated the README.

bbecquet (approved):

local_modules/qwant-maps-common/src/place_category_name.js
```diff
@@ -0,0 +1,257 @@
...
+  yoga: () => _('yoga'),
+  zoo: () => _('zoo'),
+
+  city: () => _('city'),
+  country: () => _('country'),
+  address: () => _('address'),
+  street: () => _('street'),
+};
+
+function getPlaceCategoryName({ subclass }, lang) {
```
G-Ray, line 249: Not a big deal, but we could either sur params (subclass and lang), or an object as param

bbecquet: The idea behind this separation is that the first parameter represents a place object. For now we only use the `subclass` attribute on it to determine the category string, but it may change in the future with other properties used.

G-Ray: ok it makes sense :)

G-Ray (approved):

G-Ray (approved):

PR #1037, opened 2021-03-05
[React] Convert ShareMenu to a functional component

bbecquet:
## Description
Continuing converting components one-by-one…
 
## Why
Modern component style + available hooks

### DIFF
--- a/src/components/ui/ShareMenu.jsx
+++ b/src/components/ui/ShareMenu.jsx
@@ -1,5 +1,5 @@
 /* global _ */
-import React, { Fragment } from 'react';
+import React, { useState, useRef, Fragment, useEffect } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


src/components/ui/ShareMenu.jsx
```diff
@@ -13,126 +13,114 @@ const twitterShareUrl = location => {
 
 const menu_height = 3 * 32;
```
G-Ray, line 14: Do you remember what's the reasoning behind this value ?

G-Ray (approved):

src/components/ui/ShareMenu.jsx
bbecquet: This seems to be the height of the share menu itself, made of 3 lines (copy, fb, twitter) of approx 32 of height. I guess we put it as a constant because we need it to position the menu before rendering it, but it's not beautiful :D

PR #1038, opened 2021-03-12
Install and configure prettier

bbecquet:
## Description
Install the [Prettier](https://prettier.io/) code formatter and configure it.
 - Update ESlint to v6 and install `eslint-plugin-prettier`, which consists in running Prettier as an ESlint rule.
 - Prettier and ESlint have some overlapping features. The best pratice is to disable the ESlint rules that may conflict with Prettier features. This is done automatically by using the `eslint-config-prettier` package, but I also removed useless rules in the ESlint config.
 - I also added `jest` and `node` as known environments for ESlint, so it is able to adjust some rules depending on the type of files.
 - I added two npm scripts : `prettier:check` to validate the code source, and `prettier:write` to format it. I limited prettier to the same directories as ESlint (`src`, `bin`, `build`, `tests`) 
 - I used `prettier:write` to run the formatter on the current codebase, formatting our JS files (hence the big diff in this PR).
 - Using [husky](https://github.com/typicode/husky), I added a pre-commit hook so that changed files are checked by ESlint (and prettier-ified if needed in the process) before the commit is created, ensuring the pushed code is consistent with formatting rules.

Additionally, it's a good idea to [configure your code editor](https://prettier.io/docs/en/editors.html) to integrate with Prettier, for example by autoformatting on save.

## Why
Get closer to dev practices on other Qwant projects and benefit from the same dev confort.

## Note about the config

I used this prettier config
```js
{
  // --- different rules ---
  "semi": true,
  "printWidth": 100,
  "arrowParens": "avoid",
  // ----------
  "singleQuote": true,
  "tabWidth": 2,
  "useTabs": false
}
```

Some rules are different from the one used "elsewhere" : end semi-colons on, max line length = 100, no parentheses for single args arrow functions.
This is to stay closer to our previous dev style, and not introduce a change too big right now without consulting the whole team. It is easy to change later if we decide we want to.
What is important is to have all the tools in place.


.eslintrc.json
```diff
@@ -1,116 +1,68 @@
...
-      "eslint:recommended",
-      "plugin:react/recommended",
-      "plugin:react-hooks/recommended"
+  "env": {
+    "browser": true,
+    "es6": true,
+    "jest": true,
+    "node": true
+  },
+  "plugins": ["prettier"],
```
G-Ray, line 8: According to the `eslint-plugin-prettier` doc, we should be able to get the same config in only one step using the recommended config : `This plugin ships with a plugin:prettier/recommended config that sets up both the plugin and eslint-config-prettier in one go.`

Some needed rules like "prettier/prettier": "error" would be automatically added a few eslint core rules which could cause issues would "automatically" be turned off.

Given these reasons, do you think we should use `plugin:prettier/recommended` instead ?

bbecquet: Oh, I took inspiration from the Phoenix config and didn't see this option.
It doesn't do much but it may be a good idea indeed, to simplify a bit and guard against future ESLint stuff that may conflict.
I'll give it a go and see if all works the same :+1:

G-Ray (dismissed):

G-Ray (approved):

amatissart (approved):

PR #1040, opened 2021-03-12
Qmaps 1968 Products menu to mimic the one on Search

bbecquet:
## Description
Integrate the standard Qwant Products menu panel.
Some details are not 100% like the design, but I think it's worth reviewing it already, and do (or not) the last bits in further PR.
Moreover, this PR introduces a feature flag for this so it's ok to merge step by step.

FYI, a big missing part are the links to mobile app stores. We'll decide later if we add them by duplicating search code or just accept we don't put them until we can share more core properly.

## Screenshots
|Desktop|Mobile|
|---|---|
|Capture d’écran de 2021-03-16 18-33-35|Capture d’écran 2021-03-16 à 18 34 04|


src/panel/menu/AppMenu.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+const AppMenu = ({ close, openProducts }) => {
+  const navTo = (url, options) => {
+    close();
+    window.app.navigateTo(url, options);
+  };
+
+  return <div className="menu-items">
+    <MenuItem
+      onClick={e => {
```
G-Ray, line 18: I think we should add an href to `/favs/` here :)

src/panel/menu/AppMenu.jsx
```diff
@@ -0,0 +1,62 @@
...
+import { PINK_DARK, ACTION_BLUE_BASE } from 'src/libs/colors';
+
+const AppMenu = ({ close, openProducts }) => {
+  const navTo = (url, options) => {
+    close();
+    window.app.navigateTo(url, options);
+  };
+
+  return <div className="menu-items">
+    <MenuItem
```
G-Ray, line 17: Should we add a text-underline decoration below an anchor with a `menu-item` class ?

src/panel/menu/AppMenu.jsx
bbecquet: I agree this would be a better markup than a link to "#", even if the navigation is done in JS. I'll do it.

bbecquet: It's done, thanks!

G-Ray (approved):

PR #1044, opened 2021-03-22
QMAPS-2051 update map style to add non breaking spaces in attribution text

xem:
## Description
bump qwant-basic-gl-style
cf: https://github.com/Qwant/qwant-basic-gl-style/pull/114

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.8.0",
     "@qwant/map-style-builder": "github:Qwant/map-style-builder#9808391",
-    "@qwant/qwant-basic-gl-style": "^1.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.8.0",
     "@qwant/map-style-builder": "github:Qwant/map-style-builder#9808391",
-    "@qwant/qwant-basic-gl-style": "^1.1.0",
+    "@qwant/qwant-basic-gl-style": "^1.3.1",
```
bbecquet, line 76: You have to do it using `npm install` so that the `package-lock.json` file is kept up-to-date.

amatissart: Yes. And I would expect `npm ci` to catch this. One more inconsistency from `npm` I suppose...

amatissart (approved):

PR #1045, opened 2021-03-24
QMAPS-2016 add direction in public transports' itinerary details

xem:
## Description
Tell where a public transport is headed to in itinerary details

## Screenshots
image

### DIFF
--- a/src/panel/direction/PublicTransportLine.jsx
+++ b/src/panel/direction/PublicTransportLine.jsx
@@ -21,20 +21,24 @@ const PublicTransportLine = ({ mode, info, showDirection = false }) => {
     }
   }
   const lineColor = info.lineColor ? Color('#' + info.lineColor) : Color('white');
-  return <>
-    <span
-      className={classnames('routePtLine', { 'routePtLine--dark': lineColor.isDark() })}
-      style={{
-        backgroundColor: lineColor.hex(),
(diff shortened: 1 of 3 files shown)


src/panel/direction/PublicTransportLine.jsx
```diff
@@ -20,15 +21,20 @@ const PublicTransportLine = ({ mode, info }) => {
...
+      className={classnames('routePtLine', { 'routePtLine--dark': lineColor.isDark() })}
+      style={{
+        backgroundColor: lineColor.hex(),
+        borderColor: lineColor.rgbNumber() === 0xffffff ? 'black' : lineColor.hex(),
+      }}
+    >
+      {type} {info.num}
+    </span>
+    {showDirection && <span className={classnames('routePtLine-direction', 'u-text--subtitle')}>
+      {_('Towards', 'direction')} {info.direction}
```
amatissart, line 35: I would recommend to use a formatted string here, as other languages may not use a simple prefix.  
See https://github.com/Qwant/erdapfel/blob/master/src/panel/direction/WalkLeg.jsx#L11-L13 for example.

xem: of course, my bad. fixed!

G-Ray (approved):

bbecquet (approved):

PR #1046, opened 2021-03-24
QMAPS-2017 remove distances in public transport routes & map labels

xem:
## Description
remove distances in public transport routes & map labels

## Screenshots
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-*  text=auto
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+  *  text=auto eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-  *  text=auto eol=lf
+*  text=auto
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -28,7 +28,7 @@ const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
   const element = document.createElement('div');
   element.innerHTML = `
     <div class="routeLabel-vehicleIcon ${getVehicleIcon(vehicle)}"></div>
-    <div class="${vehicle === 'publicTransport' ? 'routeLabel--public-transport' : ''}">
+    <div>
       <div class="routeLabel-duration">${formatDuration(route.duration)}</div>
       <div class="routeLabel-distance">${formatDistance(route.distance)}</div>
     </div>
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -33,10 +33,7 @@ const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
       <div class="routeLabel-distance">${formatDistance(route.distance)}</div>
     </div>
   `;
-  element.className = `routeLabel routeLabel--${anchor}`;
-  if (vehicle === 'publicTransport') {
-    element.className += ' routeLabel--single-line';
-  }
+  element.className = `routeLabel routeLabel--${anchor} routeLabel--${vehicle}`;
   element.dataset.id = route.id;
   element.onclick = () => {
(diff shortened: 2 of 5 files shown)


src/panel/direction/RouteSummaryInfo.jsx
```diff
@@ -5,17 +5,17 @@ import RouteVia from './RouteVia';
...
+const RouteSummaryInfo = ({ isFastest, route, vehicle, showDistance = false }) =>
   <div>
     <div className="u-text--title u-mb-xxxs route-summary-info-duration">
       {formatDuration(route.duration)}
     </div>
 
     <RouteVia className="u-mb-xs" route={route} vehicle={vehicle} />
 
-    <Badge className="u-mr-xs">
+    {showDistance && <Badge className="u-mr-xs">
```
G-Ray, line 16: I think spacing utils have been removed ?

bbecquet: Not yet, they are still useful until we switch to the real design system :)

src/adapters/scene_direction.js
```diff
@@ -30,10 +30,17 @@ const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
     <div class="routeLabel-vehicleIcon ${getVehicleIcon(vehicle)}"></div>
     <div>
       <div class="routeLabel-duration">${formatDuration(route.duration)}</div>
-      <div class="routeLabel-distance">${formatDistance(route.distance)}</div>
+      ${
```
bbecquet, line 33: I know in React it's common not to render stuff in the markup, but here it's less pratical.
Maybe instead of changing the markup we could just use a semantic class name on the element, like `routeLabel--publicTransport` instead of `--single-line`, and use it to hide the distance in CSS.

src/adapters/scene_direction.js
```diff
@@ -28,12 +28,15 @@ const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
   const element = document.createElement('div');
   element.innerHTML = `
     <div class="routeLabel-vehicleIcon ${getVehicleIcon(vehicle)}"></div>
-    <div>
+    <div class="${vehicle === 'publicTransport' ? 'routeLabel--public-transport' : ''}">
```
bbecquet, line 31: Why not simply use the existing `if` below which already sets a special class on the element and adapt the selector?

.gitattributes
```diff
@@ -1 +1 @@
-*  text=auto
\ No newline at end of file
+  *  text=auto eol=lf
```
bbecquet, line 1: This shouldn't be kept in this PR.
And if you have the correct git settings, everything should work fine without touching this file. Can you confirm?

src/adapters/scene_direction.js
xem: huh, I was dumb :D indeed

.gitattributes
xem: indeed

src/adapters/scene_direction.js
```diff
@@ -34,6 +34,9 @@ const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
     </div>
   `;
   element.className = `routeLabel routeLabel--${anchor}`;
+  if (vehicle === 'publicTransport') {
+    element.className += ' routeLabel--single-line';
```
bbecquet, line 38: Sorry to add yet another comment on this bunch of code, but as I suggested in a previous review, `routeLabel--publicTransport` would be a better name.
`--single-line` is too descriptive of what it does and not of why we do it. It's better when we maintain the CSS to understand why.
Moreover, there are other planned developments on this feature where we enrich the content of these labels for public transport, so it will be already ready for other stuff than just hiding a line.

BTW, you could even simplify this by writing the transport mode as a class variant without any `if`.

bbecquet (approved):

PR #1051, opened 2021-03-31
QMAPS-1883 refactor close buttons

xem:
## Description

Done:
- Mobile: remove close button on all panels except directions, menu, modal, alert, route details, favorites.
- Desktop: remove close button on POI panel only
- PoI Panel (PC/Mobile): whatever the navigation origin, put PoI name in search field and quit PoI panel when clicking close button
- Mobile PoI panel: if we're coming from a PoI list or favorites list: show a return arrow next to search input that goes back to it
- Put the return button logic in PanelManager
- Put the search field filling logic in PanelManager
- Make the back buttons work correctly (on PC and Mobile) even after we click on many PoIs in a row (history(-1) isn't enough)
- Fill search input correctly (with the right timing) when a PoI is opened
- Avoid transitions blinking when opening a poi from the list / returning to the list / clicking many PoIs in a row 
- Put "Favorites" in search field when favorites list is opened
- Fix tests

DEMO:
https://www.loom.com/share/0d9a1bc709b0403197b8611b6016174e

TODO in another ticket?
- Use the same return button on mobile for both closing suggest and returning to list?


src/components/ui/Panel.jsx
```diff
@@ -330,7 +330,9 @@ class Panel extends React.Component {
             {floatingItemsRight && size !== 'maximized' && (
               <FloatingItems position="right" items={floatingItemsRight} />
             )}
-            {onClose && <CloseButton onClick={onClose} className="panel-close" />}
+            {onClose && !isMobileDevice() && (
```
bbecquet, line 333: No need for the imperative `isMobileDevice` function, we are already in a `DeviceContext` here, `isMobile` is already available.

bbecquet:
I've noticed some problems:

The transition effect in the top bar is repeated when you go from one POI to another.
Peek 09-04-2021 15-10

When in multiple POI mode (aka. category), the category name appears briefly in the field after being replaced by the clicked POI name 
Peek 09-04-2021 15-14
This is because the search input change is not done in a centralized way like before but managed in the `PanelManager` **and** in the `POIPanel`.

BTW, all these search bar manipulations direclty in POIPanel are a serious step back from what we achieved some months ago, which was to centralize the management of its state in a single place to simplify and avoid consistency errors…
This is my fault, I should have pushed to merge https://github.com/Qwant/erdapfel/pull/997 before, it would have enforced state centralization.
In the meantime, we must find a way here to manage as much things as we can outside POIPanel.

views/index.ejs
```diff
@@ -49,6 +49,7 @@
         ></button>
         <div class="search_form__wrapper empty">
           <div class="search_form__return icon-arrow-left"></div>
+          <div class="search_form__return-to-list icon-arrow-left" onclick="window.history.go(-1); return false;"></div>
```
bbecquet, line 52: There is already an arrow element above ready to do stuff. Why not attaching some behavior to it in JS instead of creating yet another hardcoded button here.

Another problem: if I understand correctly the ticket, using the arrow when navigating on a POIs in a list should go back to the list itself. If you always go back in history, you will reach the previous POI if you have navigated between multiple ones from the same list by clicking on them.
This is a very confusing navigation pattern in this feature :(

src/panel/poi/PoiPanel.jsx
```diff
@@ -56,6 +57,13 @@ export default class PoiPanel extends React.Component {
         this.setState({ isPoiInFavorite });
       }
     });
+
+    // Show return arrow on mobile if user comes from PoI / favorites list
```
bbecquet, line 61: As always, we should avoid manipulating DOM directly in React if possible, especially stuff outside the component render tree.
Some time ago I centralized every top bar manipulations in `PanelManager`, especially in [this method](https://github.com/Qwant/erdapfel/blob/master/src/panel/PanelManager.jsx#L89-L105), to limit them and prepare the way for the top bar migration to React.

As a result, we should find how to do all this (this block and the two others below) in `PanelManager`.

src/panel/poi/PoiPanel.jsx
```diff
@@ -218,6 +230,11 @@ export default class PoiPanel extends React.Component {
       return null;
     }
 
+    // Show PoI name in search field
+    SearchInput.setInputValue(poi.name);
```
bbecquet, line 234: This line may be the only thing not easily moveable to PanelManager, as the POI is not known by it… which is a recurring problem that calls for a centralized current POI state upstream.
I think it's ok to keep it here for now.

tests/integration/tests/autocomplete.js
```diff
@@ -149,9 +149,9 @@ test('mouse navigation', async () => {
   expect(searchValue.trim()).toEqual(TypedSearch);
 
   await autocompleteHelper.clickResult(1);
-  const selectedSearchValue = await autocompleteHelper.getSearchInputValue();
```
bbecquet, line 152: This is very strange indeed… According to the failing assertation, it behaves like the second element is clicked instead of the first one, but I can't see why.

toto

tests/integration/tests/autocomplete.js
```diff
@@ -250,7 +250,7 @@ test('Search Query', async () => {
   const searchValue = await getInputValue(page, '#search');
 
   // search input is filled with query
-  expect(searchValue).toEqual(searchQuery);
+  expect(searchValue).toEqual("Musée d'Orsay");
```
amatissart, line 253: If this is the expected behavior, then the comment message is obsolete and should be update too.

But is it really an expected change in this case? I fear that it could cause frustration to the user that can't complete the query anymore, after the input value has been overridden with a possibly irrelevant result name.

tests/integration/tests/autocomplete.js
```diff
@@ -149,9 +149,9 @@ test('mouse navigation', async () => {
   expect(searchValue.trim()).toEqual(TypedSearch);
 
   await autocompleteHelper.clickResult(1);
-  const selectedSearchValue = await autocompleteHelper.getSearchInputValue();
-  const expectedLabelName = mockAutocomplete.features[0].properties.geocoding.name;
-  expect(selectedSearchValue).toEqual(expectedLabelName);
+  //const selectedSearchValue = await autocompleteHelper.getSearchInputValue();
+  //const expectedLabelName = mockAutocomplete.features[0].properties.geocoding.name;
+  //expect(selectedSearchValue).toEqual(expectedLabelName);
```
amatissart, line 154: I think that this is due to how the input value is updated twice now.  
First the suggest value is used (this is what this test case verifies), and secondly the name used on the `PoiPanel` is rendered after its content has been fetched from the API. Here a single mock file for "Musée d'Orsay" is defined for all requests to Idunn. That's why this test is currently failing.

This double update is visible on a query such as "brest".
1. Type **"brest"** in the search bar 
2. Click on the first suggestion "Brest (29200)"
3. The input value is briefly set to **"Brest (29200)"** 
4. PoiPanel is rendered and the input value is now **"Brest"**

xem: Ah. I think the double update comes from the fact that I set the search field in PanelManager AND in PoiPanel.
If I do it only in PanelManager, nothing appears in the field after a page refresh with "hotload PoI" 
If I do it only in PoiPanel, it appears with a too long delay.
So: I'll try to do everything from the panelManager with a special case when window.hotLoadPoI is detected.

tests/integration/tests/autocomplete.js
xem: Sorry this is just a residue of when I fiddled with the tests trying to make them work again. I think I can keep it as it was before

src/panel/PanelManager.jsx
```diff
@@ -145,6 +215,8 @@ export default class PanelManager extends React.Component {
         options: {},
         panelSize: 'default',
       });
+
+      SearchInput.setInputValue(_('Favorites', 'search bar'));
```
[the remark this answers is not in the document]
xem: Is an empty field better? (we said no the other day but it was just orally)
And indeed it's not in any design file.
@Charlotte-Ziegler  ?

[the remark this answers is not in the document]
xem: removed

tests/integration/tests/autocomplete.js
```diff
@@ -250,7 +250,7 @@ test('Search Query', async () => {
   const searchValue = await getInputValue(page, '#search');
 
   // search input is filled with query
-  expect(searchValue).toEqual(searchQuery);
+  expect(searchValue).toEqual('test result 1');
```
amatissart, line 253: As discussed, this behavior looks good to me at this stage. The comment related to this check should be updated though, to clarify the intention. The query does not persist in the search input anymore.

bbecquet (approved):

PR #1052, opened 2021-04-01
qwant-maps-common: message keys with spaces were missing due to named exports

amatissart:
## Description
Fixes an issue with translations in qwant-maps-common: keys with spaces were not available in the messages built by rollup.

## Why
The `json()` plugin in Rollup [includes named exports by default](https://github.com/rollup/plugins/tree/master/packages/json#namedexports), in addition to the `default` export.
But named exports are only used for valid JS names (i.e strings without spaces).
The correct way is to use the "default" export when loading the module.

### DIFF
--- a/local_modules/qwant-maps-common/src/gettext.js
+++ b/local_modules/qwant-maps-common/src/gettext.js
@@ -1,15 +1,14 @@
-import { options as i18nOptions } from "../i18next-scanner.config.js";
+import { options as i18nOptions } from '../i18next-scanner.config.js';
 
 const supportedLanguages = i18nOptions.lngs;
 const defaultLanguage = i18nOptions.defaultLng;
 const messages = {};
 supportedLanguages.forEach(lang =>
-  import(`../i18n/${lang}/resource.json`)
-    .then(resource => messages[lang] = resource)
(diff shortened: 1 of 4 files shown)


local_modules/qwant-maps-common/src/gettext.js
```diff
@@ -1,15 +1,14 @@
-import { options as i18nOptions } from "../i18next-scanner.config.js";
+import { options as i18nOptions } from '../i18next-scanner.config.js';
 
 const supportedLanguages = i18nOptions.lngs;
 const defaultLanguage = i18nOptions.defaultLng;
 const messages = {};
 supportedLanguages.forEach(lang =>
-  import(`../i18n/${lang}/resource.json`)
-    .then(resource => messages[lang] = resource)
+  import(`../i18n/${lang}/resource.json`).then(resource => (messages[lang] = resource.default))
```
amatissart, line 7: Here is the actual fix.

bbecquet:
@amatissart what do you think about switching the `namedExports` option to `false` in the plugin initialization so that the generated files don't include useless and incomplete data.

amatissart:
> @amatissart what do you think about switching the `namedExports` option to `false` in the plugin initialization so that the generated files don't include useless and incomplete data.

yes, done.

bbecquet (approved):

PR #1055, opened 2021-04-02
Use the preloaded POI (if any) center as initial map position 

bbecquet:
## Description
When a POI has been preloaded on the server side, use its center (or bbox) as initial map view.

The first commit is a refacto of the map padding management, so that it's simpler and consistent across all map view operations. Instead of passing explicit paddings on each `fitBounds`, `jumpTo`, etc. we set global paddings on the map and let MapBox-GL do its stuff.
Here, it guarantees the view from the initial center/zoom will be the same as when the POIPanel centers on the POI. 

## Why
When arriving on Maps with a single POI in the url (typically when coming from Qwant Search), prevent the "jump" from the initial map position to the POI center (happening because we center the map when the POI data is loaded asynchronously by the front).
As the POI is already preloaded on the server side, we can use its position as soon as the map loads, so there is no map change when the POI Panel re-fetch the data.
It shows we could probably avoid this fetch completely in this case, but let's focus on fixing the visual problem here.

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -29,7 +29,16 @@ function Scene() {
   this.savedLocation = null;
 }
 
+const getPoiView = poi => ({
+  center: poi.geometry.center,
+  zoom: getBestZoom(poi),
+  bounds: poi.geometry.bbox,
+});
+
 Scene.prototype.getMapInitOptions = function (locationHash) {
+  if (window.hotLoadPoi) {
+    return getPoiView(window.hotLoadPoi);
+  }
   if (locationHash) {
     return {
       zoom: locationHash.zoom,
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -89,6 +89,8 @@ Scene.prototype.initMapBox = async function (locationHash) {
     locale,
     ...(await this.getMapInitOptions(locationHash)),
   });
+  // @MAPBOX: This method isn't implemented by the Mapbox-GL mock
+  this.mb.setPadding = this.mb.setPadding || (() => {});
   this.mb.setPadding(getCurrentMapPaddings());
 
   this.popup.init(this.mb);
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -3,7 +3,7 @@ import PoiPopup from './poi_popup';
 import MobileCompassControl from '../mapbox/mobile_compass_control';
 import ExtendedControl from '../mapbox/extended_nav_control';
 import { map as mapConfig } from 'config/constants.yml';
-import { getMapPaddings, getMapCenterOffset, isPositionUnderUI } from 'src/panel/layouts';
+import { getCurrentMapPaddings, isPositionUnderUI } from 'src/panel/layouts';
 import nconf from '@qwant/nconf-getter';
 import MapPoi from './poi/map_poi';
 import { getLastLocation, setLastLocation } from 'src/adapters/store';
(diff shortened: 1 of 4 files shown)


src/panel/layouts.js
```diff
@@ -26,13 +21,20 @@ export function getMapPaddings({ isMobile, isDirectionsActive, isIframe }) {
   const topUIElement = isDirectionsActive ? '.direction-panel' : '.top_bar';
   const topUIHeight = document.querySelector(topUIElement)?.clientHeight || 0;
   return {
-    bottom: mobileBottomPanelHeight + ADDITIONAL_PADDING / 2,
+    bottom: MOBILE_BOTTOM_PADDING,
     top: topUIHeight + ADDITIONAL_PADDING / 2,
-    right: 70,
+    right: 20,
```
amatissart, line 26: I think that the reasoning behind the value 70 was to avoid meaningful items (e.g route outlines or markers) to be hidden by the floating button on the right.  
Is that not still relevant ?

src/panel/layouts.js
```diff
@@ -26,13 +21,20 @@ export function getMapPaddings({ isMobile, isDirectionsActive, isIframe }) {
   const topUIElement = isDirectionsActive ? '.direction-panel' : '.top_bar';
   const topUIHeight = document.querySelector(topUIElement)?.clientHeight || 0;
   return {
-    bottom: mobileBottomPanelHeight + ADDITIONAL_PADDING / 2,
+    bottom: MOBILE_BOTTOM_PADDING,
```
amatissart, line 24: Are we sure that there is no situation where a dynamic value would still be required here to fit the remaining space at the top of a larger panel ?

amatissart:
> As the POI is already preloaded on the server side, we can use its position as soon as the map loads, so there is no map change when the POI Panel re-fetch the data.
It shows we could probably avoid this fetch completely in this case, but let's focus on fixing the visual problem here.

To clarify, the POI Panel does not re-fetch the data when `window.hotLoadPoi` exists (it is actually its main purpose). The current implementation uses [`execOnMapLoaded()`](https://github.com/Qwant/erdapfel/blob/master/src/panel/poi/PoiPanel.jsx#L122-L130) to ensure that the map is available before setting the map position. In this case, the panel is typically displayed before the map is completely displayed, but the user can see the initial map position before "map_loaded" is triggered. Hence the visual bug.

src/panel/layouts.js
bbecquet: I haven't seen any for now. On the contrary we already display things below the panel (category POIs for example), and when we resize the panel manually we don't want that to have any effect on the map center, it slides above it.
But I agree this isn't ideal like that.

src/panel/layouts.js
bbecquet: I took this shortcut so that the usable map viewport is symmetric, otherwise POIs are not visually centered on mobile.
Actually I've always thought this was too much of map space potentially wasted for a single button which poses no problem on the majority of cases…
Maybe I was too eager to simplify all this.

amatissart (approved):

src/adapters/scene.js
```diff
@@ -81,6 +89,9 @@ Scene.prototype.initMapBox = async function (locationHash) {
     locale,
     ...(await this.getMapInitOptions(locationHash)),
   });
+  // @MAPBOX: This method isn't implemented by the Mapbox-GL mock
+  this.mb.setPadding = this.mb.setPadding || (() => {});
```
G-Ray, line 93: we might check for the NODE_ENV variable to be "test" ?

bbecquet: Do you mean wrapping that in `if (process.env.NODE_ENV === 'test') { … }`?

Unfortunately it doesn't work with our current setup, as the test version is built with the `production` mode, so Webpack just removes the block from the build and the fix doesn't get applied.
I guess it deserves a talk about our test practices, maybe we could change that, but let's do it separately from this PR.

PR #1059, opened 2021-04-08
Introduce a React hook for easy access to config

bbecquet:
## Description
Add a `useConfig` hook for easy and more natural access to the config from React components, without having to use `nconf-getter` each time.

As it's the most common case to access only a subset of the config, `useConfig` accepts an optional argument to return only the designated part.

In the future, other hooks can be added to the `/hooks` folder and index file.

### DIFF
--- a/src/panel/Menu.jsx
+++ b/src/panel/Menu.jsx
@@ -8,14 +8,13 @@ import Telemetry from 'src/libs/telemetry';
 import { Flex, CloseButton } from 'src/components/ui';
 import { IconMenu, IconApps } from 'src/components/ui/icons';
 import { DeviceContext } from 'src/libs/device';
-import nconf from '@qwant/nconf-getter';
-
-const displayProducts = nconf.get().burgerMenu?.products;
+import { useConfig } from 'src/hooks';
 
 const Menu = () => {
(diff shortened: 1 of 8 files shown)


G-Ray (dismissed):

src/panel/Menu.jsx
```diff
@@ -8,14 +8,13 @@ import Telemetry from 'src/libs/telemetry';
...
-import nconf from '@qwant/nconf-getter';
-
-const displayProducts = nconf.get().burgerMenu?.products;
+import { useConfig } from 'src/hooks';
 
 const Menu = () => {
   const [openedMenu, setOpenedMenu] = useState(null);
   const menuContainer = useRef(document.createElement('div'));
   const isMobile = useContext(DeviceContext);
+  const displayProducts = useConfig('burgerMenu');
```
amatissart, line 17: Should not that be :
```suggestion
  const displayProducts = useConfig('burgerMenu').products;
```
?

amatissart (changes requested):

src/panel/Menu.jsx
bbecquet: Indeed! Thank you.

amatissart (approved):

PR #1060, opened 2021-04-08
Use Node v12

bbecquet:
## Description
Update the base Docker image to Node v12.
Also add `engines` npm hints matching the current versions needed to build and run Erdapfel.

## Why
Node 10 is reaching its end of life at the end of the month.

We are not migrating directly to Node v14 yet because `chokidar@2`, a depencendy of Webpack 4, may break on it (https://github.com/webpack/watchpack/issues/137).
We'll address that separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,9 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Use Node.js
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2
       with:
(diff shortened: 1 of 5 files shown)


amatissart:
The GitHub workflow and the README file need to be updated too.

bbecquet:
@amatissart Thanks! Done. I've also updated the version of https://github.com/actions/setup-node to v2 in the process (no big changes).

G-Ray (approved):

amatissart (approved):

PR #1064, opened 2021-04-13
QMAPS-2105 Use 'bbox' url param as initial map position

bbecquet:
## Description
Use the `bbox` url param that can be specified by some `/places` requests arriving on Maps (for example those coming from the "Multi" instant answer) as the initial view param.

*I've not been able to prevent a remaining little jump on start, which seems to be from the bbox with no padding, to one taking into account the paddings set by `setPadding` after the map is created.
Unfortunately there is no `padding` option for the [Map instance creation](https://docs.mapbox.com/mapbox-gl-js/api/map/), and using the same paddings in the optional `fitBoundsOptions` results in an even bigger shift.*

## Why
Prevents the map initializing on a different start position before "jumping" to the bbox when the data matching the query is asynchronously fetched by the `CategoryPanel`.

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -14,6 +14,7 @@ import { createDefaultPin } from '../adapters/icon_manager';
 import LatLonPoi from './poi/latlon_poi';
 import { isMobileDevice } from 'src/libs/device';
 import { parseMapHash, getMapHash } from 'src/libs/url_utils';
+import { parseBboxString } from 'src/libs/bounds';
 import { toUrl, getBestZoom } from 'src/libs/pois';
 import Error from 'src/adapters/error';
 import { fire, listen } from 'src/libs/customEvents';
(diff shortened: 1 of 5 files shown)


src/adapters/scene.js
```diff
@@ -87,7 +95,7 @@ Scene.prototype.initMapBox = async function (locationHash) {
     maxZoom: 20,
     interactive: window.no_ui ? false : true,
     locale,
-    ...(await this.getMapInitOptions(locationHash)),
+    ...(await this.getMapInitOptions({ locationHash, bbox })),
```
amatissart, line 98: `getMapInitOptions` is no longer async: `await` can be removed.

amatissart (approved):

src/adapters/scene.js
bbecquet: Indeed, thank you.
To avoid another cycle of modify/review I'll merge as-is and remove the `await` in the related PR which touches this file.

PR #1065, opened 2021-04-15
[Clean-up] Split hotload poi fetch and OpenGraph meta

bbecquet:
## Description
Split our `og_meta` Express middleware into two separate middlewares with clear responsibilities, instead of doing not-so-related stuff in the same one:
 - `prefetch_poi`: parses the URL for a potential POI specified in the url and fetches it, making it available to the next step in the Express chain as `res.locals.poi` 
 - `og_meta`: writes the right OpenGraph meta info. If a POI has been pre-fetched before, uses its info.

## Why
Prefetching the bbox corresponding to a `?q=` query will have to be done in a middleware, similar to the hotload POI fetch, and I wanted to clean the situation before.
Moreover, any clean-up in this code will ease a potential switch to SSR later.

### DIFF
--- a/bin/middlewares/prefetch_poi.js
+++ b/bin/middlewares/prefetch_poi.js
@@ -43,13 +43,14 @@ module.exports = function (config) {
       .then(poi => {
         if (poi) {
           res.locals.poi = poi;
+          next();
         } else {
           res.redirect(307, config.system.baseUrl);
         }
       })
       .catch(error => {
         req.logger.error({ err: error });
-      })
-      .finally(next);
+        next();
+      });
   };
 };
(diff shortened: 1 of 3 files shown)


bin/middlewares/prefetch_poi.js
```diff
@@ -0,0 +1,55 @@
...
+        if (poi) {
+          res.locals.poi = poi;
+        } else {
+          res.redirect(307, config.system.baseUrl);
+        }
+      })
+      .catch(error => {
+        req.logger.error({ err: error });
+      })
+      .finally(next);
```
amatissart, line 53: `next` should not be called after `res.redirect()`. This would cause an error on the server:
```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
```

bbecquet: Thank you, it's fixed.

amatissart (approved):

PR #1066, opened 2021-04-16
QMAPS-2088 Use `?q=` intention bbox as initial map position

bbecquet:
## Description
Add a new Express middleware to interpret incoming root urls containing a `?q=` param. If it resolves to a multi-place intention or a geocoding result, redirect to either a `/places` or a `/place` url, allowing to use the bbox or POI center as the initial map position (as now supported by https://github.com/Qwant/erdapfel/pull/1064 and https://github.com/Qwant/erdapfel/pull/1055).

## Why
Prevents the map initializing on a different start position before "jumping" to the bbox when the data is asynchronously fetched by the client app.


bin/middlewares/query_bbox.js
```diff
@@ -0,0 +1,55 @@
...
+      if (error.response && error.response.status === 404) {
+        return null;
+      }
+      throw error;
+    }
+  }
+
+  return function (req, res, next) {
+    const absoluteUrl = `${req.protocol}://${req.get('host')}${req.originalUrl}`;
+    const fullTextQuery = new URL(absoluteUrl).searchParams.get('q');
```
amatissart, line 37: You would probably prefer using [`req.query`](https://expressjs.com/en/api.html#req.query)
```suggestion
    const fullTextQuery = req.query.q;
```

bbecquet: Ah yes, I'm so used to code on the client that I forgot Express has some easy shortcuts :)
Thanks!

bin/middlewares/fullText_query.js
```diff
@@ -0,0 +1,80 @@
+const axios = require('axios');
+const yaml = require('node-yaml');
+
+module.exports = function (config) {
+  const geocoderUrl = config.services.geocoder.url;
```
amatissart, line 5: This `geocoder.url` is used by the client, and may not be accessibly from the backend. Similarly to "prefetch_poi", `config.server.services.idunn.url` should be preferred when it's defined.

Actually, all environments now use a Idunn route as their geocoder URL, so this `geocoder.url` configuration could be removed completely., to derive the URL from `idunn.url`.

bin/middlewares/fullText_query.js
```diff
@@ -0,0 +1,80 @@
...
+        );
+
+        return `${config.system.baseUrl}places/?${params.toString()}`;
+      }
+
+      const firstPoi = (response.data.features || [])[0];
+      if (firstPoi) {
+        return `${config.system.baseUrl}place/${firstPoi.properties.geocoding.id}`;
+      }
+      return null;
```
amatissart, line 51: `/noresult` could be used as the redirect url in this case, no ?

bin/middlewares/fullText_query.js
```diff
@@ -0,0 +1,80 @@
...
+        if (redirectUrl) {
+          res.redirect(307, redirectUrl);
+        } else {
+          next();
+        }
+      })
+      .catch(error => {
+        req.logger.error({ err: error });
+        next();
+      });
```
amatissart, line 78: A possible regression with this redirect mechanism is that the raw query is not preserved in the search bar.
This expected behavior in this case is unclear (also discussed in #1051).  
Do you think it would be relevant to pass the raw query as an extra parameter in the URL?

bbecquet: According to the new way things will work with the search bar, I believe it would be overwritten in any case by the POI name… I don't agree with this behavior but the design team seems to be quite positive with it.

I can pass it in any case to make it available to the front to decide what to do with it.

bin/middlewares/fullText_query.js
bbecquet: Ok. I'll remove the use of this here and open an issue to simplify this config later.

bin/middlewares/fullText_query.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const categories = yaml.readSync('../../config/categories.yml');
+  const isCategoryValid = type => categories.find(category => category.name === type);
+
+  const removeNullEntries = obj =>
+    Object.entries(obj)
+      .filter(([_key, value]) => value !== null && value !== undefined)
+      .reduce((result, [key, value]) => ({ ...result, [key]: value }), {});
+
+  if (isNaN(idunnTimeout)) {
```
G-Ray, line 18: we might move this block right after l.8 ? In order to raise error earlier, and handle the NaN case next to the Number construction

bin/middlewares/fullText_query.js
```diff
@@ -0,0 +1,71 @@
...
+module.exports = function (config) {
+  const idunnBaseUrl = config.server.services.idunn.url || config.services.idunn.url;
+  const geocoderUrl = idunnBaseUrl + '/v1/autocomplete';
+  const useNlu = config.services.geocoder.useNlu;
+  const idunnTimeout = Number(config.server.services.idunn.timeout);
+
+  const categories = yaml.readSync('../../config/categories.yml');
+  const isCategoryValid = type => categories.find(category => category.name === type);
+
+  const removeNullEntries = obj =>
```
G-Ray, line 13: a `removeNullEntries` function is defined in `url_utils:js`.
Do you think we can reuse this one ?

bbecquet: Currently we can't, as it is a client module which can't be required by the server code. I plan to change that soon, but it requires some config change on the server that I didn't want to do in this PR, so I just copied this code knowing it's not ideal.

bin/middlewares/fullText_query.js
bbecquet: Yep why not.

bin/middlewares/fullText_query.js
```diff
@@ -0,0 +1,72 @@
...
+
+      return `${config.system.baseUrl}places/?${params.toString()}`;
+    }
+
+    const firstPoi = (response.data.features || [])[0];
+    if (firstPoi) {
+      return `${config.system.baseUrl}place/${firstPoi.properties.geocoding.id}?q=${query}`;
+    }
+
+    return `${config.system.baseUrl}noresult?q=${query}`;
```
amatissart, line 53: `query` should be url-encoded, otherwise extra parameters could be injected.

amatissart (commented):
### About telemetry events
This PR impacts 2 telemetry events:
 * **`app_start`**: `direct_search` field used to be `true` when `q` was present as a direct search query
 * **`suggest_selection`**: `from_url_query` field used to be `true` for the selection triggered by a direct search.
 
I would suggest to simply remove these fields that introduce unnecessary complexity, and use the existing `?client` to distinguish between the different sources of queries.
The mechanism used to trigger a search based on the URL parameter could actually be removed completely on the client side.

Then, the new middleware "fullText_query.js" could forward the `client` parameter from the original URL (if present), and set `client=direct-search` by default in the redirect URL. 

@bbecquet What do you think ? And should this cleanup be done in a separate PR?

bbecquet:
@amatissart Thanks for the encoding remark.
About the telemetry/simplification, I'll do it in a separate PR as agreed IRL :)

amatissart (approved):

PR #1070, opened 2021-04-23
Simplification: use a single top bar back button

bbecquet:
## Description
Follow-up to https://github.com/Qwant/erdapfel/pull/1051.
In that previous PR, a second arrow-left button was added in the top bar static markup, to manage specifically the case of navigating back from POI to category/list. The existing button acting as a neutral area to blur the suggest input field wasn't reused, even if it had exactly the same style.

In this PR, we refactor this button part and the state management of `<PanelManager>` so we don't need two buttons. The same DOM element is used for both actions (back and suggest blur), the switch between them being controlled by the state.

(I also tweaked autocomplete integration tests, after noticing they behaved erratically from one run to another, and differently between CI and local).

## Why
Less HTML/JS/CSS, consistency, and not diverging too much from the target architecture of the [TopBar as a React component](https://github.com/Qwant/erdapfel/pull/997), which will update its DOM and event handler dynamically, rendering only what's necessary.

### DIFF
--- a/src/panel/PanelManager.jsx
+++ b/src/panel/PanelManager.jsx
@@ -113,25 +113,17 @@ export default class PanelManager extends React.Component {
   }
 
   updateTopBarReturnButton({ poiFilters = {}, isFromFavorite, poi = {} } = {}) {
-    const topBarReturnButton = document.querySelector('.search_form__return-to-list');
-    const backAction =
-      poiFilters.category || poiFilters.query
-        ? this.backToList
-        : isFromFavorite
-        ? this.backToFavorite
-        : () => {};
-    topBarReturnButton.onclick = e => {
-      backAction(e, poiFilters);
-    };
-
     const topBarHandle = document.querySelector('.top_bar');
-
-    // Show return arrow (on mobile) if user comes from PoI / favorites list
+    const topBarReturnButton = document.querySelector('.search_form__return');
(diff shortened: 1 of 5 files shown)


src/panel/PanelManager.jsx
```diff
@@ -113,26 +118,21 @@ export default class PanelManager extends React.Component {
...
+        }
+        backAction(e, poiFilters);
+      };
       topBarHandle.classList.add('top_bar--poi-from-list');
-    }
-    // Hide return button when not on a POI anymore
-    else {
-      const topBarHandle = document.querySelector('.top_bar');
+    } else {
+      topBarReturnButton.removeAttribute('onmousedown');
```
G-Ray, line 135: should we remove the attribute on componentWillUnmount too ?

bbecquet: Why not but in practice it will have no effect as this component is never unmounted. Actually it's already full of stuff not properly cleaned, it has no `componentWillUnmount` yet…
I'm not saying it's good, but it will be easier to do it properly when :
 a) the top bar is migrated to React (which I will do at last just after this PR is merged, I'm fed up with waiting)
 b) `PanelManager` is written as a functional component, with explicit side effects (which I'll do after the React Top Bar)

xem (commented):
Code & local test: 👍

xem (approved):

PR #1071, opened 2021-04-26
Remove unused 'no UI' (iframe) mode

bbecquet:
## Description
Remove the 'no UI' mode, which was meant to be used as a quick and temporary solution for a legacy IA… but sadly never got used because modern IA got quickly implemented in the new search.
In practice it's no use to keep it. Even if we want an "embedded" mode at some point it will have different requirements than IA and we'll do it in a more robust way.
So just clean the code by removing these useless parts… and let's hope we'll be better at anticipating real needs next time.

## Why
Dead code :cry: :zombie:


xem (approved):
nice cleanup \o/

PR #1072, opened 2021-04-27
[Clean-up] Client simplifications after server-side `?q=` redirect

bbecquet:
## Description
Simplify the routing and telemetry logic on the client following the new server-side redirection of full text queries introduced in https://github.com/Qwant/erdapfel/pull/1066, following [great advice by @amatissart](https://github.com/Qwant/erdapfel/pull/1066#pullrequestreview-644882174)
 - Remove the need for a client side redirection of `/?q=` queries as start url (as it's now done exclusively on the server)
 - Remove extra telemetry fields related to tracking this special case
 - Use `'direct-search'` as a new `client` param to track this case using the generic fields

*Note: I removed the integration test case that checked the client-side redirection on app startup, because it couldn't work anymore. I replaced it with testing the 3 cases (no result, single, multi) of server-side redirect, which required some new mocking of Idunn calls from Express.*

### DIFF
--- a/bin/middlewares/fullText_query.js
+++ b/bin/middlewares/fullText_query.js
@@ -22,7 +22,8 @@ module.exports = function (config) {
       .filter(([_key, value]) => value !== null && value !== undefined)
       .reduce((result, [key, value]) => ({ ...result, [key]: value }), {});
 
-  async function getRedirectUrl(query, res) {
+  async function getRedirectUrl(req, res) {
+    const { q: query, client = 'direct-search' } = req.query;
     const response = await axios.get(geocoderUrl, {
       params: {
         lang: res.locals.language.code,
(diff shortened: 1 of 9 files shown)


bbecquet:
I'm reverting back to a draft until I add the tests for the server-side redirection.

amatissart:
> I'm reverting back to a draft until I add the tests for the server-side redirection.

ok :+1: Note that some server-side mocked requests are already defined in https://github.com/Qwant/erdapfel/blob/master/tests/integration/server_start.js using "nock". Maybe the same mechanism can be used here.

bbecquet:
@amatissart 
> Note that some server-side mocked requests are already defined

Yup I've seen that :+1: I started adding stuff there, it works fine, things should go smoothly :)

amatissart (approved):

PR #1073, opened 2021-04-27
QMAPS-2052 add transport icons in map labels

xem:
## Description
- Add public transport icons into map labels
- 3 icons max, excluding "WALK". 4th and more are omitted
- icons can be repeated (ex: bus > bus > bus)
- small, dark ">" icons between each
- maybe put the duration on the line below on mobile
- if more than 4 vehicles: put "..." in the middle (deprecated)
- UPDATE: replace "..." with a little square containing "+2", "+3", etc.
- We may need to adjust bbox padding on desktop

## Screenshots
image
image
image

### DIFF
--- a/src/adapters/scene_direction.js
+++ b/src/adapters/scene_direction.js
@@ -18,6 +18,13 @@ import { fire, listen } from 'src/libs/customEvents';
 import { getLabelPositions } from 'src/libs/route_labeler';
 import { isMobileDevice } from 'src/libs/device';
 
+const VEHICLES = {
+  TRAIN: 'train',
+  SUBWAY: 'metro',
+  BUS_CITY: 'bus',
+  TRAM: 'tram"',
+};
+
 const createMarker = (lngLat, className = '', options = {}) => {
   const element = document.createElement('div');
   element.className = className;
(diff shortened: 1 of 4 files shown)


src/adapters/scene_direction.js
```diff
@@ -26,13 +26,35 @@ const createMarker = (lngLat, className = '', options = {}) => {
 
 const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
   const element = document.createElement('div');
-  element.innerHTML = `
-    <div class="routeLabel-vehicleIcon ${getVehicleIcon(vehicle)}"></div>
+  let html = ``;
```
bbecquet, line 29: Looks like this function suddenly got more complicated, with a lot of low-level HTML manipulation.
Maybe it's time to use our util `renderStaticReact` to be able to extract that as a clean React component and still use it on the map. This is what we already do for `<PoiPopup>`.
What do you think?

xem: Sure. I tried to do it like that, it may be not exactly as you would have done it, but you're free to fix it or make me fix it if you have an idea of best practice to use :)

src/adapters/scene_direction.js
```diff
@@ -26,13 +36,59 @@ const createMarker = (lngLat, className = '', options = {}) => {
 
 const createRouteLabel = (route, vehicle, { lngLat, anchor }) => {
   const element = document.createElement('div');
-  element.innerHTML = `
-    <div class="routeLabel-vehicleIcon ${getVehicleIcon(vehicle)}"></div>
+  let content;
```
bbecquet, line 39: This works to use the JSX syntax here, but this doesn't bring much simplification in this file which is already big.
What I meant was creating a real new React component in a separate file, for example `RouteLabel.jsx`, so you can focus only on what the label renders looks like.
And here, you can call our `renderStaticReact` function to get the string representation to inject as your element `innerHTML`.

src/scss/includes/routes.scss
```diff
@@ -35,3 +35,29 @@
     }
   }
 }
+
+.publicTransportLabelItem {
```
bbecquet, line 39: Why not put that class also in the `routeLabel.scss` file?

src/scss/includes/routeLabel.scss
```diff
@@ -117,20 +121,16 @@
...
-  }
+.directions-stepByStep .routeLabel {
+  display: none;
+}
 
-  &.active .routeLabel-duration {
-    padding-top: 10px;
-  }
+@media (max-width: 640px) {
+  .routeLabel {
```
bbecquet, line 129: I think we don't want to do that for transport modes other than public transport.

bbecquet:
@xem I've pushed a commit with the label rendering in its own component, which was the intent of my first comment and is more maintainable. Let me know what you think of it, I'm interested in your feedback.
I haven't changed the visual result as it already looked nice :)

bbecquet (approved):

PR #1074, opened 2021-04-29
categories: update category list to support same set as the classifer

remi-dupre:
## Description

Update the list of categories to support the [recently added](https://github.com/Qwant/idunn/pull/217) ones that will be used when the classifier joins the party.

I specified as much icons as I could and tried to be consistent with translatable strings already defined in [`place_category_name.js`](https://github.com/Qwant/erdapfel/blob/master/local_modules/qwant-maps-common/src/place_category_name.js) to ease the translation process.

### DIFF
--- a/config/categories.yml
+++ b/config/categories.yml
@@ -75,9 +75,157 @@
 # Only categories with 'icon' are visible as action buttons.
 # Below are the categories that are only accessible
 # via queries (with detected intention) or URL
-- label: _('recycling')
-  name: recycling
-- label: _('station')
-  name: station
+- label: _('french food')
+  name: food_french
+- label: _('pizzeria')
+  name: food_pizza
+- label: _('burger')
+  name: food_burger
+- label: _('italian food')
+  name: food_italian
+- label: _('kebab')
+  name: food_kebab
+- label: _('sandwich')
+  name: food_sandwich
+- label: _('asian food')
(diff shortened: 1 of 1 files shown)


amatissart:
With the current implementation all categories with an `icon` appear in the ServicePanel.
And that's currently the only use case for these icons, so I think that they should be removed/commented from the new categories.

image

bbecquet:
@amatissart

> And that's currently the only use case for these icons

They are also used in the suggest dropdown, when it's not defined a generic icon is used.
|With icon|Without|
|---|---|
|Capture d’écran de 2021-04-29 14-51-21|Capture d’écran de 2021-04-29 14-51-37|

@remi-dupre the generic icons are enough for now and I agree in this PR we should remove the `icon` fields. (BTW, an icon without a color definition has less interest, because it will default to gray and won't respect the color scheme of the map.)

More generally it seems to me we should soon discuss how to share more efficiently these types/icons/colors/translations configs. For now there is a lot of duplication across projects. It would be great to have a single source of truth for that.

remi-dupre:
> With the current implementation all categories with an `icon` appear in the ServicePanel.

How did I manage to miss that :scream:

I indeed only checked it was updating the suggest dropdown, I'll remove this right away!

bbecquet (approved):

PR #1075, opened 2021-04-29
QMAPS-2019 Add start/end times on public transport result summaries 

bbecquet:
## Description
Use the new `start_time` and `end_time` fields provided by the API on public transport results to format and display them in the summary cards.

## Screenshots
|Before|After|
|---|---|
|Screen%20Shot%202021-04-29%20at%2011 59 03|Screen%20Shot%202021-04-29%20at%2011 56 24|

### DIFF
--- a/src/libs/time.js
+++ b/src/libs/time.js
@@ -1,0 +1,10 @@
+function getIntlLocales() {
+  const lang = window.getLang();
+  const locales = [lang.locale].concat(lang.fallback || []);
+  // Intl expects '-' in locales, such as "en-GB"
+  return locales.map(l => l.replace(/_/g, '-'));
+}
+
+export function getTimeFormatter(format = {}) {
+  return Intl.DateTimeFormat(getIntlLocales(), format);
+}
--- a/src/libs/time.js
+++ b/src/libs/time.js
@@ -9,6 +9,6 @@ export function getTimeFormatter(format = {}) {
   return Intl.DateTimeFormat(getIntlLocales(), format);
 }
 
-export function stripTimeZone(utcString) {
-  return utcString.substring(0, 19);
+export function stripTimeZone(isoString) {
+  return isoString.substring(0, 19);
 }
--- a/src/libs/time.js
+++ b/src/libs/time.js
@@ -8,3 +8,7 @@ function getIntlLocales() {
 export function getTimeFormatter(format = {}) {
   return Intl.DateTimeFormat(getIntlLocales(), format);
 }
+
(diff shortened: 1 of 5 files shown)


src/panel/direction/RouteStartEndTimes.jsx
```diff
@@ -0,0 +1,25 @@
...
+const RouteStartEndTimes = ({ start, end, className }) => {
+  if (!start || !end) {
+    return null;
+  }
+
+  const timeFormatter = getTimeFormatter({ hour: '2-digit', minute: '2-digit' });
+
+  return (
+    <div className={cx('u-bold', className)}>
+      {timeFormatter.format(new Date(start))} - {timeFormatter.format(new Date(end))}
```
amatissart, line 15: _Again that painful moment when we have to deal with timezones in Javascript :sob:_ 

This does not look correct to me. `new Date(start)` is a Date object that is no longer aware of the timezone offset returned by the server. As a result, the formatted time is implicitly converted into the user timezone, instead of the timezone localized at the start or end point. This is not what is expected, is it ?

I don't know any ideal solution using JS `Date` (see [this question](https://stackoverflow.com/q/17751698/5634396)).
In this use case though, stripping the offset part (e.g `+0300`) from the server response before calling `new Date()` could probably do the trick. The datetime will then be interpreted as being defined in the user timezone without any conversion, which is technically wrong but good enough for formatting purposes.

bbecquet: Yes sorry I forgot that (or more probably my subconscience wanted to).
Let's just strip the offset as you propose, so we'll be in the same case as the opening hours.

src/libs/time.js
```diff
@@ -0,0 +1,14 @@
...
+  // Intl expects '-' in locales, such as "en-GB"
+  return locales.map(l => l.replace(/_/g, '-'));
+}
+
+export function getTimeFormatter(format = {}) {
+  return Intl.DateTimeFormat(getIntlLocales(), format);
+}
+
+export function stripTimeZone(utcString) {
+  return utcString.substring(0, 19);
```
amatissart, line 13: nitpick: maybe `isoString` would be a more correct name for this parameter, as the datetime could actually be expressed in any timezone.

bbecquet: Indeed, and UTCString has a completely different meaning for `Date` objects so this was confusing.
I've changed it, thanks.

amatissart (approved):

PR #1077, opened 2021-04-29
QMAPS-2018 Add total walking time for public transport routes

bbecquet:
## Description
Add the total walking time in route summaries for public transport routes.

I'm not so sure about the design though… Seems to me it can be misleading to have the "Fastest route indication" so close, like it refers to this value.

Capture d’écran de 2021-04-29 19-18-40


xem (commented):
The code is OK,
But indeed the design is not optimal imo.
(at least, there should be some kind of separator or a line break between the walk time and the "fastest route" part.)

amatissart:
A minor conflict is to be solved on this branch.  
Otherwise, I am ok to merge. Adjustments on the design can be provided separately if necessary.

xem (approved):
Unless "style={{ fontSize: 11 }}" could be done better with scss, it's ok for me!

PR #1079, opened 2021-05-03
Suggest clean-up and code refacto

bbecquet:
## Description
Multiple simplifications and clean-up in the `<Suggest>` and `<SuggestDropDown>` components.

**Best reviewed commit by commit**

(Another thing I wanted to add was to entirely remove the keyboard management from SuggestDropDown to merge it with the handlers already defined in Suggest, but I still have some problems with it, will make another separate PR later :))

## Why
The `<Suggest>` component has always been a pain point and a big source of bugs. It has a complex life cycle due to a lot of specific UX stuffs and still imperative DOM manipulation of the input fied.
This PR tries to remove part of the complexity with simpler code and better isolation of side-effects.

### DIFF
--- a/src/components/ui/SuggestsDropdown.jsx
+++ b/src/components/ui/SuggestsDropdown.jsx
@@ -29,7 +29,9 @@ const SuggestsDropdown = ({ className = '', suggestItems, onSelect, onHighlight 
 
       if (key === 'ArrowUp') {
         e.preventDefault(); // prevent cursor returning at beggining
-        newHighlighted = suggestItems[suggestItems.indexOf(highlighted) - 1] || null;
+        newHighlighted = !highlighted
+          ? suggestItems[suggestItems.length - 1]
+          : suggestItems[suggestItems.indexOf(highlighted) - 1] || null;
       }
 
       setHighlighted(newHighlighted);
--- a/src/components/ui/SuggestsDropdown.jsx
+++ b/src/components/ui/SuggestsDropdown.jsx
@@ -11,37 +11,29 @@ const SuggestsDropdown = ({ className = '', suggestItems, onSelect, onHighlight 
     const keyDownHandler = e => {
       const { key } = e;
 
-      if (key === 'ArrowDown') {
-        const h = highlighted === null ? -1 : highlighted;
-
-        if (h < suggestItems.length - 1) {
-          setHighlighted(h + 1);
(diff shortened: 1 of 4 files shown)


src/components/ui/SuggestsDropdown.jsx
```diff
@@ -11,37 +11,29 @@ const SuggestsDropdown = ({ className = '', suggestItems, onSelect, onHighlight
...
 
-      if (key === 'Enter') {
-        if (highlighted !== null) {
-          e.preventDefault(); // prevent search input submit with its current content (highlighted POI name)
-          onSelect(suggestItems[highlighted]);
-        }
+      if (key === 'ArrowUp') {
+        e.preventDefault(); // prevent cursor returning at beggining
+        newHighlighted = suggestItems[suggestItems.indexOf(highlighted) - 1] || null;
       }
```
amatissart, line 33: It's no longer possible to cycle up through the list with  :arrow_up:  'ArrowUp'. Is it intentional ?

The current behavior was specified for consistency with Qwant Search, and is actually useful to access easily to the favorites that may appear at the end of the suggestions.

bbecquet: Thank you, that was a mistake, I fixed it.

xem (commented):
Nice cleanup, congrats!
I'm surprised the suggest isLoading flag is not (hasn't ever been) used... ?

bbecquet:
@xem 

> I'm surprised the suggest isLoading flag is not (hasn't ever been) used... ?

Yes, I don't remember it, wasn't there a loading indicator in the field at some point? Or the field turned to gray?
Anyway, when the operation is perceived as instantaneous it's better not to flash things in the UI :)

amatissart:
> @xem
> 
> > I'm surprised the suggest isLoading flag is not (hasn't ever been) used... ?
> 
> Yes, I don't remember it, wasn't there a loading indicator in the field at some point? Or the field turned to gray?
> Anyway, when the operation is perceived as instantaneous it's better not to flash things in the UI :)

This flag `isLoading` has been introduced in https://github.com/Qwant/erdapfel/pull/689 to display a "no result" message, and then the implementation changed in https://github.com/Qwant/erdapfel/pull/703. So yes, it's definitely obsolete.

amatissart (approved):

PR #1082, opened 2021-05-05
QMAPS-2113 use qwant's "l" laguage cookie

xem:
## Description
- Use Qwant's "l=..." cookie (if possible) as the interface language of Qwant Maps
- Fallback on the first available/useable language in the accepted-languages browser header (same as before)
- Fallback on Qwant Maps default language if nothing fits (same as before)

## Why
Better coherence with Qwant.com settings

### DIFF
--- a/bin/app.js
+++ b/bin/app.js
@@ -12,6 +12,8 @@ const getReqSerializer = require('./serializers/request');
 const cookieParser = require('cookie-parser');
 
 const app = express();
+app.use(cookieParser());
+
 const promRegistry = new promClient.Registry();
 
 function App(config) {
@@ -156,8 +158,6 @@ function App(config) {
       onerror: req.logger.error({ err }),
     })(err);
   });
-
-  app.use(cookieParser());
 }
 
 App.prototype.start = function (port) {
--- a/bin/app.js
+++ b/bin/app.js
@@ -9,6 +9,7 @@ const fakePbf = require('./middlewares/fake_pbf/index');
 const compression = require('compression');
 const mapStyle = require('./middlewares/map_style');
 const getReqSerializer = require('./serializers/request');
+const cookieParser = require('cookie-parser');
 
 const app = express();
 const promRegistry = new promClient.Registry();
@@ -155,6 +156,8 @@ function App(config) {
       onerror: req.logger.error({ err }),
     })(err);
   });
+
+  app.use(cookieParser());
 }
 
 App.prototype.start = function (port) {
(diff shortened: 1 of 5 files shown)


bin/middlewares/user_language.js
```diff
@@ -7,7 +7,14 @@ module.exports = function (languageConfig) {
   const supportedLanguages = languageConfig.supportedLanguages;
   return function (req, res, next) {
     const langHeaders = req.acceptsLanguages();
-    if (langHeaders) {
+
+    // If cookie "l" is set and corresponds to a supported language, use it
+    if (req.cookies && req.cookies.l) {
```
amatissart, line 12: It looks like `req.cookies` is always undefined. I suppose we need to enable the `cookie-parser` middleware as suggested by https://expressjs.com/en/4x/api.html#req.cookies

bin/middlewares/user_language.js
```diff
@@ -7,7 +7,14 @@ module.exports = function (languageConfig) {
...
+      res.locals.language = supportedLanguages.find(lang => lang.code === req.cookies.l);
+    }
+
+    // Else, fallback to the first accepted language in the headers
+    else if (langHeaders) {
```
amatissart, line 17: Don't you think that `accept-language` header could still be used, if `l` is not a supported language ?

xem: yes, of course, and as far as I understand, `langHeaders` represents that header. That's why I use it in my "else", as a fallback.
sorry if I made a mistake here, I don't see the problem you describe :)

bin/middlewares/user_language.js
xem: Ah indeed sorry, I was confused by the fact that I couldn't read any cookie locally, I thought it was due to my local instance of Maps not "linked" to a local instance of Phoenix.
After installing and using cookie-parser in app.js, I think we have something correct.
I couldn't manage to actually test it with a real "l=..." cookie though, as I didn't figure out yet how to set such a cookie locally...
I'd like to know more though, if you have more information on how to do that it's welcome. Thanks!

bin/app.js
```diff
@@ -155,6 +156,8 @@ function App(config) {
       onerror: req.logger.error({ err }),
     })(err);
   });
+
+  app.use(cookieParser());
```
bbecquet, line 160: As Express middlewares are run in the same order in which they are declared, this must be moved to the top, otherwise `user_language` will always find an undefined `req.cookies` object as `cookieParser` won't have run on the request.
https://stackoverflow.com/a/25210756/7001932
For this kind of generic middleware, it's a good idea to put them just after the creation of the Express instance.

bin/middlewares/user_language.js
bbecquet: @xem the problem is you will never pass in your `else if` condition if a `l` cookie was detected but the array `find` didn't match its value with a supported language.

bbecquet:
@xem
From your last commits, it looks like your prettier config doesn't work anymore :/
It should be entirely automatic, saving you from the pain of doing that by hand. We'll find how to restore it.

bbecquet (approved):

PR #1083, opened 2021-05-11
QMAPS-1926 Allow scroll in suggest on iOS

bbecquet:
## Description
Make it possible to scroll suggest results on iOS.
Previously, the whole page content scrolled. This was due to the virtual keyboard on iOS browsers not changing the viewport height. As a result, the list of suggestions fit entirely on the screen height (even if half of it was behind the keyboard) so it wasn't scrollable.
Here, we use a recent web API, [Visual Viewport](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API), which is supported in Safari 13+ and gives access to the real visible height, without the space taken by the keyboard. We use it to give a fixed height to the suggest list, so it becomes scrollable.

We also have to manage the case of an empty list or a list too small to be scrollable. In that case, the dropdown background element will scroll the page like before. So there is a special test for that to detect if the suggestions overflow their container or not. If not, we simply cancel any `touchmove` action.

|Before|After|
|---|---|
|https://user-images.githubusercontent.com/243653/119379835-41e1ec80-bcc0-11eb-888d-1c59b814256f.MOV|https://user-images.githubusercontent.com/243653/119379907-532af900-bcc0-11eb-9a70-ebfedb19da74.MOV|

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -109,8 +109,20 @@ const Suggest = ({
       // Giving a fixed height to the container makes the content scrollable
       outputNode.style.height = window.visualViewport.height - TOP_BAR_HEIGHT + 'px';
 
+      const suggestions = document.querySelector('.autocomplete_suggestions');
+      const cancelTouchScrollIfNotOverflow = e => {
+        const hasOverflow =
+          suggestions &&
+          suggestions.getBoundingClientRect().height > outputNode.getBoundingClientRect().height;
+        if (!hasOverflow) {
+          e.preventDefault();
+        }
+      };
+      outputNode.addEventListener('touchmove', cancelTouchScrollIfNotOverflow);
+
       return () => {
         outputNode.style.height = 'auto';
+        outputNode.removeEventListener('touchmove', cancelTouchScrollIfNotOverflow);
       };
     }
   }, [isMobile, items, dropdownVisible, outputNode]);
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState, useCallback } from 'react';
 import ReactDOM from 'react-dom';
 import debounce from 'lodash.debounce';
 import { bool, string, func, object } from 'prop-types';
-
+import { useDevice } from 'src/hooks';
 import SuggestsDropdown from 'src/components/ui/SuggestsDropdown';
 import { fetchSuggests, getInputValue, modifyList } from 'src/libs/suggest';
 
@@ -25,6 +25,7 @@ const Suggest = ({
   const [highlighted, setHighlighted] = useState(null);
   const [hasFocus, setHasFocus] = useState(false);
   const dropdownVisible = hasFocus && isOpen && outputNode;
+  const { isMobile } = useDevice();
 
   const close = () => {
     setIsOpen(false);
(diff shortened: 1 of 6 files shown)


bbecquet:
Little update.
I think I've got everything working in a local PoC. By working, I mean:
 - when there are scrollable results, they can be scrolled without scrolling the whole body
 - when there are no results or a result list smaller than the overflow threshold, touching them don't scroll the whole body
 - touching the search bar doesn't scroll the body either

Actually, this also fixes a long standing bug on small non-iOS devices, where the last items of the result lists couldn't be reached, hidden by the keyboard.

The thing is the implementation is a bit messy for now. And if we build on top of the current master, it will stay that way, with global selectors and such. I really think finishing the [React conversion of the TopBar](https://github.com/Qwant/erdapfel/pull/997) before that would make things easier and cleaner.

xem (dismissed):
I approve the code.
Sadly, even on Browserstack + local testing, I can't manage to test it on iOS.
Let's test with a real iphone once it's in plive, ok?

src/components/ui/Suggest.jsx
```diff
@@ -101,6 +102,31 @@ const Suggest = ({
     }
   };
 
+  useEffect(() => {
+    if (isMobile && dropdownVisible && window.visualViewport) {
```
amatissart, line 106: Note that `window.visualViewport` is not available by default on Firefox for desktop, including in the responsive view. 
(the flag `dom.visualviewport.enabled` can be enabled via "about:config").  
So it could make testing a bit confusing, especially for a newcomer. Do you think it could be documented somewhere?

src/components/ui/Suggest.jsx
```diff
@@ -101,6 +102,31 @@ const Suggest = ({
...
+        }
+      };
+      outputNode.addEventListener('touchmove', cancelTouchScrollIfNotOverflow);
+
+      return () => {
+        outputNode.style.height = 'auto';
+        outputNode.removeEventListener('touchmove', cancelTouchScrollIfNotOverflow);
+      };
+    }
+  }, [isMobile, items, dropdownVisible, outputNode]);
```
amatissart, line 128: Maybe this effect could also listen to [the visualViewport `resize` event](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API#interfaces), to adapt the element height after the keyboard is closed.
This is quite minor though.

src/scss/includes/direction-panel.scss
```diff
@@ -36,6 +36,8 @@
       position: fixed;
       top: 64px;
       width: 100vw;
+      background-color: $background;
+      overflow-y: scroll;
```
amatissart, line 40: Out of curiosity, why is this explicit "scroll" required, instead of "auto" ?

bbecquet: You're right it's not, I guess I got heavy-handed while battling with Safari :) I'll put `auto` instead.

src/components/ui/Suggest.jsx
bbecquet: I've added it, as it was less difficult than what I thought at first.
Let's hope I didn't break anything else :P

src/components/ui/Suggest.jsx
bbecquet: Good idea, I've added a link to the API doc.

amatissart (approved):

PR #1084, opened 2021-05-12
QMAPS-1589 map UI style changes + mobile fade-out for the scale

xem:
## Description

1) change style of map control buttons -margins, colors, backgrounds, icons, text, etc (still in discussion with Design):

image
image

2) make scale disappear on mobile after 2s of no map interaction (click, long touch, move, zoom)
fadeout1

### DIFF
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -474,8 +474,7 @@ Scene.prototype.moveMobileBottomUI = function (bottom = 0) {
     return;
   }
   const uiControls = [
-    '.mapboxgl-ctrl-attrib',
-    '.map_control__scale',
+    '.map_control__scale_attribute_container',
     '.mapboxgl-ctrl-geolocate',
     '.direction_shortcut',
   ];
--- a/src/adapters/scene.js
+++ b/src/adapters/scene.js
@@ -38,6 +38,20 @@ const getPoiView = poi => ({
   bounds: poi.geometry.bbox,
 });
 
+const hideMobileScale = function () {
+  const item = document.querySelector('.map_control__scale');
+  if (item) {
+    item.classList.add('fadeOut');
+  }
+};
+
+const showMobileScale = function () {
+  const item = document.querySelector('.map_control__scale');
+  if (item) {
+    item.classList.remove('fadeOut');
+  }
+};
+
 Scene.prototype.getMapInitOptions = function ({ locationHash, bbox }) {
   if (bbox) {
     try {
@@ -120,11 +134,6 @@ Scene.prototype.initMapBox = function ({ locationHash, bbox }) {
   this.lastDoubleTapTimeStamp = 0;
   this.lastTouchEndTimeStamp = 0;
   this.mb.on('touchend', _e => {
-    // Start a new 2s idle timeout
-    MOBILE_IDLE_TIMEOUT = setTimeout(() => {
-      this.hideMobileScale();
-    }, MOBILE_IDLE_DELAY_MS);
-
     const timeStamp = Date.now();
     // maybe we should also check the distance between the two touch events…
     if (timeStamp - this.lastTouchEndTimeStamp < this.DOUBLE_TAP_DELAY_MS) {
@@ -134,11 +143,7 @@ Scene.prototype.initMapBox = function ({ locationHash, bbox }) {
   });
 
   this.mb.on('load', () => {
-    // Start 2s idle timeout
-    MOBILE_IDLE_TIMEOUT = setTimeout(() => {
-      this.hideMobileScale();
-    }, MOBILE_IDLE_DELAY_MS);
-
+    fire('restart_idle_timeout');
     this.onHashChange();
     new SceneDirection(this.mb);
     new SceneCategory(this.mb);
@@ -175,15 +180,7 @@ Scene.prototype.initMapBox = function ({ locationHash, bbox }) {
     this.clickDelayHandler = null;
 
     this.mb.on('click', e => {
-      // Cancel idle status
-      this.showMobileScale();
-      clearTimeout(MOBILE_IDLE_TIMEOUT);
-
-      // Start a new 2s idle timeout
-      MOBILE_IDLE_TIMEOUT = setTimeout(() => {
-        this.hideMobileScale();
-      }, MOBILE_IDLE_DELAY_MS);
-
+      fire('restart_idle_timeout');
       if (e.originalEvent.cancelBubble) {
         return;
       }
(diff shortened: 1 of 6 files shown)


src/scss/includes/map_theme.scss
```diff
@@ -116,16 +129,67 @@
...
+      &.mapboxgl-compact-show {
+        width: calc(100vw - 95px);
+      }
 
       .mapboxgl-ctrl-attrib-button {
         top: auto;
         right: auto;
         bottom: 0;
         left: 0;
+        background: url('../images/feather/info.svg') no-repeat center center rgba(255,255,255,0.8);
```
bbecquet, line 192: You forgot to commit this new icon, the attribution circle is empty for me.

src/adapters/scene.js
```diff
@@ -118,6 +120,11 @@ Scene.prototype.initMapBox = function ({ locationHash, bbox }) {
   this.lastDoubleTapTimeStamp = 0;
   this.lastTouchEndTimeStamp = 0;
   this.mb.on('touchend', _e => {
+    // Start a new 2s idle timeout
+    MOBILE_IDLE_TIMEOUT = setTimeout(() => {
```
bbecquet, line 124: Instead of having to duplicate that everywhere and separate the timeout creation and clear, a better way would be to listen to the existing `map_user_interaction` event, which is already used for something similar (maybe with more conditions to fire it).
That way, everything would be in the same place, with no duplication and with a straightforward algo: on *any* map interaction, show the scale and start the timeout.

src/scss/includes/menu.scss
```diff
@@ -136,8 +136,8 @@ $productDrawerWidth: 744px;
 
     &-apps {
       position: absolute;
-      bottom: $spacing-m; 
-      right: $spacing-l; 
+      bottom: $spacing-m;
```
bbecquet, line 139: Are changes in this file related to the scale stuff? Looks like a git add mistake.

src/scss/includes/map_theme.scss
```diff
@@ -116,16 +129,67 @@
...
+        border-radius: 10px;
+        height: 20px;
+        width: calc(100% + 20px);
+        left: -8px;
+        top: -3px;
+        z-index: -1;
+        position: absolute;
+      }
+
+      &:after {
```
bbecquet, line 165: This is a lot of CSS to get this looks. I see why the `:before` pseudo may be useful, but is there something which prevents keeping the same structure as the desktop control, using a border directly on the element instead of an `:after`?

bbecquet:
Another remark. In certain cases I noticed the scale disappeared while I was moving the map. It seems to happen when you move the map, you lift your finger for a short time, and then resume the panning. A timeout probably isn't cleared properly. Putting everything in the same place whould ease debugging those cases.

src/adapters/scene.js
xem: The way I see it:

the timer should start when the map has loaded (in scene.js)

It should stop / reset when we interact with the map (touch, pan, click, long click, zoom)
(we could put that in map_user_interaction, which is in PanelManager.jsx)

It should restart when the map is not interacted with anymore (touchend, moveend, ...)

How would you proceed to share the same timer between these two components?
Or would you simply put a simple "stop-reset-restart" in map_user_interaction, and nothing else?

src/scss/includes/menu.scss
xem: it seems like it's my linter that removed extra spaces at the end of each line. should I revert them?

bbecquet: Sorry I hadn't seen it was just space removal on these two lines, but about the spacing change below and the added linebreaks? Looks like this file should be just removed from this PR.

src/adapters/scene.js
bbecquet: What I propose:
- Remove all these `setTimeout` in all these different event
- Factorize them in a single one, in a listener to `map_user_interaction`, which is already thrown on pans and zooms, maybe it needs to be thrown on more map events.
- Put the listener where you want. The one in `PanelManager` deals with the panel minimization, so it's in its right place, but here we want to deal with map controls, it's probably better to define a second listener for it in `scene.js` or somewhere else.
- In the listener, I think it should be enough to clear the timeout each time we receive the event and start a new one

src/scss/includes/menu.scss
xem: I'd rather keep it. In this ticket I moved the scale, attribution, zoom elements according to the new Design, and they're now offseted 16px from the right of the screen. If the menu button stays offseted 12px (spacing-s), it looks bad for me: 
image

src/scss/includes/map_theme.scss
xem: sadly no, I didn't find any other solution than this :after trick to keep the scale element with black straight borders, and have a rounded background behind it. On desktop, the rounded background is applied to the parent of the scale and the attribution, but that's not possible here, as I want to make the scale disappear while the (i) button stays visible.

src/scss/includes/menu.scss
bbecquet: It is the point of shifting the zoom/geoloc slightly to the left: now they centered in the same column as the menu button.
118653498-e14a3f80-b7e7-11eb-832a-b69700410788

In any case, if it's not an explicit part of the issue about the scale, don't add this change here.

src/adapters/scene.js
xem: I did not fire map_user_interaction on other events than the ones that already do it because map_user_interaction  minimizes the current panel, and for the other events (click, long touch) we do NOT want the current panel to minimize.

So instead, in these events, I'll fire directly my new event: restart_idle_timeout, whose listener is placed in scene.js

I hope it's ok for you!

bbecquet: Alright, as long as the timeout code isn't duplicated it's good.

src/scss/includes/map_theme.scss
```diff
@@ -214,8 +270,13 @@
...
   }
 
-  .mapboxgl-ctrl-bottom-right .mapboxgl-ctrl, .mapboxgl-ctrl.mapboxgl-ctrl-attrib {
+  .mapboxgl-ctrl-bottom-right .mapboxgl-ctrl {
     float: inherit;
+    margin-right: 10px;
+  }
+
+  .mapboxgl-ctrl-bottom-right  {
+    height: 55px;
```
bbecquet, line 279: Is this really required? This doesn't seem to change much. Our Mapbox control CSS overrides are already a mess, far too complicated, far too many lines for so little user value, let's try no to add anything new.

bbecquet (commented):
We forgot about the compass control on desktop, which is shifted because of the changed margin for everything (whyyyyy? :cry:).
Capture d’écran de 2021-05-20 14-44-26

It's not too late to revert the margin change on desktop for controls other than scale, which was supposed to be the only change…

PR #1086, opened 2021-05-20
Convert PanelManager to a functional component

bbecquet:
## Description
Following the [React implementation of the top bar](https://github.com/Qwant/erdapfel/pull/997) and while it's still fresh, I converted the PanelManager to a functional component.

## Why
- Clarifying the side effects and state transitions
- Use our hooks `useDevice` and `useConfig`
- Good occasion to simplify some old or recent things

### DIFF
--- a/src/panel/PanelManager.jsx
+++ b/src/panel/PanelManager.jsx
@@ -177,26 +177,16 @@ const PanelManager = ({ router }) => {
 
   const backToList = (e, poiFilters) => {
     e.stopPropagation();
-    const queryObject = {};
-    const mappingParams = {
-      query: 'q',
-      category: 'type',
+    const { query, category, ...rest } = poiFilters;
+    const queryObject = {
+      q: query,
+      type: category,
+      ...rest,
     };
-
-    for (const name in poiFilters) {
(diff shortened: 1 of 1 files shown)


src/panel/PanelManager.jsx
```diff
@@ -9,128 +8,81 @@ import DirectionPanel from 'src/panel/direction/DirectionPanel';
...
-    };
-
-    this.mainSearchInputRef = React.createRef();
+function getTopBarAppValue({ poiFilters = {}, poi = {}, query } = {}) {
+  if (poi.name) {
+    return poi.name;
+  }
+  if (poiFilters.category) {
+    return CategoryService.getCategoryByName(poiFilters.category)?.getInputValue();
   }
```
amatissart, line 24: What if the category is not found ?  
For consistency, I think that an empty string should be used as a default here too, similarly to the next line.
(eg: http://localhost:3000/places/?type=invalidType could trigger a query to "undefined" after the input is focused)

bbecquet: Sure. I've made the change, thanks.

amatissart (approved):
Looks good to me.

PR #1092, opened 2021-05-24
Qmaps 1734 suggest errors

xem:
## Description
Use new, centered error message
- in suggest
- when pressing enter after a search with no result
- when coming from an url with no result

Not done in this ticket:
- itinerary errors (route not found / public transport)
- links to a page explaining how to add places on the map
- handle clicks on error suggest items to take the link into account

## Why
Design

## Screenshots

image

### DIFF
--- a/src/panel/NoResultMessage.jsx
+++ b/src/panel/NoResultMessage.jsx
@@ -5,11 +5,11 @@ const NoResultMessage = () => {
   return (
     <>
       <p className="u-center u-mb-xs u-text--smallTitle">
-        {_('Sorry, we could not find this place 🏝️', 'suggest')}
+        {_('Sorry, we could not find this place🏝️', 'suggest')}
       </p>
-      <p className="u-center u-text--subtitle u-mb-l">
+      <p className="u-center u-text--subtitle">
         {_(
-          'Please try to correct your query or rewrite it with more details about the location (city, country, ...)',
+          'Please try to correct your query or rewrite it with more details about the location (city, country, …)',
           'suggest'
         )}
       </p>
--- a/src/panel/NoResultMessage.jsx
+++ b/src/panel/NoResultMessage.jsx
@@ -1,0 +1,20 @@
+/* globals _ */
+import React from 'react';
+
+const NoResultMessage = () => {
+  return (
+    <>
+      <p className="u-mb-xs u-text--smallTitle">
+        {_('Sorry, we could not find this place 🏝', 'suggest')}
+      </p>
+      <p className="u-text--subtitle u-mb-l">
+        {_(
+          'Please try to correct your query or rewrite it with more details about the location (city, country, ...)',
+          'suggest'
+        )}
+      </p>
+    </>
+  );
+};
+
+export default NoResultMessage;
--- a/src/panel/NoResultMessage.jsx
+++ b/src/panel/NoResultMessage.jsx
@@ -5,7 +5,7 @@ const NoResultMessage = () => {
   return (
     <>
       <p className="u-mb-xs u-text--smallTitle">
-        {_('Sorry, we could not find this place 🏝', 'suggest')}
+        {_('Sorry, we could not find this place 🏝️', 'suggest')}
       </p>
       <p className="u-text--subtitle u-mb-l">
         {_(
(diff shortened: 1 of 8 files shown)


src/panel/NoResultPanel.jsx
```diff
@@ -1,26 +1,18 @@
...
-
+const NoResultPanel = () => {
   return (
     <Panel close={close}>
       <div style={{ padding: '16px 16px 32px 16px' }}>
-        <p className="u-mb-xs u-text--smallTitle">{_('No results found.')}</p>
-        <p className="u-text--subtitle u-mb-l">
-          {_('Check the spelling of your search or add more details, such as city or country.')}
+        <NoResultMessage />
+        <p className="u-center">
```
bbecquet, line 13: As said before on Mattermost, let's not include that link in this PR, as we don't even know when it will be available in prod.
It will be very easy to add later.

src/panel/NoResultMessage.jsx
```diff
@@ -0,0 +1,20 @@
+/* globals _ */
+import React from 'react';
+
+const NoResultMessage = () => {
+  return (
+    <>
+      <p className="u-center u-mb-xs u-text--smallTitle">
+        {_('Sorry, we could not find this place 🏝️', 'suggest')}
+      </p>
+      <p className="u-center u-text--subtitle u-mb-l">
```
bbecquet, line 10: I think we don't need this bottom margin until we display something under the message. It looks a bit odd, I would remove the `u-mb-l` entirely.

src/panel/NoResultMessage.jsx
```diff
@@ -0,0 +1,20 @@
...
+        {_(
+          'Please try to correct your query or rewrite it with more details about the location (city, country, ...)',
```
bbecquet, line 12: Sorry for being such a typo-nerd, but let's use the real ellipsis character `…` here instead of three dots.

src/panel/NoResultPanel.jsx
```diff
@@ -1,26 +1,14 @@
...
-
+const NoResultPanel = () => {
   return (
     <Panel close={close}>
       <div style={{ padding: '16px 16px 32px 16px' }}>
-        <p className="u-mb-xs u-text--smallTitle">{_('No results found.')}</p>
-        <p className="u-text--subtitle u-mb-l">
-          {_('Check the spelling of your search or add more details, such as city or country.')}
-        </p>
-        <a onClick={tryNewSearch} href="#">
```
bbecquet, line 21: Is is ok to just remove this action before we have something else to replace it?
Maybe we should keep it the meantime (for this case only, not during the suggest).

Or, if it's ok to get rid of it already, it deserves some cleaning in `PanelManager` to remove the `resetInput` function passed to `NoResultPanel` (if you do that, rebase your PR first as `PanelManager` changed a lot recently)

src/panel/NoResultMessage.jsx
```diff
@@ -0,0 +1,20 @@
+/* globals _ */
+import React from 'react';
+
+const NoResultMessage = () => {
+  return (
+    <>
+      <p className="u-center u-mb-xs u-text--smallTitle">
+        {_('Sorry, we could not find this place🏝️', 'suggest')}
```
amatissart, line 8: I would add a space before the emoji, for consistency with #1093

src/panel/NoResultPanel.jsx
```diff
@@ -14,13 +15,12 @@ const NoResultPanel = ({ resetInput }) => {
...
     <Panel close={close}>
       <div style={{ padding: '16px 16px 32px 16px' }}>
-        <p className="u-mb-xs u-text--smallTitle">{_('No results found.')}</p>
-        <p className="u-text--subtitle u-mb-l">
-          {_('Check the spelling of your search or add more details, such as city or country.')}
+        <NoResultMessage />
+        <p className="u-center">
+          <a onClick={tryNewSearch} href="#">
+            {_('Try a new search query')}
+          </a>
```
amatissart, line 22: The margin above the link has been lost here
image

amatissart (commented):
Very minor styling comments

src/panel/NoResultPanel.jsx
xem: I adjusted the spaces around the message, and above the link according to the designs

bbecquet (approved):

PR #1096, opened 2021-05-27
Qmaps 2145 poi phone number

bbecquet:
## Description
- redesign the POI `Block` component to be closer to the designs https://app.zeplin.io/project/5a99616161f517c8b8591627/screen/5f05e9b752e3b96eba4519dc: full-width, all area clickable, gray background on hover.
- adapt some blocks to the new component: `Website`, `Contact` and `Phone`
- change the `Phone` component so it hides the phone number on desktop for PJ POIs

### DIFF
--- a/src/panel/poi/blocks/Phone.jsx
+++ b/src/panel/poi/blocks/Phone.jsx
@@ -8,7 +8,7 @@ import { useDevice, useI18n } from 'src/hooks';
 const PhoneBlock = ({ block, poi }) => {
   const { _ } = useI18n();
   const { isMobile } = useDevice();
-  const [isHidden, setHidden] = useState(true);
+  const [isHidden, setHidden] = useState(!isMobile && isFromPagesJaunes(poi));
 
   useEffect(() => {
     setHidden(!isMobile && isFromPagesJaunes(poi));
--- a/src/panel/poi/blocks/Phone.jsx
+++ b/src/panel/poi/blocks/Phone.jsx
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import Block from 'src/panel/poi/blocks/Block';
+import Telemetry from 'src/libs/telemetry';
 import { isFromPagesJaunes } from 'src/libs/pois';
 import { useDevice, useI18n } from 'src/hooks';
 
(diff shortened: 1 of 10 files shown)


xem (changes requested):
Good job!
Everything seems perfect, except the phone number that has become a blue link, while it's black on Zeplin.
image
I'm okay with the blue link though, so I can approve if that's fine for you too

bbecquet:
@xem thanks for the review :+1:
You're right about the link color, I didn't try to force it black yet. I think they will be some new details about this feature during the next planning, so I let some stuff aside, focusing on the `Block` changes first.

xem (approved):

bbecquet:
@xem Since your review, I've updated this PR to add two things:
 - on desktop, instead of a link, propose a "copy to clipboard" action when the number is visible (see some old designs for the ref)
 - when clicking the number on mobile or revealing it on desktop, trigger the same telemetry event as the blue rounded CTA

src/panel/poi/blocks/Phone.jsx
```diff
@@ -1,16 +1,65 @@
...
+    ? {
+        onClick: () => {
+          setHidden(false);
+          sendEvent();
+        },
+      }
+    : {
+        onClick: () => copyToClipboard(block.local_format),
+        actionIcon: <IconFileCopy width={20} />,
+        tooltip: _('Copy'),
```
amatissart, line 48: I am not sure about this change. This action was already doable manually, whereas the standard "href" link is now lost.   
On most systems, it can be used to forward the number to a suitable application, using the international format that is now completely inaccessible.

xem: how about a "copy" button + a "call" button aside to the number?

bbecquet: The designer's idea is that every "block" should have a **unique** and **contextual** action which is the more useful and logical for the user on the current device.
For desktop it's been changed to "copy to clipboard". It may seem like we lost something, but actually there is already a "call" CTA clearly visible in the button bar, and phone/desktop integration is not that common, so I see the point.

bbecquet: Ok, it was decided "IRL" to discuss this copy-to-clipboard action later separately, and keep the block as a link to the `tel:` url even on desktop for now, so I reverted some of the latest commits.

amatissart (approved):

PR #1102, opened 2021-05-31
Change error message when no matching point in area

bbecquet:
## Description
New wording and action for the error message when there is no matching point in the area for a "category" result.
 
## Screenshots
|Before|After|
|---|---|
|Screen%20Shot%202021-05-31%20at%2015 25 36|Screen%20Shot%202021-05-31%20at%2015 37 36|

### DIFF
--- a/src/panel/category/CategoryPanelError.jsx
+++ b/src/panel/category/CategoryPanelError.jsx
@@ -24,7 +24,7 @@ const CategoryPanelError = ({ zoomIn }) => {
         ),
       }
     : {
-        title: _("Hmm, looks like a no-man's land 🙅‍♂️", 'categories'),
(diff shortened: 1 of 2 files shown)


xem (approved):
Looks much better now!
I approve but, any idea why the panel is so tall on mobile?
image

(it's not a regression coming from this branch, but I wonder if it should be fixed... by making the panel as high as its content)

bbecquet:
@xem
Ah, yes we could. In theory it's just a matter of passing the right `fitContent` property to the `Panel` component in `CategoryPanel`, and make sure its reset when there is no more error to ensure the normal behavior is kept.
This dynamic panel sizing is one of the fragile parts of the app, which used to give a lot of bugs, so there may be some traps :) You could make a dedicated PR for that if you want.

PR #1104, opened 2021-06-01
QMAPS-2159 fix error messages height on mobile

xem:
## Description
error panels on mobile were too tall
now they're as tall as the text
NB: this fix shouldn't impact the display of these panels when there's no error

## Screenshots
image
image

### DIFF
--- a/src/panel/NoResultPanel.jsx
+++ b/src/panel/NoResultPanel.jsx
@@ -13,7 +13,7 @@ const NoResultPanel = ({ resetInput }) => {
   };
 
   return (
-    <Panel close={close}>
+    <Panel close={close} fitContent="default">
       <div style={{ padding: '20px 34px' }}>
         <NoResultMessage />
         <p className="u-center u-mt-s">
--- a/src/panel/NoResultPanel.jsx
+++ b/src/panel/NoResultPanel.jsx
@@ -13,7 +13,7 @@ const NoResultPanel = ({ resetInput }) => {
   };
 
   return (
-    <Panel close={close} fitContent="default">
+    <Panel close={close} fitContent={['default']}>
       <div style={{ padding: '20px 34px' }}>
         <NoResultMessage />
         <p className="u-center u-mt-s">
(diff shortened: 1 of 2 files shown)


src/panel/NoResultPanel.jsx
```diff
@@ -13,7 +13,7 @@ const NoResultPanel = ({ resetInput }) => {
   };
 
   return (
-    <Panel close={close}>
+    <Panel close={close} fitContent="default">
```
bbecquet, line 16: As declared in the `Panel` proptypes, `fitContent` should be an array. It works here because you're lucky `indexOf` exists on strings too, but it may break later with just a string.

src/panel/category/CategoryPanel.jsx
```diff
@@ -165,6 +165,7 @@ const CategoryPanel = ({ poiFilters = {}, bbox }) => {
       floatingItemsLeft={[
         isMobile && shouldShowBackToQwant() && <BackToQwantButton key="back-to-qwant" isMobile />,
       ]}
+      fitContent={!pois || pois.length === 0 ? 'default' : 'minimized'}
```
bbecquet, line 168: Same thing here, it should be an array.

Moreover, you must not pass `'default'` in the non-error case, as it breaks the normal behavior. You should pass an empty array instead, which tells `Panel` to use static sizes instead for all panel sizes.

xem: thanks for the explanations, I hadn't use this mechanism until today and it wasn't super clear for me how to use it properly.
Fixed it, hope it's ok now. It seems to work nice according to my tests :)

bbecquet:
@xem

> thanks for the explanations, I hadn't use this mechanism until today and it wasn't super clear for me how to use it properly.

You're right, this feature is powerful but confuse and hard to discover. But it's great when our existing components can be used as-is to bring enhancements like this :+1:

bbecquet (approved):

PR #1107, opened 2021-06-02
QMAPS-2155 add user feedback in itinerary list

xem:
## Description
Add feedback form at the bottom of itinerary lists, only if there are results
This is a beginning of implementation, as feedback's design and behavior (especially on mobile) need more work.
See QMAPS-2154

## Screenshots
image


bbecquet (approved):

bbecquet:
Damn it, I approved that too quickly (and merged as I needed it for testing design changes).
Despite the description, it is always displayed, even when there are no route results… I'll fix that in another PR -_-

PR #1108, opened 2021-06-02
Qmaps 2153 suggest feedback

xem:
## Description
Add feedback form at the bottom of suggest (main + itinerary), only if there are results, and only if something has been typed in the field
This is a beginning of implementation, as feedback's design and behavior (especially on mobile) need more work.
See QMAPS-2154

## Screenshots
image


src/components/ui/SuggestsDropdown.jsx
```diff
@@ -1,24 +1,31 @@
...
+            onMouseDown={() => {
+              onSelect(suggestItem);
+            }}
+            className={classnames({ selected: highlighted === suggestItem })}
+          >
+            <SuggestItem item={suggestItem} />
+          </li>
+        ))}
+      </ul>
+      {value.length > 0 && suggestItems.length > 0 && !suggestItems[0].errorLabel && (
```
bbecquet, line 25: It's safer to test only `value` for truthiness instead of its length, otherwise it will crash if provided with `null` or `undefined`.

bbecquet (commented):
Now that the functional bits are merged, we could complete this PR like https://github.com/Qwant/erdapfel/pull/1113 so the whole feature is done in one round. Here are the missing bits:
 - pass a `questionId` prop so the event logs and local storage work
 - pass a `context` string prop with the query and the map position
 - display only in `maximized` panel size
 
 And as we discussed, let's not display the feedback component in the itinerary autocompletes.

xem:
I'm on it!

xem (commented):
Hi @bbecquet @amatissart,
Finishing the feedback in suggest was a bit tricky, so I made the  following choices, I hope it's ok for you (feel free to make changes tomorrow if not):
- add a value prop and a useFeedback prop (as discussed on Slack) to enable the feedback only on the main field and not in suggest fields, and to enable it only if the field's value is not empty
- use onMouseDown events (instead of onClick) on the yes/no/x buttons in order to be able to react to the feedback clicks before the suggest list closes itself
- The context passed to UserFeedbackYesNo is encodeURIComponent(value) + document.location.hash. I didn't change the context passed to the 2 other feedback forms (list / routes). Should their context be built like this one?
- I chose to put the event.stopPropagation() directly in UserFeedbackQuestion (in the three buttons' "onmousemove"). It prevents the suggest from closing when we click each button. Should this line be moved inside the callback?

bbecquet:
@xem Nice work with the tricky mouse stuff :)
Ok with the approach taken here, as it's consistent with what we did elsewhere and it works fine. I just removed the `stopPropagation` as `preventDefault` is enough.

(After it's merged I think I'll take so time to change how the `blur` is managed on the suggest, so we don't need this `mouseDown` hack anymore. At first it was limited to a single place, now it's beginning to be everywhere.)

About the context, I'm ok with that too, as it's what was decided before. No need to change the other feedbacks context, its shape will have to be interpreted specifically for each anyway. There may be changes to do on that later for more robustness, but it's too early to tell.

xem:
Great. Ok for your fix, feel free to merge :)

bbecquet:
@xem Ah, one more thing: the feedback block in the suggest on iOS doesn't behave correctly with the scroll rules introduced in https://github.com/Qwant/erdapfel/pull/1083.
As I have the iPhone I'll add another commit for that.

bbecquet (approved):

PR #1111, opened 2021-06-03
Feedback success

bbecquet:
## Description
Define a new `Alert` component (based on the Design System implementation in search) and use it to display a temporary success message after a Feedback is submitted.

*Note: the diff is a bit large because there was already an `Alert` component in Maps, actually matching the `Notification` component of the Design System, so I renamed it in the first commit so we stay consistent with the DS components.*

## Screenshots

https://user-images.githubusercontent.com/243653/120677152-15378d00-c497-11eb-9717-cd116d8de05b.mp4


xem (changes requested):
Code + local testing is OK for me!
Though, the message disappears after 3s, which is very fast.
I can barely read all the message, and a visually impaired user may find that too fast. How about something between 5-10s?

bbecquet:
@xem

> Though, the message disappears after 3s, which is very fast.

I re-used the same value as the component from the design system.
The point of such a short notice, as noted in the DS source code, is that success messages are "unimportant", so it's ok to make them disappear after a short time. The main feedback is that the user sees a green stuff so they understand their action was correctly taken into account, and they can go on with their normal use without being interrupted too long in their flow.
If the message was really important we would use a component which doesn't disappear on its own anyway. And a component which disappears very long after it appeared is too confusing and disturbing.

That said, I agree this may not ideal from an accessibility point of view, but IMO the msg delay isn't the main problem here, the markup isn't optimized to signal clearly something has changed as result to a user action. We could discuss that more globaly with the designers.

xem (approved):

PR #1112, opened 2021-06-04
QMAPS-2157 feedback remember close

bbecquet:
## Description
Implement the mechanism to keep a user feedback question hidden if the user dismissed it by clicking on the close button.
 1. when a question is answered or dismissed, store the action (yes|no|dismiss) and the current datetime in localstorage
 2. use a localstorage key based on the `questionId` so all questions across the app are independant, and dismissing one will not hide the others 
 3. before rendering the question, check if there is a stored value
  - if not, means the user previously did nothing with this question => display the question
  - if there is a value but which isn't `dismiss`, it means the user already answered this question but probably in a different context (query, route, etc.) => display the question
  - if the user previously dismissed the question, check if enough days have passed. If not, don't display the question.

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -93,3 +93,4 @@ covid19:
 
 userFeedback:
   enabled: false
+  dismissDurationDays: 30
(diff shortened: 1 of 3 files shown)


bbecquet:
Closed because the branch had the bad ticket number, re-opened as https://github.com/Qwant/erdapfel/pull/1114

PR #1114, opened 2021-06-04
QMAPS-2162 Hide previously dismissed question during some days

bbecquet:
## Description
Implement the mechanism to keep a user feedback question hidden if the user dismissed it by clicking on the close button.
 1. when a question is answered or dismissed, store the action (yes|no|dismiss) and the current datetime in localstorage
 2. use a localstorage key based on the `questionId` so all questions across the app are independant, and dismissing one will not hide the others 
 3. before rendering the question, check if there is a stored value
  - if not, means the user previously did nothing with this question => display the question
  - if there is a value but which isn't `dismiss`, it means the user already answered this question but probably in a different context (query, route, etc.) => display the question
  - if the user previously dismissed the question, check if enough days have passed. If not, don't display the question.

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -93,3 +93,4 @@ covid19:
 
 userFeedback:
   enabled: false
+  dismissDurationDays: 15
--- a/src/components/ui/UserFeedbackYesNo.jsx
+++ b/src/components/ui/UserFeedbackYesNo.jsx
@@ -4,27 +4,28 @@ import { UserFeedbackQuestion, Alert } from './index';
 import { useConfig, useI18n } from 'src/hooks';
 import { IconThumbUp, IconThumbDown } from './icons';
 import { ACTION_BLUE_BASE } from 'src/libs/colors';
-import { sendAnswer } from 'src/libs/userFeedback';
+import { sendAnswer, rememberAnswer, shouldBeDisplayed } from 'src/libs/userFeedback';
 
 const UserFeedbackYesNo = ({ questionId, context, question }) => {
-  const { enabled: userFeedBackEnabled } = useConfig('userFeedback');
+  const { enabled: userFeedBackEnabled, dismissDurationDays } = useConfig('userFeedback');
   const { _ } = useI18n();
   const [isAnswered, setAnswered] = useState(false);
   const [isDismissed, setDismissed] = useState(false);
+  const display = shouldBeDisplayed(questionId, dismissDurationDays);
 
   const closeQuestion = () => {
     sendAnswer(questionId, 'dismiss', { context });
+    rememberAnswer(questionId, 'dismiss');
     setDismissed(true);
   };
 
   const onAnswer = answer => () => {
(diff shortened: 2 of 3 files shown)


xem (changes requested):
I detect many issues when testing this branch, but maybe it'll be fixed after the rebase?
- on mobile (only), after an itinerary search, when I click on on yes / no / x, the feedback form blinks, reappears, and nothing is written in localStorage
- on PC, in the itinerary feedback, in my localStorage the key's name is: "qmaps_v1_userFeedback_undefined" (missing questionId I guess?)
- Still in the itinerary feedback, if I click Yes the localStorage entry contains "yes", if I click X it's "dismiss", but if I click no, nothing seems to be written / modified in localStorage
- the feedback form doesn't appear anymore in the main field's suggest... ?
- it appears well in the itinerary's suggest but It doesn't seem that the clicks are listened properly (when I click yes/no/x the suggest closes but not the feedback form, and nothing is written in localStorage)
- The other forms work well, but they reappear after clicking yes/no then doing a F5. If I understand well, it's on purpose. But can we avoid showing the same form if the url is the same as one where we already answered? We could have a localstorage entry for every URL where we've clicked yes/no, to avoid asking the same thing at the same place many times?

bbecquet:
@xem here is a point by point anwser to your review.
Most of the problems were because the integration of the feedback component on directions wasn't completed yet, I should have indicated the feature in this PR could be tested only with POI lists at the time. Now that https://github.com/Qwant/erdapfel/pull/1113 is merged and this feature rebased, it should work.

> on mobile (only), after an itinerary search, when I click on on yes / no / x, the feedback form blinks, reappears, and nothing is written in localStorage

This was due to conflicts betwen the panel touch resizing and the feedback element. Doesn't happen now that the feedback block is displayed only on a maximized panel.

> on PC, in the itinerary feedback, in my localStorage the key's name is: "qmaps_v1_userFeedback_undefined" (missing questionId I guess?)

The feedback block on routes didn't have a `questionId` at the time.

> Still in the itinerary feedback, if I click Yes the localStorage entry contains "yes", if I click X it's "dismiss", but if I click no, nothing seems to be written / modified in localStorage

Same.

> the feedback form doesn't appear yet in the fields' suggest... until another branch is merged (questionId will need to be adapted there too)

Yes, normal at this point and out of the scope of this PR which only controls the dismiss remembrance.

> The other feedback forms work well, but they reappear after clicking yes/no then doing a F5. If I understand well, it's on purpose. But can we avoid showing the same form if the url is the same as one where we already answered? We could have a localstorage entry for every URL where we've clicked yes/no, to avoid asking the same thing at the same place many times?

Yes, even if there are different opinions in the team about that, that is on purpose. Remembering the full context would pose some questions itself, so for now we agreed IRL to just remember the dismiss in this PR, and let the door open for later refinements.
The code isn't too complicated yet, but if we introduce complex conditions it would be nice to do it separately and have a new test suite for it.

xem (approved):

PR #1116, opened 2021-06-10
QMAPS-2145 Tweak scale and attribution map controls 

bbecquet:
## Description
Style adjustements to the scale and attribution controls after design review.
- Mobile:
  - the attribution and scale controls are now aligned with the center of the map button on the right
  - both controls now have the same height, translucent background color, and main color 
  - the attribution control opens to the bottom
- Desktop:
  - the scale value is spaced vertically a bit more from the scale bar

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2021-06-10 16-36-33|Capture d’écran de 2021-06-10 16-33-12|
|Capture d’écran de 2021-06-10 16-36-40|localhost_3000_(iPhone 6_7_8)|
|Capture d’écran de 2021-06-10 16-37-26|Capture d’écran de 2021-06-10 16-37-45|

### DIFF
--- a/src/scss/includes/map_theme.scss
+++ b/src/scss/includes/map_theme.scss
@@ -98,7 +98,7 @@
   color: #353c52;
   background-color: transparent;
   height: 12px;
-  line-height: 10px;
+  line-height: 7px;
 }
 
 .map_control__scale_attribute_container {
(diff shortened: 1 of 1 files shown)


xem:
I preferred when it was aligned to the bottom of the geiloc button.  Feels weird like that imo

amatissart (approved):

PR #1118, opened 2021-06-11
Don't display feedback more than once during a session

bbecquet:
## Description
Adjust the rules of display for feedbacks so each answered question is not shown anymore to the user during a "session" (until the page is reloaded).

## Why
 - don't bother the user with too many requests for feedback
 - reduce the risk of being spammed with bad quality feedbacks

### DIFF
--- a/src/libs/userFeedback.js
+++ b/src/libs/userFeedback.js
@@ -22,7 +22,12 @@ function sendAnswer(questionId, answer, { context } = {}) {
 const storagePrefix = `qmaps_v${version}_userFeedback_`;
 const dayToMs = days => days * 24 * 60 * 60 * 1000;
 
+const answeredDuringSession = [];
+
 function shouldBeDisplayed(questionId, hideForDays = 15) {
+  if (answeredDuringSession.includes(questionId)) {
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


src/libs/userFeedback.js
```diff
@@ -36,6 +41,7 @@ function shouldBeDisplayed(questionId, hideForDays = 15) {
 }
 
 function rememberAnswer(questionId, answer) {
+  answeredDuringSession.push(questionId);
   localStorage.setItem(
     `${storagePrefix}${questionId}`,
     JSON.stringify({ answer, date: new Date().toISOString() })
```
amatissart, line 47: Not related to this PR: using the functions provided by "adapters/store.js" would be safer. They take care of potential errors on the browsers where `localStorage` access is restricted.

amatissart (approved):

src/libs/userFeedback.js
bbecquet: Indeed, I'll create a new one for that.

bbecquet: Hop! https://github.com/Qwant/erdapfel/pull/1119

PR #1122, opened 2021-06-16
QMAPS-1797 new poi popup

xem:
## Description
- when a marker is hovered on the map, make it possible to hover its popup too
- if we hover another area of the map or another pin, the current popup closes
- show the same information than the top of PoI panel (name, rating, category, hours) + the poi address inside the poi popup
- the CTA buttons will be added in another MR
- make the poi popup's border radius 12px
- make this system work for both PJ (list) and OSM (no-list) markers & popups

## Screenshots
PJ POI:

image

PJ OSM:

image

Hover behavior (hover area shown in red here)

hover

### DIFF
--- a/src/adapters/poi_popup.js
+++ b/src/adapters/poi_popup.js
@@ -4,7 +4,7 @@ import { Popup } from 'mapbox-gl--ENV';
 import ApiPoi from './poi/idunn_poi';
 import { isMobileDevice } from 'src/libs/device';
 import ReactPoiPopup from 'src/components/PoiPopup';
-import { listen } from 'src/libs/customEvents';
+import { fire, listen } from 'src/libs/customEvents';
 
 const WAIT_BEFORE_DISPLAY = 350;
 
@@ -53,6 +53,7 @@ PoiPopup.prototype.addListener = function (layer) {
 
   this.map.on('mouseleave', layer, async () => {
     clearTimeout(this.timeOutHandler);
+    fire('start_close_popup_timeout');
   });
 };
 
--- a/src/adapters/poi_popup.js
+++ b/src/adapters/poi_popup.js
@@ -7,7 +7,7 @@ import ReactPoiPopup from 'src/components/PoiPopup';
 import { fire, listen } from 'src/libs/customEvents';
 
 const WAIT_BEFORE_DISPLAY = 350;
-const WAIT_BEFORE_CLOSE = 200;
+const WAIT_BEFORE_CLOSE = 350;
 
 function PoiPopup() {}
 
--- a/src/adapters/poi_popup.js
+++ b/src/adapters/poi_popup.js
@@ -7,6 +7,7 @@ import ReactPoiPopup from 'src/components/PoiPopup';
 import { fire, listen } from 'src/libs/customEvents';
 
 const WAIT_BEFORE_DISPLAY = 350;
+const WAIT_BEFORE_CLOSE = 200;
 
 function PoiPopup() {}
 
@@ -15,21 +16,27 @@ PoiPopup.prototype.init = function (map) {
   this.popupHandle = null;
   this.timeOutHandler = null;
   this.activePoiId = null;
+  this.closeTimeoutHandler = null;
 
   listen('open_popup', (poi, e) => {
     if (isMobileDevice() || isTouchEvent(e)) {
       return;
     }
     this.createPJPopup(poi, e);
+    fire('stop_close_popup_timeout');
   });
   listen('close_popup', () => this.close());
-  listen('map_mark_poi', poi => {
-    this.close();
-    this.activePoiId = poi.id;
-  });
   listen('clean_marker', () => {
     this.close();
     this.activePoiId = null;
+  });
+
+  listen('stop_close_popup_timeout', () => clearTimeout(this.closeTimeoutHandler));
+  listen('start_close_popup_timeout', () => {
+    clearTimeout(this.closeTimeoutHandler);
+    this.closeTimeoutHandler = setTimeout(() => {
+      fire('close_popup');
+    }, WAIT_BEFORE_CLOSE);
   });
 };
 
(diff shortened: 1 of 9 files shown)


bbecquet (commented):
Your technique to keep the popup open works, but sadly it's only for POI markers from a list. It should be the case for POIs from the standard map layer.
I think we can find a more generic way (also simpler) that works on both cases. Ideally, with minimal changes on marker code.

For example, we could use a timeout to delay just enough the popup closure on `close_popup`, but this may be a little arbitrary and have side effects.

Another idea is using an invisible area around the popup to receive mouse events.

xem:
Thanks! I went with the invisible hover area.

bbecquet:
This technique seems to work fine, but there is currently a big UX regression: the POI markers themselves can't be clicked anymore…

amatissart:
> This technique seems to work fine, but there is currently a big UX regression: the POI markers themselves can't be clicked anymore…

Yes indeed. Fortunately, it seems to be quite easy to fix with
```diff
   &:after {
     content: '';
     position: absolute;
     top: -50px;
     left: -50px;
     width: calc(100% + 100px);
     height: calc(100% + 100px);
     z-index: -1;
+    pointer-events: none;
   }
 }
```

Needs to be tested further though.

amatissart:
> Fortunately, it seems to be quite easy to fix ...

Ok, it's not so easy. We still need the `mouseout` so that the popup closes properly. :grimacing:

xem:
Hello,
I fixed my config issues! thanks @bbecquet @amatissart 
At first I wanted to give the marker a z-index higher than the popup's "red zone" so it's clickable.
That's not realistic because OSM markers are not HTML divs, they're layers, so they can't be put on top of a popup div.
So the solution I went with is: remove the red zone around the popup.
The result is the following:
1) hover the marker: the popup appears close to it. (and it's still clickable)
2) a) move the mouse on the popup: it works fine
2) b) move the mouse elsewhere: the popup doesn't disappear until we hover another marker or quit the current view.
3) as soon as the popup is un-hovered or the view changes (ex: quit PoI list, click a list item, etc), the popup disappears.

I propose that we check that in real conditions.
I think it's finr like that, but if it's a too big problem that the popup doesn't close if the mouse goes somewhere else, we can still add extra code to close it after a fixed timeout, or if the mouse goes too far from it.
ok?
cc. @Charlotte-Ziegler

xem:
Re, I have implemented a timeout, that makes the popup disappear after 1s if both the marker and the popup are not hovered.
@amatissart has PoC'd another approach, with a surrounding hover area that closes the popup. It's fine too.
hover2

src/adapters/scene_category.js
```diff
@@ -133,22 +141,35 @@ export default class SceneCategory {
...
     if (this.selectedPoi?.id !== poi.id) {
       this.highlightPoiMarker(poi, true);
-      fire('open_popup', this.getPointedPoi(e), e.originalEvent);
+      fire('open_popup', poi, e.originalEvent);
     }
   };
 
   handleLayerMarkerMouseLeave = () => {
     this.map.getCanvas().style.cursor = '';
-    this.highlightPoiMarker(this.hoveredPoi, false);
```
bbecquet, line 147: Why this removal? As a result, the last hovered marker isn't "unhighlighted'.

bbecquet (commented):
I'm ok with a timeout based approach, but the current version still behaves strangely, mainly for tile-based POIs. The popups close immediately if you go from one POI to another.

Related I guess, some code specific to `scene_category.js` should be moved to `poi_popup.js` so we are sure the behavior is shared between every kind of markers.

src/adapters/scene_category.js
```diff
@@ -133,22 +141,35 @@ export default class SceneCategory {
     });
   };
 
-  handleLayerMarkerMouseMove = e => {
+  handleLayerMarkerMouseOver = e => {
     this.map.getCanvas().style.cursor = 'pointer';
     const poi = this.getPointedPoi(e);
+
+    // Un-highlight the previously highlighted PoI if a new one is hovered
```
bbecquet, line 148: This is the role of the `handleLayerMarkerMouseLeave` method to do that.

bbecquet:
@amatissart @xem 
I simplified and fixed some stuff in the hover logic, it now works the same with the tile POI and the list ones. One stuff that could be improved: the POI could keep its hover state when the popup is hovered. But this isn't so simple to do properly for both types, I think it can wait to be treated separately.

As a second commit, I implemented the target design, with shadows and correct font-sizes.

I had to overwrite some CSS stuff with the `PoiItem` rendered DOM, and I used the `inList` bool prop to get the same "compact" rendering as in the category panel.
This isn't ideal semantically, we are not in a list here, but doing it really clean would require some refacto to make `PoiItem` more versatile. Same, I think it's worth waiting a bit, there will probably be design adjustments to make anyway (like, do we display full titles or ellipsis in the popup, etc.).

src/adapters/poi_popup.js
```diff
@@ -118,6 +126,7 @@ PoiPopup.prototype.getPopupAnchor = function (event) {
 };
 
 PoiPopup.prototype.close = function () {
+  fire('stop_close_popup_timeout');
   if (this.popupHandle) {
     this.popupHandle.remove();
     this.popupHandle = null;
```
amatissart, line 132: I think there is a subtle race condition here, when "mouseleave" occurs during the request to the server (after `timeOutHandler` is executed and before the popup is actually open).

`showPopup()` will call `close()` and cancel the `closeTimeoutHandler` here, whereas the mouse has already left the layer. So, the popup is never closed.

Maybe it would be more appropriate to check that `this.popupHandle` exists for this call too:
```suggestion
PoiPopup.prototype.close = function () {
  if (this.popupHandle) {
    fire('stop_close_popup_timeout');
    this.popupHandle.remove();
    this.popupHandle = null;
```

src/adapters/poi_popup.js
```diff
@@ -1,12 +1,13 @@
...
+import ReactDOM from 'react-dom';
 import { Popup } from 'mapbox-gl--ENV';
 import ApiPoi from './poi/idunn_poi';
 import { isMobileDevice } from 'src/libs/device';
 import ReactPoiPopup from 'src/components/PoiPopup';
-import { listen } from 'src/libs/customEvents';
+import { fire, listen } from 'src/libs/customEvents';
 
 const WAIT_BEFORE_DISPLAY = 350;
+const WAIT_BEFORE_CLOSE = 200;
```
amatissart, line 10: Isn't 200ms a bit short to move from the marker to the popup content in some conditions ?

bbecquet: @xem used 1 second at first, which was a long time to make the popup disappear if the user just left the marker without hovering the popup.
But maybe I got too strong with 200ms. What would you recommend? 500ms? Or maybe we can use the same value as the previous line.

amatissart: Ok, let's use 350ms as a precaution.

amatissart (approved):

PR #1123, opened 2021-06-16
Adjust categories labels

amatissart:
## Description
Wording adjustments on the new categories that will be visible after the classifier is enabled on Idunn.  
(See also https://github.com/Qwant/idunn/pull/249 )

### DIFF
--- a/config/categories.yml
+++ b/config/categories.yml
@@ -75,33 +75,33 @@
 # Only categories with 'icon' are visible as action buttons.
 # Below are the categories that are only accessible
 # via queries (with detected intention) or URL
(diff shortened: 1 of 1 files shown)


xem (approved):
nice! I'm just not sure about Doctors office with an "s" ?

amatissart:
> nice! I'm just not sure about Doctors office with an "s" ?

I will keep it this way for now, for consistency with "qwant-maps-common":  
https://github.com/Qwant/erdapfel/blob/1ec6bda/local_modules/qwant-maps-common/src/place_category_name.js#L84

PR #1124, opened 2021-06-16
QMAPS fix suggest display for ios <= 13

xem:
## Description
Suggest had disappeared on iOS <= 13 since https://github.com/Qwant/erdapfel/pull/1108
It's there, on top of the map, but stays hidden.
Removing the position:fixed seems to solve the problem, and not cause any regression, so here it is :)

### DIFF
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -225,7 +225,6 @@ input[type="search"] {
   }
 
   .search_form__result {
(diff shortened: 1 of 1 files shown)


bbecquet (approved):
Can't remember why I changed this prop, probably while fighting again with the iOS scroll behavior. Doesn't seem to change anything in bad indeed, so good catch :+1:

PR #1127, opened 2021-06-17
Service panel: give more space to category name and allow ellipsis

remi-dupre:
## Description

Remove horizontal padding and allow for ellipsis for category names in the service panel.

## Why

This allows to display larger category translations while keeping it centered, ellipsis being used as a last resort if there is not enough space.

## Screenshots

| master (modified strings for showcase) | HEAD |
|-|-|
| Screenshot 2021-06-17 at 13-50-12 Qwant Maps | Screenshot 2021-06-17 at 13-48-57 Qwant Maps |

### DIFF
--- a/src/scss/includes/components/mainActionButton.scss
+++ b/src/scss/includes/components/mainActionButton.scss
@@ -1,6 +1,6 @@
 .mainActionButton {
   cursor: pointer;
-  padding: 8px 0px;
+  padding: $spacing-xs $spacing-xxxs;
   border-radius: 8px;
 
   &-label {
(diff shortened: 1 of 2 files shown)


src/scss/includes/components/mainActionButton.scss
```diff
@@ -1,6 +1,6 @@
 .mainActionButton {
   cursor: pointer;
-  padding: 8px;
+  padding: 8px 0px;
```
bbecquet, line 3: I would keep a small value instead of 0px, to prevent two consecutive long strings appear completely glued.
`4px` looks ok to keep "Supermarché" full, but maybe its safer to use `2px` to have some margin for the different fonts depending on platforms.
BTW, this is a good case to use "official" spacing values :) See `spacing.scss` for ref.

remi-dupre: :+1: 
I'm not sure if `border-radius: 8px;` right bellow shouldn't be treated a spacing and left as it is?

bbecquet (approved):

PR #1128, opened 2021-06-18
Babel config update

bbecquet:
## Description
 - Update @babel/preset-env
 - Remove some unused or obsolete plugins from dependencies
 - Update the browser target list:
   - remove IE 11
   - set Safari minimal version to 8. Search uses 10, but I checked, we still have some (very few) visits with 8 and 9… Can be discussed later
 - Merge the dev/prod `babel.config.js` (in the 'build' directory) with the one used by Jest (in project's root), as it was confusing. Let's just have a single one at the root, as it's a de facto standard.

## Why
I have my hands deep in the build config these days and noticed stuff could be improved.


babel.config.js
```diff
@@ -1,8 +1,32 @@
...
-}
+module.exports = function (api) {
+  api.cache(true);
+
+  const presets = [
+    [
+      '@babel/preset-env',
+      {
+        targets: {
+          browsers: ['Chrome >= 60', 'Firefox >= 44', 'Safari >= 8', 'Edge >= 17'],
```
amatissart, line 9: In addition to Safari, it could be worth to look at the minimal version of Firefox we need to support.  

With
```diff
-          browsers: ['Chrome >= 60', 'Firefox >= 44', 'Safari >= 8', 'Edge >= 17'],
+          browsers: ['Chrome >= 60', 'Firefox >= 60', 'Safari >= 10', 'Edge >= 17'],
```
the size of the first bundle would shrink from 527KB to 420KB.

amatissart (approved):
Nice!

PR #1136, opened 2021-06-22
Properly ignore intention about unknown category in fullText_query middleware

amatissart:
## Description
The intention validation used on server side is made consistent with the implementation defined in "src/adapters/intention.js"

## Why
The validation will transparently ignore intentions that refer to an unknown category. This is more robust and allows new categories to be defined on the back end without disturbing the application.

### DIFF
--- a/bin/middlewares/fullText_query.js
+++ b/bin/middlewares/fullText_query.js
@@ -19,7 +19,7 @@ module.exports = function (config) {
   const isIntentionValid = intention =>
     intention &&
     intention.filter &&
-    (intention.filter.q || isCategoryValid(intention.filter.category));
+    (!intention.filter.category || isCategoryValid(intention.filter.category));
 
   // @TODO: import it from client lib src/libs/url_utils.js when possible
   const removeNullEntries = obj =>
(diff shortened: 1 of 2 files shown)


bin/middlewares/fullText_query.js
```diff
@@ -33,13 +38,13 @@ module.exports = function (config) {
...
     const intention = (response.data.intentions || [])[0];
-    if (intention && intention.filter) {
+    if (isIntentionValid(intention)) {
       const { q, bbox, category } = intention.filter;
       const params = new URLSearchParams(
         removeNullEntries({
           q,
           bbox: bbox && bbox.join(','),
-          type: isCategoryValid(category) ? category : null,
+          type: category,
```
bbecquet, line 47: I'm not sure so it's better to ask: isn't there a case where `q` exists (therefore the result of `isIntentionValid` is true), and `category` also exists but isn't valid?

amatissart: In practice, the API would not currently build an intention with both `q` and `category` set.  
Nevertheless you're right, any intention with an unknown category should be ignored. I just updated the condition to make it more explicit about this case.

bbecquet (approved):

PR #1142, opened 2021-06-28
QMAPS-2178 add popup buttons

xem:
## Description
Add buttons at the bottom of OSM/PJ Popups (directions / phone / booking / like / share)

## Screenshots
image

### DIFF
--- a/src/components/PoiPopup.jsx
+++ b/src/components/PoiPopup.jsx
@@ -2,13 +2,13 @@ import React, { useState } from 'react';
 import PoiItem from './PoiItem';
 import { fire } from 'src/libs/customEvents';
 import ActionButtons from '../panel/poi/ActionButtons';
-import nconf from '@qwant/nconf-getter';
 import Telemetry from '../libs/telemetry';
 import { addToFavorites, isInFavorites, removeFromFavorites } from '../adapters/store';
+import { useConfig } from '../hooks';
 
 const PoiPopup = ({ poi }) => {
   const [inFavorites, setInFavorites] = useState(isInFavorites(poi));
-  const isDirectionActive = nconf.get().direction.enabled;
+  const isDirectionActive = useConfig('direction').enabled;
 
   const openDirection = () => {
     Telemetry.sendPoiEvent(poi, 'itinerary');
--- a/src/components/PoiPopup.jsx
+++ b/src/components/PoiPopup.jsx
@@ -1,8 +1,46 @@
 import React from 'react';
 import PoiItem from './PoiItem';
 import { fire } from 'src/libs/customEvents';
+import ActionButtons from '../panel/poi/ActionButtons';
+import nconf from '@qwant/nconf-getter';
+import Telemetry from '../libs/telemetry';
+import { addToFavorites, isInFavorites, removeFromFavorites } from '../adapters/store';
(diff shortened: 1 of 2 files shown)


xem:
@bbecquet the functions I introduced in this file are not copy-pasted but mostly similar to other functions present in the project, do you think they could (should) be factorized better than that? Thanks :)

bbecquet:
> @bbecquet the functions I introduced in this file are not copy-pasted but mostly similar to other functions present in the project, do you think they could (should) be factorized better than that? Thanks :)

It works fine, it's ok for me like that :+1:
Actually I'd say we shouldn't bother too much, because both things will probably change soon:
 - we have ongoing talks about telemetry, the way we do it may change in the near future,
 - to be managed properly, favorite add/remove operations should use a global context instead of a local state, so everything updates simultaneously. I've wanted to do it for a long time, but the presence of this new button will be an incentive :)

src/components/PoiPopup.jsx
```diff
@@ -1,8 +1,48 @@
...
 import PoiItem from './PoiItem';
 import { fire } from 'src/libs/customEvents';
+import ActionButtons from '../panel/poi/ActionButtons';
+import nconf from '@qwant/nconf-getter';
+import Telemetry from '../libs/telemetry';
+import { addToFavorites, isInFavorites, removeFromFavorites } from '../adapters/store';
 
 const PoiPopup = ({ poi }) => {
+  const [inFavorites, setInFavorites] = useState(isInFavorites(poi));
+  const isDirectionActive = nconf.get().direction.enabled;
```
bbecquet, line 11: There is now a `useConfig` hook as a simpler and more react-ish alternative to nconf. You can already find it used elsewhere in the project.

src/components/PoiPopup.jsx
```diff
@@ -13,7 +53,17 @@ const PoiPopup = ({ poi }) => {
         fire('close_popup');
       }}
     >
-      <PoiItem poi={poi} withOpeningHours withImage inList />
+      <div className="u-mb-s">
+        <PoiItem poi={poi} withOpeningHours withImage inList />
+      </div>
+      <ActionButtons
```
bbecquet, line 59: The button size should be reduced to match the design.
As we don't have the `small` size in our `Button` element, you probably have to overide some stuff in `popup.scss` like I did for text elements.

src/components/PoiPopup.jsx
xem: thanks! should I create a ticket to remove all the remaining nconf occurrences?

bbecquet: I'd say we don't need one. There is only one occurrence left in React components, in PoiPanel, and it can't be changed until it has been turned to a functional component… which I'm working on :slightly_smiling_face: .

xem: er, I meant these: 
image

bbecquet: They don't need to be modified nor become React components, helpers like util functions and data processing don't need to be React :) And we can't remove `nconf` completely. `useConfig` uses it behind the scene, it's just syntaxic sugar to avoid  doing `nconf.get().foo…` everywhere.

PR #1148, opened 2021-07-01
Add a POI block for social network links

bbecquet:
## Description
Add a block in POI panel to display social network links, if available.

## Screenshots
Capture d’écran de 2021-07-01 18-01-11


src/components/ui/Flex.jsx
```diff
@@ -3,20 +3,35 @@ import classnames from 'classnames';
...
-        ...style,
-      }}
-      {...rest}
-    >
-      {children}
-    </div>
-  )
+  (
+    {
+      as = 'div',
```
bbecquet, line 8: The prettier reformatting makes it hard to see, but I've only added the capability to pass a custom HTML tag to this element, if we want something else than a `<div>`. This avoids some unnecessary element nesting.
This is a feature in the corresponding component of the DS, so it will ease the migration.

src/panel/poi/blocks/Block.jsx
```diff
@@ -22,7 +22,11 @@ const Block = ({
       title={tooltip}
       {...rest}
     >
-      <i className={`block-icon icon-${icon}`} />
+      {typeof icon === 'string' ? (
+        <i className={`block-icon icon-${icon}`} />
+      ) : (
+        <div className="u-mr-s">{icon}</div>
```
bbecquet, line 28: Needed change because this new icon wasn't in the icon font and we prefer using SVG icons now.

bbecquet:
In case someone wonders, is here what it looks like when the POI has the 4 possible social networks:
Capture d’écran de 2021-07-01 18-24-17

It simply breaks the line. I don't think we need to bother with something more complicated.

xem (approved):

PR #1150, opened 2021-07-06
QMAPS-2203 prevent multiple geolocation popups per session

xem:
## Description
Once the position is requested (with "my position" suggest item or Mapbox's Geolocation button), and the permission has been granted, watch the user's position for all the session.
This is done by polyfilling getCurrentPosition and watchPosition with a custom function that mimics them without requesting the browser many times

## Why
calling many times getCurrentPosition or watchPosition duting the same session trigger Firefox Desktop's permission popup many times which is annoying

## TO DO
test to be sure that everything still works on other OS / devices / browsers than Windows 10 / PC / Fx & Chrome

## TO DEBATE
Is it OK to "watch" the user's position even when the Mapbox geolocation button is disabled? (I'd say yes).
No information about the user's position is sent to our servers, and the user is aware that his position is known with this icon:
image

## Screenshots
gps

### DIFF
--- a/src/adapters/poi/specials/navigator_geolocalisation_poi.js
+++ b/src/adapters/poi/specials/navigator_geolocalisation_poi.js
@@ -25,7 +25,7 @@ export default class NavigatorGeolocalisationPoi extends Poi {
   async geolocate(options = { displayErrorModal: true }) {
     return new Promise((resolve, reject) => {
       this.status = navigatorGeolocationStatus.PENDING;
-      navigator.geolocation.getCurrentPosition(
+      window.getPosition(
         position => {
           this.setPosition({ lat: position.coords.latitude, lng: position.coords.longitude });
           resolve();
--- a/src/libs/geolocation.js
+++ b/src/libs/geolocation.js
@@ -26,3 +26,42 @@ export function handleError(error) {
     fire('open_geolocate_not_activated_modal');
   }
 }
+
+window.position = null;
+window.positionWatchID = null;
+
+// Centralize all the position requests in a single function,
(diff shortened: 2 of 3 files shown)


bbecquet (changes requested):
I'm really against this solution, sorry.

First, there is no more distinction in your API between `watchPosition` and `getCurrentPosition`. Meaning if you really want to ask a second `watchPosition`, you won't even get the periodic updates.
And meaning if you ask a `getCurrentPosition` first, without having the Mapbox watch button activated in the first place (a common case if you just want an itinerary), you will trigger a `watchPosition` that will send success callbacks forever.

Next, this is a major break of user confidence to display all the signs of not following their positions anymore, while still doing so. Even if we don't do anything server-side with that position.
Granted, the address bar icon you pointed means "I gave permission to this site to access my position". But this doesn't mean "My position is currently being monitored". For this, the user has to rely on specific UI elements of the site, like the blue marker and the button. These are almost conventions, we shouldn't break it.

There is no more way to completely stop this process once it's launched, so there are also battery life considerations.

Finally, overriding native functions should be done in very last resort, to avoid breaking other things and being in a worse case than previously.
I don't think this single bug is enough for that.

xem:
I also believe this "bug" is not worth all that.
But I also don't think it's useful to be able to disable geolocation during a browsing session.
What if the geolocation button simply disappeared after it has been clicked once?

amatissart:
A workaround for the geolocation issues on Firefox Desktop has been implemented in #1157.  
Even if the authorization may appear twice in a session, errors should hopefully be rare. So this fix does not seem necessary for now.

PR #1151, opened 2021-07-07
QMAPS-1699 redesign map buttons

xem:
## Description
Design fixes for the map's buttons
Enhance existing CSS
I discussed the details with Design team, so it's normal if everything is not like in Zeplin:

PC:
- 4px radius
- Geoloc is not merged with zoom anymore
- Zoom+ (with radius only on top) is merged with Zoom- (with radius only on the bottom)
- Hover style is the same as Menu/Product buttons (light grey)

PC & mobile:
- When geolocation is enabled, put the icon inside the button in blue 

NB: Geoloc/+/- Icons from Remix will be introduced in another ticket

## Why
Design!

## Screenshots

image

### DIFF
--- a/src/scss/includes/component.scss
+++ b/src/scss/includes/component.scss
@@ -10,7 +10,7 @@ $ui_margin: 12px;
 }
 
 @mixin card_radius() {
-  border-radius: 3px;
+  border-radius: 4px;
 }
 
 @mixin card_shadow() {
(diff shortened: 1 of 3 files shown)


bbecquet:
There seems to be a slight gray gradient on the + button, instead of single separator line. As a result, the + button appears to be below the - button.

Capture d’écran de 2021-07-08 17-12-33

xem:
Fixed!

amatissart (approved):

PR #1160, opened 2021-07-20
QMAPS-2220 change page title depending on the app state

bbecquet:
## Description
Introduce a new React hook to easily change the page title as a side effect, and use it in the app panels.

|Panel|Page title|
|---|---|
|Service|Qwant Maps|
|POI|\<poi name\> - Qwant Maps|
|Category|\<query or category\> - Qwant Maps|
|Favorites|Favorite places - Qwant Maps|
|Direction|Directions - Qwant Map|

These rules can be improved later, for example by including the origin and destination points in the direction panels, but let's start simple.

## Why
Facilitate navigation in browser history + accessibility

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2021-07-20 15-29-33|Capture d’écran de 2021-07-20 11-33-49|

### DIFF
--- a/src/panel/category/CategoryPanel.jsx
+++ b/src/panel/category/CategoryPanel.jsx
@@ -7,7 +7,7 @@ import PoiItemList from './PoiItemList';
 import PoiItemListPlaceholder from './PoiItemListPlaceholder';
 import CategoryPanelError from './CategoryPanelError';
 import Telemetry from 'src/libs/telemetry';
-import { useConfig, useDevice } from 'src/hooks';
+import { useConfig, useDevice, usePageTitle } from 'src/hooks';
 import IdunnPoi from 'src/adapters/poi/idunn_poi';
 import { getVisibleBbox } from 'src/panel/layouts';
 import { fire, listen, unListen } from 'src/libs/customEvents';
@@ -17,6 +17,7 @@ import { sources } from 'config/constants.yml';
 import { BackToQwantButton } from 'src/components/BackToQwantButton';
 import { shouldShowBackToQwant } from 'src/libs/url_utils';
 import { Panel, PanelNav, SourceFooter, UserFeedbackYesNo } from 'src/components/ui';
+import { capitalizeFirst } from 'src/libs/string';
 
 const DEBOUNCE_WAIT = 100;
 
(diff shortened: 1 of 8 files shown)


amatissart:
What about the `envName` that is currently used as a suffix in the `<title>`:
https://github.com/Qwant/erdapfel/blob/1d67287fb4d07a9803e715fb794f6dd842369224/views/index.ejs#L5-L7

It would be lost during the navigation, right ?

bbecquet:
Thanks @amatissart for the `envName`.
I fixed it in https://github.com/Qwant/erdapfel/pull/1160/commits/4fd378131cfdcbe347dc7fec1f59c23fb865908b

xem:
Great, but it seems buggy when envName is not defined (like in prod)
image

bbecquet:
> Great, but it seems buggy when envName is not defined (like in prod)
> There should be neither [object object], nor " - " at the end of the title :-)
> Here's what I see locally with the default config.yml
> 
> Also, the tests don't pass.
> 
> image

src/panel/category/CategoryPanel.jsx
```diff
@@ -64,6 +65,8 @@ const CategoryPanel = ({ poiFilters = {}, bbox }) => {
   const { isMobile } = useDevice();
   const { maxPlaces } = useConfig('category');
 
+  usePageTitle(capitalizeFirst(poiFilters.category || poiFilters.query));
```
amatissart, line 68: `poiFilters.category` is the value defined by the API and is not translated. Also should `poiFilters.query` really be capitalized ?

Perhaps this logic could be shared with the existing [`getTopBarAppValue()`](https://github.com/Qwant/erdapfel/blob/1d67287fb4d07a9803e715fb794f6dd842369224/src/panel/PanelManager.jsx#L19-L30).

bbecquet: I first went into refactoring once again parts of upper app state, but finally I just defined a new function to share this https://github.com/Qwant/erdapfel/pull/1160/commits/925c432320505892c763fcd74f9ab097c828a5eb

amatissart (approved):

PR #1161, opened 2021-07-21
QMAPS-2033 add "add to home" infobox on mobile after the 5th visit

xem:
## Description
add "add to home" infobox on mobile after the 5th visit
TODO: use com center link when it's ready

## Screenshots

addtohome


src/components/AddToHomeInfoBox.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+const AddToHomeInfoBox = () => {
+  const [closed, setClosed] = useState(get('add_to_home_closed'));
+  const closeBetaPopup = () => {
+    set('add_to_home_closed', 1);
+    setClosed(true);
+  };
+
+  const visits = parseInt(get('visits') || 0) + 1;
+  set('visits', visits);
```
bbecquet, line 14: This is a good example of a side effect that we want to control more and safely separate from the component rendering. 

If we put it here in the render function, we will increment `visits` each time this component is re-rendered, which can happen at any moment in the app not related to a single hit.
You can already see it in these cases:
 - when the window is resized and the device is changed from desktop to mobile (as it's rendered as a child of a `DeviceContext.Provider`)
 - after the box is displayed and we call `setClosed(true)`

To ensure this increment happens only on mounting the component, you could put it inside a `useEffect(…, [ ])`.
But this still means if for some reason we unmount and remount this component during a user session, it will be counted twice.
So the safer approach IMO is to update this counter outside this component, where we are 100% sure we'll execute code only once per session. For example, as a sideEffect of `RootComponent` or in `app_panel`. The `get` can still happen here as it doesn't modify anything.

bbecquet:
The `<Notification>` component has been removed as part of https://github.com/Qwant/erdapfel/pull/1172. It was to avoid dead code, but I forgot it was used by this PR.
I rebased it with a commit restoring the component.

PR #1162, opened 2021-07-21
QMAPS-2008 top bar adjustments

xem:
## Description
- The Qwant logo lacked 1 pixel in width
- Align search bar text and suggest items text when typing:
- Desktop:
image

- Mobile:
image

### DIFF
--- a/src/scss/includes/placeIcon.scss
+++ b/src/scss/includes/placeIcon.scss
@@ -1,5 +1,4 @@
 .placeIcon {
-  display: block;
   font-size: 32px;
   border-radius: 50%;
   width: 36px;
--- a/src/scss/includes/search_form.scss
+++ b/src/scss/includes/search_form.scss
@@ -63,7 +63,7 @@ input[type="search"] {
 // Logo
 .search_form__logo__button {
   cursor: pointer;
-  width: 36px;
+  width: 37px;
   height: 36px;
   flex-shrink: 0;
   background: url(../images/qwant-logo.svg) no-repeat;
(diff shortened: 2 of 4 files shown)


src/scss/includes/search_form.scss
```diff
@@ -63,7 +63,7 @@ input[type="search"] {
 // Logo
 .search_form__logo__button {
   cursor: pointer;
-  width: 36px;
+  width: 37px;
```
bbecquet, line 66: This change (the other ones also) looks like a "magical" value, which reveals a very fragile layout.
We should address such questions more globally, by compositing standard margin/padding and size values, instead of fine-tuning everything constantly.

src/scss/includes/placeIcon.scss
```diff
@@ -2,7 +2,7 @@
   display: block;
   font-size: 32px;
   border-radius: 50%;
-  width: 36px;
+  width: 34px;
```
bbecquet, line 5: Another change that seems very arbitrary. And as a result, the favorite icon in your screenshots looks compressed.

src/scss/includes/search_form.scss
xem: I added this pixel because the logo seems truncated on the right (on my PC)
With the 37px width, it's fine.
image

xem:
@bbecquet I agree with your remarks. I simplified my code to remove most magic values, just adjusted a couple margins to align things properly.

bbecquet (approved):

PR #1166, opened 2021-07-29
QMAPS 2208 Add delivery block in POI details

bbecquet:
## Description
Add the "delivery" info in the "details" block of a POI.
As we want to display only active options, I added a util function (and its test cases) to help determining if we should render the block.

## Screenshots
Capture d’écran de 2021-07-29 13-44-13


src/panel/poi/blocks/helpers.js
```diff
@@ -0,0 +1,3 @@
+export function hasDisplayableDelivery(delivery) {
+  return !!delivery && Object.entries(delivery).some(([_key, value]) => value === 'yes');
```
amatissart, line 2: This is not very future-proof. 
If a new key is added in the API response (we would expect that to be backward-compatible), the `DeliveryBlock` may be displayed whereas none of the 3 known labels is enabled. 

So this logic is necessarily intertwined with the rendering of the block. Perhaps it would make sense to define such an helper in the same file as the block itself?

bbecquet: Indeed! I changed it.
Thanks for the suggestion of bringing the helper and the block together, it may be the cleaner approach for all these blocks actually.

amatissart (approved):

PR #1167, opened 2021-07-29
QMAPS-1625 [Refacto] Replace icomoon icons

bbecquet:
## Description
Replace the **icomoon** icons across the app by their SVG counterparts from the Remix set (or rarely from individual files coming from the designs).

As a result, some icons changed a bit, for example in the POI Panel, but this makes them more compliant with the designs and consistant with Search.

The following icons are replaced:
 - all icons of POI blocks
 - star icons, used for reviews
 - icons of the "Share" menu
 - vehicle types for directions
 - chevrons and arrows
 - geoloc, zoom +, -

## NB 
 - For the geoloc and zoom +/- icons, used in the map buttons (which are not React and partially rendered by Mapbox-GL itself), I used the `renderReactAsStaticMarkup` trick so the SVG are inlined and the current styling works. This was the quickiest way to get rid of icomoon but this feels overkill. Maybe it's worth trying to clean this part of the app soon in the future.
 - I voluntarily let some icons out: the ones that are only used by components not used anymore, like the Covid block and the public transport components. If we reintroduce them, we'll probably have to change their design anyways, so we can switch to SVG icons at the moment.
 - I didn't change the documentation in this PR, I'll do it in the next one.

## Why
Drop Icomoon, to avoid maintaining three different systems of pictograms (SVG icons from Remix, POI types icons from the map font, and icomoon).
Ease the maintenance, simplify some code, config and doc.
One less font file to fetch when the app loads.


src/scss/includes/placeIcon.scss
```diff
@@ -8,9 +8,4 @@
   align-items: center;
   justify-content: center;
   flex-shrink: 0;
-
-  // indicates it's using the icomoon font
-  &:not(.icon) {
-    font-size: 20px;
-  }
```
amatissart, line 15: A similar css rule exists in "vehicleSelector.scss" and could be removed. 
(The vehicle icons are slightly displaced to the top).
https://github.com/Qwant/erdapfel/blob/2d20d7a0449b42ad36bfc0aed0487e1962bc3630/src/scss/includes/vehicleSelector.scss#L28

amatissart (commented):
> I voluntarily let some icons out: the ones that are only used by components not used anymore, like the Covid block and the public transport components. If we reintroduce them, we'll probably have to change their design anyways, so we can switch to SVG icons at the moment.

Maybe we could at least mark this obsolete usage with TODO comments, for easier update in the future?  
As far as I can see the affected components are:
 * `Text` (only used by `Recycling`)
 * `RouteVia`
 * `RouteSummaryInfo`
 * `TransportLineLeg`
 * `Covid19`

bbecquet:
@amatissart 

> Maybe we could at least mark this obsolete usage with TODO comments, for easier update in the future?

Good advice, I've done it. Except for the `Text` component, which I've removed completely.

src/panel/poi/blocks/Block.jsx
```diff
@@ -22,7 +22,7 @@ const Block = ({
       title={tooltip}
       {...rest}
     >
-      {typeof icon === 'string' ? <i className={`block-icon icon-${icon}`} /> : icon}
+      <div className="block-icon">{icon}</div>
```
amatissart, line 25: Nitpick: This class "block-icon" adds an extra margin, which is not suitable for icons used in the "Details" section. See this spectacular Before/After gif:
Peek 03-08-2021 17-14

amatissart (commented):
Looks good to me, maybe except minor alignment issue.

Do you still prefer updating the documentation and Readme in another PR?

bbecquet:
@amatissart I fixed the alignment bug, thanks for spotting it! And I also pushed some doc updates to remove references to Icomoon everywhere.

Thanks for the review, and sorry for this big PR.

amatissart (approved):

PR #1173, opened 2021-08-06
QMAPS-2068 navigate in app menus through history

bbecquet:
## Description
Allow the app menu to be closed by navigating back (and also to be remembered on page reload).  
For that, store the menu state in the url using the same mechanism as Search: a `drawer` query string param indicating the currently opened menu panel (aka. drawer)
In Maps, can be `app` or `products`.

Required a new event sent by the router on page loads so `<Menu>` can be notified.
It's still a bit hackish, not declarative enough for React, but I think it's the minimum change for the current app structure.

It gave me clearer ideas to clean up our router/history implementation (for now there are still too many URL manipulations in deep components), ideally migrating to React Router, but let's do that separately.

### DIFF
--- a/src/libs/url_utils.js
+++ b/src/libs/url_utils.js
@@ -45,9 +45,14 @@ export function joinPath(parts) {
     .join('/');
 }
 
-export function getCurrentUrl() {
-  const { pathname, search, hash } = window.location;
-  return `${pathname}${search}${hash}`;
+export function getAppRelativeUrl() {
+  const { search, hash } = window.location;
+  return `${getAppRelativePathname()}${search}${hash}`;
+}
+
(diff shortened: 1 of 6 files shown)


xem (changes requested):
I agree, this code is fine but a route refactoring later would be nice!

- The url is well updated when opening/closing the burger menu or the products menu,
- Going back in the History correctly reopens the menu
- But after a page reload, the menu isn't open!

bbecquet:
@xem Thanks for the page load bug!
It happened because the `routeChange` listener is created after the first event occurs, so I added a case for that in the mounting effect of `Menu`.

xem (approved):
amazing

PR #1176, opened 2021-08-16
Fix relative URL routing used by Menu

bbecquet:
## Description
Follow-up to https://github.com/Qwant/erdapfel/pull/1173 
Fix a letter case issue (`baseURL` instead of `baseUrl`), which caused the retrieval of app-relative urls to return wrong values on domains which use a prefix part (ex: prod or dev, which use `/maps`).
The prefix wasn't removed, so opening/closing the menu created urls with repeated prefix.
Capture d’écran de 2021-08-16 15-55-51

This function is hackish anyway, switching to React Router should make it simpler to work with relative urls.

### DIFF
--- a/src/libs/url_utils.js
+++ b/src/libs/url_utils.js
@@ -49,13 +49,8 @@ export function joinPath(parts) {
     .join('/');
 }
 
-export function getAppRelativeUrl() {
-  const { search, hash } = window.location;
-  return `${getAppRelativePathname()}${search}${hash}`;
-}
-
 export function getAppRelativePathname() {
-  const appBase = (window.baseURL || '/').replace(/\/$/, '');
+  const appBase = (window.baseUrl || '/').replace(/\/$/, '');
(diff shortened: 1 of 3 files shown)


xem (dismissed):

src/libs/url_utils.js
```diff
@@ -55,7 +55,7 @@ export function getAppRelativeUrl() {
 }
 
 export function getAppRelativePathname() {
-  const appBase = (window.baseURL || '/').replace(/\/$/, '');
+  const appBase = (window.baseUrl || '/').replace(/\/$/, '');
```
amatissart, line 58: Easy mistake, especially as our global `baseUrl` should not be confused with [`document.baseURI`](https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURI) (similar value except it contains a complete URL: protocol, host, etc.)

amatissart (approved):

amatissart (approved):

xem (approved):

PR #1177, opened 2021-08-16
QMAPS-434 add poi stars for lodging and restaurants

xem:
## Description
- cleanup commented code
- add "stars" block with two fields: has_stars (unknown/yes) / nb_stars (float > 0 if it is known, 0 or undefined if not known)
- Fake data (to remove when the API is ready, hence the draft status)
- display the block ("Starred {type}"   or   "{Type} with {n} stars")
- Ensure proper capitalization in the languages where the string starts (or does not start) with "{Type}"

## Screenshots
image
image

### DIFF
--- a/src/hooks/useI18n.js
+++ b/src/hooks/useI18n.js
@@ -9,6 +9,7 @@ export const useI18n = () => {
 
   return {
     _: window._,
+    _n: window._n,
     locale,
     lang,
     getLocalizedUrl: getLocalizedUrl(lang),
--- a/src/panel/poi/blocks/Details.jsx
+++ b/src/panel/poi/blocks/Details.jsx
@@ -14,18 +14,7 @@ const DetailsBlock = ({ poi }) => {
   const accessibility = findBlock(poi.blocks, 'accessibility');
   const internetAccess = findBlock(poi.blocks, 'internet_access');
   const delivery = findBlock(poi.blocks, 'delivery');
-  // const stars = findBlock(poi.blocks, 'stars');
-  // Example:
-  const stars = {
-    type: 'stars',
-    ratings: [
-      {
-        has_stars: 'yes',
-        nb_stars: 0,
-        // kind: 'restaurant',
-      },
-    ],
-  };
+  const stars = findBlock(poi.blocks, 'stars');
 
   if (!accessibility && !internetAccess && !hasStars(stars) && !hasActiveDeliveryModes(delivery)) {
     return null;
--- a/src/panel/poi/blocks/Details.jsx
+++ b/src/panel/poi/blocks/Details.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import StarsBlock from './Stars';
+import StarsBlock, { hasStars } from './Stars';
 import AccessibilityBlock from './Accessibility';
 import InternetAccessBlock from './InternetAccess';
 import DeliveryBlock, { hasActiveDeliveryModes } from './Delivery';
@@ -17,11 +17,17 @@ const DetailsBlock = ({ poi }) => {
   // const stars = findBlock(poi.blocks, 'stars');
   // Example:
   const stars = {
-    has_stars: 'yes', // yes or unknown
-    nb_stars: 4, // float
+    type: 'stars',
+    ratings: [
+      {
+        has_stars: 'yes',
+        nb_stars: 0,
+        // kind: 'restaurant',
+      },
+    ],
   };
 
-  if (!accessibility && !internetAccess && !hasActiveDeliveryModes(delivery)) {
+  if (!accessibility && !internetAccess && !hasStars(stars) && !hasActiveDeliveryModes(delivery)) {
     return null;
   }
 
--- a/src/panel/poi/blocks/Details.jsx
+++ b/src/panel/poi/blocks/Details.jsx
@@ -1,20 +1,25 @@
 import React from 'react';
+import StarsBlock from './Stars';
 import AccessibilityBlock from './Accessibility';
-// import BreweryBlock from './Brewery';
 import InternetAccessBlock from './InternetAccess';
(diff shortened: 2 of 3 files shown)


src/panel/poi/blocks/Stars.jsx
```diff
@@ -0,0 +1,35 @@
+/* global _n */
```
bbecquet, line 1: We should use the hook for this function too, it's easy to export it from `useI18n` like we did for `_` (should have done it when creatinf the hook, sorry)

src/panel/poi/blocks/Stars.jsx
```diff
@@ -0,0 +1,35 @@
...
+import React from 'react';
+import Block from './Block';
+import { IconStar } from 'src/components/ui/icons';
+import { ACTION_BLUE_BASE } from 'src/libs/colors';
+import { capitalizeFirst } from 'src/libs/string';
+import { useI18n } from '../../../hooks';
```
bbecquet, line 7: You can use the alias `src/hooks` like you did with the lines before, that way it's easier to move files.

src/panel/poi/blocks/Details.jsx
```diff
@@ -1,20 +1,25 @@
...
-  // const brewery = findBlock(poi.blocks, 'brewery');
-  // {brewery && <BreweryBlock block={brewery} />}
+  // const stars = findBlock(poi.blocks, 'stars');
+  // Example:
+  const stars = {
+    has_stars: 'yes', // yes or unknown
+    nb_stars: 4, // float
+  };
 
   if (!accessibility && !internetAccess && !hasActiveDeliveryModes(delivery)) {
```
bbecquet, line 24: There will be the same problem with this block, we shouldn't render the whole "details" section if nothing is rendered inside, so we have to add a check there. And the condition is potentially complex (because the block can exist with a 'unknown' value), like for delivery it may require a dedicated function.

src/panel/poi/blocks/Details.jsx
```diff
@@ -1,20 +1,25 @@
...
-
+  const subclass = poiSubClass(poi.subClassName);
   const accessibility = findBlock(poi.blocks, 'accessibility');
   const internetAccess = findBlock(poi.blocks, 'internet_access');
   const delivery = findBlock(poi.blocks, 'delivery');
-  // const brewery = findBlock(poi.blocks, 'brewery');
-  // {brewery && <BreweryBlock block={brewery} />}
+  // const stars = findBlock(poi.blocks, 'stars');
+  // Example:
+  const stars = {
```
bbecquet, line 19: Just to be sure, this fake data is just a subpart of the stars block. It's ok for now to focus on the rendering, but there will still be the actual parsing and filtering to implement.
Maybe you can already use the "real" format:
```json
{
  "type": "stars",
  "ratings": [
    {
      "has_stars": "yes",
      "nb_stars": 4,
      "kind": "restaurant",
     } 
  ]        
}
```

src/panel/poi/blocks/Stars.jsx
xem: Sure, what hook do you mean, usei18n()._n ?
and if we add it, does it mean that this comment /* global _n */ will no longer be necessary?
Thanks

src/panel/poi/blocks/Details.jsx
xem: sure, though the last time I discussed this with Adrien today, "kind" actually seemed unnecessary. poi.subclass seemed enough. I'll leave it out for now but of course, if any use case appears, let's put it back

src/panel/poi/blocks/Stars.jsx
bbecquet: You can export `_n` the same way it's already done with `_` in the hook here https://github.com/Qwant/erdapfel/blob/master/src/hooks/useI18n.js#L10-L15.
That way you can import both in one round:
```js
const { _, _n } = useI18n();
``` 
And yes, as a result, the global component will have to be removed, as the linter won't complain anymore.
That's the point of it, reducing the use of globals in the app by scoping them. For now the hook acts just as a relay, as you see it's re-exporting `window` globals. But each time we use it it's one less global, so when there are no more it will be a single point to migrate to a better system.

src/panel/poi/blocks/Stars.jsx
```diff
@@ -0,0 +1,41 @@
...
+    <Block simple icon={<IconStar fill={ACTION_BLUE_BASE} width={20} />}>
+      {_('Starred {subclass}', 'poi', { subclass })}
+    </Block>
+  );
+};
+
+export default StarsBlock;
+
+export const hasStars = stars =>
+  stars && stars.ratings && stars.ratings[0] && stars.ratings[0].has_stars === 'yes';
```
bbecquet, line 41: Nice to have defined this function :+1: 
If you want, you can greatly simplify the syntax here with the optional chaining character:
```js
export const hasStars = stars => stars?.ratings?.[0]?.has_stars === 'yes' || false;
// the '|| false' is to ensure we return a boolean, as optional chaining returns `undefined` if one part doesn't exist 
```

src/hooks/useI18n.js
```diff
@@ -9,6 +9,7 @@ export const useI18n = () => {
 
   return {
     _: window._,
+    _n: window._n,
```
amatissart, line 12: As discussed in #1179, it may be required to defined `_n` as a reserved name. Otherwise it could be renamed in the minimized bundle, and the translation keys would not be parsed.  

https://github.com/Qwant/erdapfel/blob/1e4919d4592df3d50d60b339c62364473ccfe0fc/build/webpack.config.js#L97-L99

xem: ok, see https://github.com/Qwant/erdapfel/pull/1180

xem:
Thanks @bbecquet , I'm starting that in a separate MR

remi-dupre (approved):

PR #1178, opened 2021-08-17
Qmaps 1796 fix direction form pins

xem:
## Description

- make the form pins become blue/red only when the field content is a valid poi or location
- else, keep the icons light grey on blur, and dark grey on focus and when typing in the fields
- when editing a valid field, remove the blue/red color and return to dark grey

Nice side effects:
- no more icons blinking when the route is being computed
- everything is ok when clicking the "x" button and the "invert fields" button
- everything is ok when filling the fields by clicking on the map / moving a pin on the map to edit a field

### DIFF
--- a/src/panel/direction/DirectionInput.jsx
+++ b/src/panel/direction/DirectionInput.jsx
@@ -107,7 +107,7 @@ class DirectionInput extends React.Component {
               <input
                 ref={inputRef}
                 id={`direction-input_${pointType}`}
-                className={classnames({ valid: point !== null })}
+                className={classnames({ valid: !!point })}
                 type="search"
                 required
                 autoComplete="off"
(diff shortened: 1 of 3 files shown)


src/panel/direction/DirectionInput.jsx
```diff
@@ -106,6 +107,7 @@ class DirectionInput extends React.Component {
               <input
                 ref={inputRef}
                 id={`direction-input_${pointType}`}
+                className={classnames({ valid: point !== null })}
```
bbecquet, line 110: Just testing `point` for truthiness with `{ valid: point }` would be simpler and more robust, if for some reason it can become `undefined`. For example if the state management of the parent components change.

xem: How about:   !!point
Unnecessary?

bbecquet: Unnecessary because classnames checks the truthiness of the argument, it doesn't need a real boolean. But if you find it clearer, no problem for me as it will behave correctly with `undefined`.

bbecquet (approved):

PR #1179, opened 2021-08-18
QMAPS transifex update

xem:
tx pull


language/message/br.po
```diff
@@ -313,9 +313,20 @@ msgstr ""
 msgid "No"
 msgstr ""
 
+msgid "http://www.w3.org/2000/svg"
+msgid_plural "0 0 24 24"
+msgstr[0] ""
+msgstr[1] ""
+msgstr[2] ""
+msgstr[3] ""
+msgstr[4] ""
```
amatissart, line 322: This must be a false positive due to how `gettext` parses the bundle file. Weird but has no effect.

amatissart (approved):

language/message/br.po
xem: so can we merge or do you want me to remove it?

amatissart: That's ok. Let's merge this PR.  

Maybe we should set `_n` as a reserved name, in addition to `_` though:
https://github.com/Qwant/erdapfel/blob/1e4919d4592df3d50d60b339c62364473ccfe0fc/build/webpack.config.js#L97-L99
This could actually be related to #1177, in order to parse the new pluralized string correctly.

xem: sure, I did it in https://github.com/Qwant/erdapfel/pull/1180

PR #1181, opened 2021-08-24
QMAPS-434 show stars summary in POI header

xem:
## Description
If a PoI is starres, sdd "1 star" / "N stars" / "Starred" (if we don't know how many stars it has) in the PoI's header, next to the subclass, preceded by a median dot.

## Screenshots

image
image
image


src/panel/poi/blocks/StarsSummary.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+  if (block.ratings[0].nb_stars && block.ratings[0].nb_stars > 0) {
+    return (
+      <span>
+        {' · ' + capitalizeFirst(_n('%d star', '%d stars', block.ratings[0].nb_stars, 'poi'))}
+      </span>
+    );
+  }
+
+  return <span>{' · ' + capitalizeFirst(_('Starred', 'poi'))}</span>;
```
remi-dupre, line 20: I'm not a huge fan of how this block is prefixed with a middle dot, it makes the block dependent of the context in which it is displayed. I'd say that like in `StarsBlock` we'd rather pass the subclass as a parameter and build the whole line here.

src/panel/poi/blocks/StarsSummary.jsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export default StarsSummary;
```
remi-dupre, line 23: I'm not how I would do that but the structure of this code is quite redundant, maybe the string could be built from a specific function that would take a "separator" parameter which indicates if we want it to be full-text or with a middle dot. Do we know if this difference of behavior is actually willingly pushed by the design?

src/components/PoiItem.jsx
```diff
@@ -8,12 +8,15 @@ import Address from 'src/components/ui/Address';
 import classnames from 'classnames';
 import poiSubClass from 'src/mapbox/poi_subclass';
 import { capitalizeFirst } from 'src/libs/string';
+import { findBlock } from 'src/libs/pois';
+import StarsSummary from 'src/panel/poi/blocks/StarsSummary';
```
remi-dupre, line 12: Shouldn't `StarsSummary` be defined in `src/components`, like `OpeningHour` is for instance?

src/panel/poi/blocks/StarsBlock.jsx
```diff
@@ -30,7 +30,7 @@ const StarsBlock = ({ block, subclass }) => {
 
   return (
     <Block simple icon={<IconStar fill={ACTION_BLUE_BASE} width={20} />}>
-      {_('Starred {subclass}', 'poi', { subclass })}
+      {capitalizeFirst(_('Starred {subclass}', 'poi', { subclass }))}
```
remi-dupre, line 33: As I understand it, it will have no effect with the English string, but would enforce capitalization of the first word for other languages. Shouldn't we just expect the translations to be consistent with the capitalization of the reference?

remi-dupre (commented):
I have never had to spend a lot of time working with `erdafel`'s codebase, so don't hesitate to criticize my points :wink:

src/panel/poi/blocks/StarsSummary.jsx
xem: For the dot I realized after pushing that it should be displayed with css, fixing it

src/panel/poi/blocks/StarsBlock.jsx
xem: indeed, for this component, it can be handled entirely in the translations

src/panel/poi/blocks/StarsSummary.jsx
xem: edit: ah no, there's another mechanism for displaying the dot in the poi list, I'll use this one instead

src/panel/poi/blocks/StarsBlock.jsx
xem: In this precise string I belive we have to force the "capitalize first".
Ex: in french we will say "{subclass} étoilé", and in english "Starred {subclass}"
subclass is lowercase in every case.
So when subclass = "restaurant" for example,
we need a way to have: FR: "Restaurant étoilé" (with a R) / EN: "Starred restaurant" (with a r)
IMO only capitalizeFirst (or CSS tricks) can achieve this, but not the translation.

remi-dupre: I completely missed that case.

This is typically a case where I show that I'm a novice regarding this kind of issues. Nice catch!

src/components/PoiItem.jsx
```diff
@@ -28,8 +32,16 @@ const PoiItem = React.memo(
               <ReviewScore reviews={reviews} poi={poi} inList={inList} />
             </div>
           )}
-          <div className="poiItem-subclassAndHours">
-            <div className="poiItem-subclass">{subclass}</div>
+          <div className="poiItem-subclassStarsAndHours">
+            <span className="poiItem-subclass">{subclass}</span>
+            {!inList && subclass && hasStars && '\u00A0⋅\u00A0'}
```
bbecquet, line 37: Not your fault, but all these rules about list or not are too complicated…
Maybe it'll be worth splitting the list/popup card again, or we can try to convince the designers to always display everything the same way.

src/components/PoiItem.jsx
```diff
@@ -8,14 +8,18 @@ import Address from 'src/components/ui/Address';
...
+import { useI18n } from 'src/hooks';
 
 const PoiItem = React.memo(
   ({ poi, withOpeningHours, withImage, withAlternativeName, className, inList, ...rest }) => {
     const reviews = poi.blocksByType?.grades;
 
     const subclass = capitalizeFirst(poiSubClass(poi.subClassName));
-
+    const stars = findBlock(poi.blocks, 'stars');
+    const hasStars = stars?.ratings?.[0]?.has_stars === 'yes';
```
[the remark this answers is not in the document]
bbecquet: I've factorized some bits in the last commit, to re-use some parts of the block implementation and simplify PoiItem (the list and dot stuff is already complicated enough…)

bbecquet (approved):

PR #1186, opened 2021-09-06
QMAPS-2285 show old browser modal for iOS < 10 and adapt babel config

xem:
## Description
- show old browser modal on iOS < 10 (these OS have a Safari < 10 which doesn't support some of our JS/CSS)
- adapt babel config to safari >= 10 to save space

## Why
Avoid JS errors / non-working website on old devices without explanation

## Screenshots

image

### DIFF
--- a/bin/middlewares/unsupported_browser.js
+++ b/bin/middlewares/unsupported_browser.js
@@ -1,5 +1,20 @@
+function iOSversion(userAgent) {
+  if (/iP(hone|od|ad)/.test(userAgent)) {
+    const v = userAgent.match(/OS (\d+)_(\d+)_?(\d+)?/);
+    return [parseInt(v[1], 10), parseInt(v[2], 10), parseInt(v[3] || 0, 10)];
+  }
+}
+
 function isUnsupported(userAgent = '') {
-  return userAgent.match('Trident'); // old IE detection
+  // iOS < 10 detection
+  if (iOSversion(userAgent)[0] < 10) {
+    return true;
+  }
+
+  // old IE detection
+  if (userAgent.match('Trident')) {
+    return true;
+  }
 }
 
 module.exports = function (config) {
--- a/bin/middlewares/unsupported_browser.js
+++ b/bin/middlewares/unsupported_browser.js
@@ -1,5 +1,5 @@
 function isOldIE(userAgent) {
-  return userAgent.match('Trident'); // Test for IE <= 11
+  return userAgent.match('Trident') || false; // Test for IE <= 11
 }
 
 function isOldiOS(userAgent) {
@@ -9,12 +9,11 @@ function isOldiOS(userAgent) {
       return parseInt(v[1], 10) < 10; // test if major iOS version is < 10
     }
   }
+  return false;
 }
 
 function isUnsupported(userAgent = '') {
-  if (isOldIE(userAgent) || isOldiOS(userAgent)) {
-    return true;
-  }
+  return isOldIE(userAgent) || isOldiOS(userAgent);
 }
 
 module.exports = function (config) {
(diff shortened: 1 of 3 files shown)


bin/middlewares/unsupported_browser.js
```diff
@@ -1,5 +1,20 @@
+function iOSversion(userAgent) {
+  if (/iP(hone|od|ad)/.test(userAgent)) {
+    const v = userAgent.match(/OS (\d+)_(\d+)_?(\d+)?/);
+    return [parseInt(v[1], 10), parseInt(v[2], 10), parseInt(v[3] || 0, 10)];
```
bbecquet, line 4: Beware, this code is not very robust to strange UA strings, which isn't uncommon.
Not a big deal if we flag something as supported when it's not, but if the `.match(…)` fails, `v` will be `null` and the array parse will crash and the page will display a dirty error.
To check for that on local, you have to set the `TILEVIEW_server_unsupportedBrowsers_redirect` env var to `true`, like it's done in prod.

So, we should wrap this fragile code in a `try catch` at least… (but what should be returned?)

bin/middlewares/unsupported_browser.js
```diff
@@ -1,5 +1,20 @@
...
+  }
+}
+
 function isUnsupported(userAgent = '') {
-  return userAgent.match('Trident'); // old IE detection
+  // iOS < 10 detection
+  if (iOSversion(userAgent)[0] < 10) {
```
bbecquet, line 10: Instead of putting some low-level implementation detail here (the `[0]`), maybe you could rename the function `isOldSafari` and check this inside.
That way you know you can put `return false;` in your `try/catch` and this function can be rewritten more simplfy, without redundant comments
```js
function isUnsupported(userAgent = '') {
    return isOldSafari(userAgent) || isOldIE(userAgent);
}
```

bin/middlewares/unsupported_browser.js
xem: of course, I used this env var with the value "true" locally, I just didn't push it :) 
thanks for the review, I'm going to make my code less dirty

bin/middlewares/unsupported_browser.js
xem: hi, I made the code cleaner, but with no try/catch, tell me if you find it robust enough :)

bin/middlewares/unsupported_browser.js
```diff
@@ -1,10 +1,24 @@
...
+}
+
+function isOldiOS(userAgent) {
+  if (/iP(hone|od|ad)/.test(userAgent)) {
+    const v = userAgent.match(/OS (\d+)_(\d+)_?(\d+)?/);
+    if (v && v[1]) {
+      return parseInt(v[1], 10) < 10; // test if major iOS version is < 10
+    }
+  }
+}
```
bbecquet, line 12: Let's add a `return false;` here so it's clearer and doesn't just rely on the function returning `undefined`.

bin/middlewares/unsupported_browser.js
```diff
@@ -1,10 +1,24 @@
...
+
 function isUnsupported(userAgent = '') {
-  return userAgent.match('Trident'); // old IE detection
+  if (isOldIE(userAgent) || isOldiOS(userAgent)) {
```
bbecquet, line 15: I don`t get why you prefer writing
```js
if (expressionThatReturnsABool) {
   return true;
} 
// and once again, missing `return false;` here 
```
than just
```js
return expressionThatReturnsABool;
```

xem: sorry for the zig-zags, I still have issues to make choices between legibility and avoiding to write unnecessary code
for me the "return false" were unnecessary, and the "if(thing) return true" were much easier to read, but no prob, it's fixed :D

bbecquet (approved):

PR #1187, opened 2021-09-09
Qwant-ponents dependency integration and first use

bbecquet:
## Description
Install the newly published [`@qwant/qwant-ponents`](https://www.npmjs.com/package/@qwant/qwant-ponents) package as a project dependency and use it to replace a first component in our code by the common implementation.

Here is what I did in details, commit by commit:
1. install the `@qwant/qwant-ponents` dependency.

2. ensure the package JS files are transformed by Babel.
As the `@qwant/qwant-ponents` package publishes only raw source files and not built versions, we need to ensure they are built by Babel when imported in our code. The thing is, we normally don't build NPM packages (as it's long, and useless for most packages which are published as ES5). So we need to change the Babel config to include `@qwant/qwant-ponents`. There was already a similar exception for `@qwant/qwant-basic-gl-style` so I just configured Babel to transform all `@qwant` modules.

3. ensure the package CSS files are correcly built.
The way Sass/CSS is handled is different in `qwant-ponents` than in Erdapfel. Instead of building a big CSS file from Sass imports, CSS is imported as modules from JS files and bundled only if used.
This requires a new loader config in the Webpack config, using the JS as entry points.

4. use `<Alert>` from the qwant-ponents instead of our `<Alert>`.
To test the qwant-ponents and demonstrate they work, I replaced the `Alert ` component used in the user feedback questions by the one from the Design System. I chose this component as first cause it's used in one place only and our version was built as a copy of the one from the `qwant-ponents`, so it was easy to replace with minimal changes.
Other replacements (probably with more complicated changes required) will occur in their own PR.

5. clean the old `<Alert>` implementation.
Following the replacement, we can remove the previous implementation. This includes the JSX code, custom CSS and also an icon file. (Most UI icons are also parts of qwant-ponents so we'll be able to replace them in the future).

_Note: you may notice the main JS bundle size increasing **a lot** during development. This is because the Webpack tree-shaking doesn't happen in development, so the whole qwant-ponents code is built even if most of it isn't used._
_In production build, the tree-shaking works fine and the bundle size increases only slightly._


bbecquet:
> build/test causes no problems on Windows

Great news :+1:

PR #1190, opened 2021-09-15
update W10's installation doc

xem:
## Description
updated doc

### DIFF
--- a/docs/src/windows.md
+++ b/docs/src/windows.md
@@ -18,20 +18,6 @@ sudo apt update && sudo apt install -y build-essential git libexpat1-dev libssl-
 libncurses5-dev libbz2-dev liblzma-dev \
 libsqlite3-dev libffi-dev tcl-dev linux-headers-generic libgdbm-dev \
 libreadline-dev tk tk-dev
-curl https://pyenv.run | bash
-```
-
-- Edit ~/.bashrc (ex: ```nano ~/.bashrc```) and add this code at the end (replace {USERNAME} with your Ubuntu subsystem's session name):
-
-```
-export PATH="/home/{USERNAME}/.pyenv/bin:$PATH"
-eval "$(pyenv init -)"
-eval "$(pyenv virtualenv-init -)"
-```
-
-- Use updated .bashrc
-```
-source ~/.bashrc
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/windows.md
```diff
@@ -23,47 +22,37 @@ libreadline-dev tk tk-dev
...
+npm i webpack-cli -g
+```
+
 - Install Python 3 and its libraries:
 
 ```
 pyenv install 3.7.2
 pyenv global 3.7.2
 pip install pip --upgrade
+sudo apt install python3-pip
```
bbecquet, line 46: I don't remember, why is Python needed for Erdapfel?

docs/src/windows.md
```diff
@@ -23,47 +22,37 @@ libreadline-dev tk tk-dev
...
 ```
 export PATH="/home/{USERNAME}/.pyenv/bin:$PATH"
 eval "$(pyenv init -)"
 eval "$(pyenv virtualenv-init -)"
 ```
 
+- Use updated .bashrc and install webpack
+```
+source ~/.bashrc
+npm i webpack -g
```
bbecquet, line 36: Why do you need webpack(-cli) as global installs?
They are installed locally in Erdapfel by npm, this should be enough.

docs/src/windows.md
```diff
@@ -77,6 +66,7 @@ export TILEVIEW_category_enabled=true
 export TILEVIEW_events_enabled=false
 ```
 
+- Delete package-lock.json
```
bbecquet, line 69: Why do you need to delete it?

docs/src/windows.md
```diff
@@ -13,6 +11,7 @@ git clone https://github.com/Qwant/erdapfel.git
 ```
 
 - Open a Ubuntu bash in this folder (shift + right-click > Open Linux CLI here)
+- Install [nvm](https://github.com/nvm-sh/nvm#install--update-script), then node 12 (`nvm install 12` + `nvm use 12`), then update npm (`npm install -g npm@latest`)
```
bbecquet, line 14: For now we rely on npm 7, so it may be safer to specify `npm@7` instead of `npm@latest`.

bbecquet (approved):

PR #1192, opened 2021-09-15
Update windows.md

xem:
## Description
Update Windows doc

### DIFF
--- a/docs/src/windows.md
+++ b/docs/src/windows.md
@@ -18,6 +18,7 @@ sudo apt update && sudo apt install -y build-essential git libexpat1-dev libssl-
 libncurses5-dev libbz2-dev liblzma-dev \
 libsqlite3-dev libffi-dev tcl-dev linux-headers-generic libgdbm-dev \
 libreadline-dev tk tk-dev
+sudo apt-get install gettext
 ```
 
 - Then enter the [commands](#commands) below to build and run Erdapfel


docs/src/install and build.md
```diff
@@ -1,31 +1,28 @@
...
 
-- Open a Ubuntu bash in this folder (shift + right-click > Open Linux CLI here)
-- Install [nvm](https://github.com/nvm-sh/nvm#install--update-script), then node 12 (`nvm install 12` + `nvm use 12`), then update npm (`npm install -g npm@7`)
-- Install tools, libraries:
+- Open a terminal in the "erdapfel" folder (on Windows: shift + right-click > Open Linux CLI here)
+- Install [nvm](https://github.com/nvm-sh/nvm#install--update-script), then node 12 (`nvm install 12` + `nvm use 12`), then npm 7 (`npm install -g npm@7`)
+- Install the project's tools and libraries:
 
 ```
 sudo apt update && sudo apt install -y build-essential git libexpat1-dev libssl-dev zlib1g-dev \
```
bbecquet, line 17: I don't know why it's there, and as a Linux or iOS web developer who is a new comer to the project, I would definitely not do that without knowing how it is useful to a "simple" NodeJS project.
Do you remember what it's for on W10?

bbecquet (commented):
I don't think this doc is useful anymore at all under this form…
For example, does a web dev project hosted on GitHub explain how to install Git?

And for the other stuff (env vars, building/running), most of it is already written in README.md. I agree it can be clarified, but we could probably improve the main file with simple bits instead of duplicating stuff here and making things more confusing.

For the Windows specificity for example, a simple line to mention WSL can be added, as everything else is common now.

docs/src/install and build.md
xem: no - it's possible that they're just old requirements for the Erdapfel from 2018.
So, if wsl is the only specific thing to do, I agree we can remove this file and add a little word about it in the main readme.

bbecquet (approved):

PR #1196, opened 2021-09-21
Display subclass name in top bar for nameless POIs

bbecquet:
## Description
Fix a bug where, for nameless POIs (now clickable), the topbar isn't filled with anything, so it doesn't display the cross button to close the POI panel, making it impossible to navigate back with the mouse.
For these POIs, display their subclass name in the bar.

## Screenshots
|Before|After|
|---|---|
|Capture d’écran de 2021-09-21 14-32-48|Capture d’écran de 2021-09-21 14-32-28|


remi-dupre (approved):

xem (changes requested):
Nice but the tests don't pass anymore?

xem (approved):

PR #1199, opened 2021-09-21
Qmaps 2290 save search history

xem:
## Description
- Use the feature flag searchHistoryConfig.enabled
- save last 100 queries in local storage, ordered by date (last item of the list is the most recent)
- if an existing query is searched again, it's put at the end of the list
- queries are saved by name if we type a valid query and press enter, or select a suggest.
- queries are not saved if no results were found

## Screenshots

Example: LocalStorage after searching Nice, Lyon, Lyon 3eme, Negresco and Lyon again. Only one occurrence of "Lyon" is visible at the end.

image

### DIFF
--- a/src/adapters/search_history.js
+++ b/src/adapters/search_history.js
@@ -27,9 +27,10 @@ export function saveQuery(q) {
 export function deleteQuery(q) {
   const searchHistory = get(SEARCH_HISTORY_KEY) || [];
   const index = searchHistory.indexOf(q);
-  if (index > -1) {
-    searchHistory.splice(index, 1);
+  if (index === -1) {
+    return;
   }
+  searchHistory.splice(index, 1);
   // Serialize the list and save it in localStorage
   set(SEARCH_HISTORY_KEY, searchHistory);
 }
--- a/src/adapters/search_history.js
+++ b/src/adapters/search_history.js
@@ -1,0 +1,40 @@
+import { get, set, del } from './store';
+
+const search_history = get('search_history') || [];
+
+// Add a query to the list
+export function save_query(q) {
+  // Delete query if it's already in the list
+  if (search_history.includes(q)) {
+    delete_query(q);
+  }
+
+  // Put the query at the end of the array
+  search_history.push(q);
+
+  // Limit the list to the 100 last items
+  if (search_history.length > 100) {
+    search_history.unshift();
+  }
+
+  // Serialize the list and save it in localStorage
+  set('search_history', search_history);
+  console.log(search_history);
+}
+
+// Delete a query from the list
+export function delete_query(q) {
+  const index = search_history.indexOf(q);
+  if (index > -1) {
+    search_history.splice(index, 1);
+  }
+  // Serialize the list and save it in localStorage
+  set('search_history', search_history);
+  console.log(search_history);
+}
+
+// Delete the whole search history
+export function delete_search_history() {
+  del('queries_list');
+  search_history = [];
+}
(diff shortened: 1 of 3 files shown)


src/adapters/search_history.js
```diff
@@ -0,0 +1,38 @@
...
+  // Delete query if it's already in the list
+  if (search_history.includes(q)) {
+    delete_query(q);
+  }
+
+  // Put the query at the end of the array
+  search_history.push(q);
+
+  // Limit the list to the 100 last items
+  if (search_history.length > 100) {
```
bbecquet, line 16: Let's move this 100 in a constant.

xem:
@bbecquet comme discuté ce matin avec @amatissart , stocker place.name ne sera pas suffisant pour se rendre au bon endroit sans ambiguïté lorsqu'on cliquera sur l'item en question.
Il s'agirait de stocker en même temps que le nom, soit un id du truc recherché (mais est-ce que tous les trucs ont un id? poi, ville, intention, autre?) ou bien directement l'url cible (ce qui me semble le plus simple).

Ex:
- je tape Nice puis presse la touche Entrée
- le site ajoute dans localStorage une entrée qui pourrait ressembler à :  `{ name: "Nice (06000-06300)", path: "/place/admin:osm:relation:170100@Nice_06000-06300" }`

Merci de nous dire ton avis :)

bbecquet:
@xem ok to store something else to reduce ambiguity, but I don't think we should use the url.
It doesn't match the case of address fields used in the direction panel, which shouldn't redirect to the POI panel. Also, it creates an unecessary coupling between parts of the app (now the suggest feature has to know about the nav structure of the app), and it's not future-proof as it's serialized and hard to change.

Actually, I think the most logical and generic thing to store is the POI itself. That way the history system will be really close to the favorite feature and can re-use most of its bits (serialization, display name function, etc.).

src/adapters/search_history.js
```diff
@@ -0,0 +1,40 @@
+import { get, set, del } from './store';
+
+const SEARCH_HISTORY_KEY = 'search_history';
+const HISTORY_SIZE = 100;
+
+// Add a query to the list
+export function save_query(q) {
```
bbecquet, line 7: This is still snake case… and I know why I thought we had a camelCase rule for ESLint, it's in Phoenix.
I'll take a note to add the same.

src/adapters/search_history.js
```diff
@@ -0,0 +1,40 @@
...
+
+  // Serialize the list and save it in localStorage
+  set(SEARCH_HISTORY_KEY, searchHistory);
+}
+
+// Delete a query from the list
+export function delete_query(q) {
+  const searchHistory = get(SEARCH_HISTORY_KEY) || [];
+  const index = searchHistory.indexOf(q);
+  if (index > -1) {
```
bbecquet, line 30: It would be better to early return on `-1` here, so we don't overwrite the history if it didn't change.

src/adapters/search_history.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // Retrieve the search history
+  const searchHistory = get(SEARCH_HISTORY_KEY) || [];
+
+  // Put the query at the end of the array
+  searchHistory.push(q);
+
+  // Limit the list to the 100 last items
+  if (searchHistory.length > HISTORY_SIZE) {
+    searchHistory.shift();
```
bbecquet, line 19: `shift` is a bit fragile, as it won't react properly if we happen to reduce HISTORY_SIZE.
We can get the n last elements of the array with a negative argument passed to `slice`, which will work in all cases.

bbecquet (approved):

xem:
Nice additions and remkarks!
I noted that intentions are skipped for now. 👍

PR #1202, opened 2021-09-29
Qmaps 2289 history in directions

xem:
## Description
- Forked @bbecquet 's branch QMAPS-1689
- Discussed with the team that Direction's history should actually be shared with TopBar's history (at least for now, let's see if it's ok in plive)
- Converted DirectionInput to a Functional React Component (pretty sure I didn't break anything)
- Enabled history display inside Direction's suggests
- TODO: save the queries typed/selected inside the directions inputs inside the global history (as we do in TopBar)
- TODO: check/finalize style once Benjamin's branch is merged.
- TODO (@bbecquet) : we should probably find a way to display both history and favorites items inside the suggests,. Currently, if an item is in favorites but not in the history (it can happen), it is not listed at all. It could be present before or after the history's items, no?

## Screenshots
image


bbecquet:
@xem this is dangerous to mix a non-trivial refacto (class to functional) with the feature development. This component has tricky focus side-effects, and we must be extra sure nothing breaks.
At the very least it should be done in two separate commits, but better still, it would make a great separate PR now that you want to tackle the subject :)

PR #1203, opened 2021-09-30
[QMAPS 1689] History suggestions in main search input

bbecquet:
## Description
Display suggestions from the history in the main search, with different rules depending on mobile/desktop and typed string. Also affect some stuff about suggested favorites (number and one-line style)

If there are adjustments to make with these rules or other fine-tuning to do, as we are behind a feature flag I suggest we do it in further tickets and move on with this one if the overall logic is ok.
For example, there are questions about history items that are also storred in favorites. This can be addressed later.

## Screenshots
|With no input|With input matching history|
|---|---|
|Capture d’écran de 2021-09-30 12-22-50|Capture d’écran de 2021-09-30 12-24-14|


xem (approved):
Reviewed it yesterday before the PR and it was fine for me!
Ok to discuss discuss the details like favorites and intentions in another context, and merge this for now

PR #1206, opened 2021-10-04
Use <Button> from the qwant-ponents

bbecquet:
## Description
Replace our custom implementation of `<Button>` by a direct import from the design system.
Mostly direct conversion, with some props or style adjustements sometimes.
Where it was straightforward, I also replaced some button icons by those of the qwant-ponents.

## Why
 - Same UX and API than the other Qwant projects
 - Less code to maintain


xem (approved):
Code: it's a nice cleanup!
All the buttons I tested locally seem to work as well as before. :)

bbecquet:
@xem thanks for the approval!
I didn't check yesterday after pushing but there are some failing tests, probably due to changes in selectors… I'll check that.

xem (approved):

PR #1207, opened 2021-10-05
QMAPS hide the poi present in the other field from Direction fields' suggest

xem:
## Description
- Tell each field which value is present in the other field
- Exclude this value from the current field's suggest (especially the history suggest)
- It's a proposition, and of course it can be debated

## Screenshots

Before;
image

After:
image


bbecquet (approved):
Nice :+1:

PR #1218, opened 2021-10-21
Use external lib for labelling alternative routes

bbecquet:
## Description
Replace the internal implementation of the alternative route labeller by a dependency to https://www.npmjs.com/package/alt-route-labeller (which is the same algo anyway :))

## Why
Less code to maintain!


bbecquet:
@fatal69100 Thanks for your review!
I had actually forgotten to push the latest version using v0.3… Great you noticed it :)
The 0.3 is better as it can work directly on an array of geometries, removing the need for the `featureCollection` transformation, which can be costly on long routes.

PR #1228, opened 2021-11-02
QMAPS-2368 favorites when focusing empty field with no history

xem:
## Description
- display up to 2 favorites when focusing an empty field if no history items are available
- if history is enabled and not empty, then only the history items are displayed

## Screenshots

image


fatal69100 (approved):
LGTM

remi-dupre (approved):

PR #1234, opened 2021-11-29
Qmaps 2373 / 2374 history modals

xem:
## Description
Show confirmation modals before clearing or disabling local search history.
Both use the same component 
Disabling the history also provokes clearing the history
Desktop / Mobile designs are different (buttons stacked on mobile)
Also in this MR: updated the caniuse dependency

## Screenshots
Modal 1 (seen on desktop):
image

Modal 2 (seen on mobile):
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5240,9 +5240,9 @@
       }
     },
     "node_modules/caniuse-lite": {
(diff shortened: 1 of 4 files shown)


fatal69100 (approved):
LGTM

PR #1237, opened 2021-12-08
QMAPS-2356 show history prompt in suggest if feature flag enabled and prompt not yet answered

xem:
## Description
- ensure suggest doesn't close when the mouse clicks outside of the window (not asked in the ticket but very convenient for debug and UX)
-  when the search field is focused and empty, show history prompt panel instead of the suggest if the history feature flag is enabled and is this prompt is not yet answered
- TODO (in next MRs): make buttons clickable, provide link URL, final design

## Why

## Screenshots
image

### DIFF
--- a/src/adapters/search_history.js
+++ b/src/adapters/search_history.js
@@ -14,6 +14,14 @@ export function getHistoryEnabled() {
   return get(SEARCH_HISTORY_KEY + '_enabled');
 }
 
+export function setHistoryPrompt(value) {
+  set(SEARCH_HISTORY_KEY + '_prompt', value);
+}
+
+export function getHistoryPrompt() {
+  return get(SEARCH_HISTORY_KEY + '_prompt');
+}
+
 export function getHistory() {
   return get(SEARCH_HISTORY_KEY) || [];
 }
--- a/src/components/TopBar.jsx
+++ b/src/components/TopBar.jsx
@@ -143,8 +143,11 @@ const TopBar = ({ value, setUserInputValue, inputRef, onSuggestToggle, backButto
                   onFocus();
                 }}
                 onBlur={() => {
-                  setFocused(false);
-                  onBlur();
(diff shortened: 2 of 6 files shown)


src/components/TopBar.jsx
```diff
@@ -143,8 +143,11 @@ const TopBar = ({ value, setUserInputValue, inputRef, onSuggestToggle, backButto
                   onFocus();
                 }}
                 onBlur={() => {
-                  setFocused(false);
-                  onBlur();
+                  // The mouseLeave flag allows to keep the suggest open when clicking outside of the browser
```
fatal69100, line 146: :+1:

src/adapters/search_history.js
```diff
@@ -14,6 +14,14 @@ export function getHistoryEnabled() {
   return get(SEARCH_HISTORY_KEY + '_enabled');
 }
 
+export function setHistoryPrompt(value) {
```
fatal69100, line 17: This will probably be used on the next PR when the button will be clickable, but it always weirds to have code unused in a PR

fatal69100 (approved):

src/panel/app_panel.js
```diff
@@ -35,6 +35,15 @@ export default class App {
...
     );
+
+    // The mouseLeave flag allows to keep the suggest open when clicking outside of the browser
+    window.mouseLeave = false;
+    window.document.onmouseleave = () => {
+      window.mouseLeave = true;
+    };
+    window.document.onmouseenter = () => {
+      window.mouseLeave = false;
+    };
```
remi-dupre, line 46: On my computer it only appears to work on Chromium but no on Firefox.

This could be specific to i3 which handle window focus slightly differently, does it work on other setups?

xem: nice catch, fixed!

remi-dupre (approved):

PR #1238, opened 2021-12-13
QMAPS fix js for safari < 14

xem:
## Description
Rewrite the mobile detection using older JS code (window.onresize instead of media query matching event)

## Why
safari < 14 bugged


fatal69100 (approved):

remi-dupre (approved):
:+1:

PR #1240, opened 2021-12-13
QMAPS-2357, QMAPS-2358, QMAPS-2295 history in suggest

xem:
## Description

if the history feature flag is enabled:

- show history prompt in suggest, on focus, under the topBar field, if it hasn't been answered yet and if the field is empty (hide the favorites in that case)
- keep it visible until yes or no has been clicked, because otherwise everything disappears before the click is registered properly. 
- show the accept / refuse messages and enable / don't enable the history after clicking the yes / no buttons
- don't show the message again after filling and emptying the field or after navigating the app and focusing the field again
- show a message in the suggest footer if the field is focused and empty and if history items are present.
- include mobile/pc info in telemetry
- TODO: design
- TODO: final links (for history page and help center)

## Screenshots
image
image
image
image

### DIFF
--- a/src/components/ui/Suggest.jsx
+++ b/src/components/ui/Suggest.jsx
@@ -2,10 +2,12 @@ import React, { useEffect, useState, useCallback, useRef } from 'react';
 import ReactDOM from 'react-dom';
 import debounce from 'lodash.debounce';
 import { bool, string, func, object } from 'prop-types';
-import { useDevice, useI18n } from 'src/hooks';
+import { useConfig, useDevice, useI18n } from 'src/hooks';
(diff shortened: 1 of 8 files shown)


src/components/ui/Suggest.jsx
```diff
@@ -178,26 +285,47 @@ const Suggest = ({
         },
         highlightedValue: highlighted ? getInputValue(highlighted) : null,
       })}
-      {dropdownVisible &&
+      {(dropdownVisible || historyPromptVisible) &&
```
fatal69100, line 288: Is `historyPromptVisible` a necessary condition to render the next Portal ?

fatal69100 (approved):

PR #1243, opened 2022-01-14
QMAPS-2376 tripadvisor single

Grafikart:
## Description
Add tripadvisor review for the source on details view for a POI


xem:
Thank you for this PR.
I could review the code (great) and test locally:

- PoI: http://localhost:3000/place/ta:poi:3350685@Garnie_Pizza
image

As you said, the TA and OSM icons are not well aligned. Otherwise, it's ok!

- Search: I'm not sure if it's the right one? http://localhost:3000/places/?q=hotel&bbox=9.4779912%2C47.1030974%2C9.6354047%2C47.2166679#map=11.44/47.1599130/9.5566980
image

xem (approved):

PR #1249, opened 2022-01-26
QMAPS-2329 feat: reservation module

Grafikart:
## Description

Add a reservation module for tripadvisor hotel

## Why

Why not ;)

## Screenshots

image


xem:
Hi @Grafikart , thanks for the PR. Much progress!

Here are the last details that need to be fixed, as discussed on Mattermost:

- add spellcheck="false" on the fields to avoid these wavy lines on focus on Chrome
image

- this problem may also be present in the IA

- the date picker is not fully visible on 13" desktops. As seen with Design, it can be moved up while they're thinking of a smaller version.
image

- Use the following labels: "18 years old and more", "Between 0 and 17 years old", "Age of child N"

Besides that it looks great, thanks.

xem (approved):

PR #1252, opened 2022-01-31
QMAPS-2325 add history logs

xem:
## Description
Add 7 logs related to History

(the logs requests are visible in the network panel of the browser, named "events")

image


remi-dupre (approved):
lgtm

PR #1259, opened 2022-02-09
Qmaps 2394 dedup intentions

xem:
## Description
Searches containing:
- Known category + place  (ex: "hotel nice")
- Known category "around this position" (ex: "hotel")
- Unknown category + place (ex: "macdo nice")
Must be present only once in the history panel, even if the search is performed many times

image

NB: This PR also fixes a React key duplication error in the History Panel. 
NB2: until now, the dedup test was looking for the "fullTextQuery" item, but this item has been removed, and afaik, it was not necessary to check the equality between two queries.

### DIFF
--- a/src/adapters/search_history.js
+++ b/src/adapters/search_history.js
@@ -56,7 +56,7 @@ export function saveQuery(item) {
 export function deleteQuery(item) {
   const searchHistory = getHistory();
   let index;
-  for (index = searchHistory.length - 1; index > 0; index--) {
+  for (index = searchHistory.length - 1; index >= 0; index--) {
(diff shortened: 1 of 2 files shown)


remi-dupre (commented):
Currently the `deleteQuery` function assumes there is at most one duplicate. While this is true in a normal use it may be more robust to remove all stored elements that match the new one? (For example, even though this was not in production yet, changes similar to this one that change the equality criteria may lead to having several duplicates of a same item in the history)

xem:
@remi-dupre I agree, but can it be considered as a future nice-to-have?

remi-dupre (approved):

PR #1266, opened 2022-02-13
QMAPS-2329 fix datepicker spacer

Grafikart:
## Description
Fix the datepicker actions at the bottom of the screen on mobile

## Why
The Datepicker is supposed to take 100vh in height but since it's constrained by an already fixed .panel the actions are pushed out of the screen. 

Improves Issue QMAPS-2329


pascaloliv:
Ticket ended on this PR : https://github.com/Qwant/erdapfel/pull/1292

PR #1267, opened 2022-02-14
Clean config

fatal69100:
## Description
remove `config.services.geocoder.url` and `config.server.services.idunn.url` config and add `path` module in nodejs to concatenate path

## Why
There are too much duplication configuration that make it less maintainable


fatal69100:
I didn't explore why tests are failing

PR #1272, opened 2022-02-16
Remove Transifex step build

fatal69100:
## Description
Remove push build step that send values string to translate to Transifex

## Why
We are not subscribing to Transifex anymore


xem:
ok for me but are you sure this line needs to be removed? 
sudo apt-get install -yq python3-pip python3-setuptools python3-wheel

xem (approved):

PR #1279, opened 2022-02-18
QMAPS-2509 - Local history - Fix navigation on an intention without bbox

pascaloliv:
## Description
This PR create a valid navigation URL - even if there isn't a bbox in the stored history item.

## Why
If the user search "Parc" and pick the first result (Nearest)) - then  - on Local Search History panel - there was a blocking js exception if you try to click on "Parc" (Nearest) because the intention stored in the localStorage does not contains a bbox.


xem (dismissed):
This fix is ok! I open a new ticket for the missing category translations in the history panel

xem (approved):

PR #1280, opened 2022-02-18
QMAPS-2428 - Tripadvisor / Pagesjaunes in history

pascaloliv:
## Description
Add TripAdvisor and PagesJaunes POI to the history.
Add POI category to the history too.

### DIFF
--- a/language/message/fr.po
+++ b/language/message/fr.po
@@ -1501,7 +1501,7 @@ msgstr "Afficher les favoris"
 #: public/build/javascript/bundle.js:99757
 #: public/build/javascript/bundle.js:100002
 msgid "My history"
-msgstr "Mon historique de rechercche"
+msgstr "Mon historique de recherche"
 
 #: public/build/javascript/bundle.js:99758
 msgid "Show history"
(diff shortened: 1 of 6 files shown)


xem:
To be discussed with Charlotte.

During my tests on this branch, I remarked that the following elements were added in the history:
- intention (ex: parc / parc nice / hotel / hotel nice / restaurant / restaurant nice)
- TripAd POI (ex: one hotel from a poi list)
- PJ POI (ex: a restaurant from a poi list)
- OSM PoI (ex: type"tour eiffel" or "societe generale nice musiciens" or "hotel negresco nice" in the search field)

And these elements were not:
- LatLon POI (gps coordinates if clicking anywhere on the map)
- OSM poi (if it is clicked on the map)


Other point to discuss wwith Charlotte:
Should we show intentions from the history in the suggest? I think we should not. See example below:

image

xem (dismissed):
approving for now, but to discuss before release

xem:
image
i think the "nearby" should not appear in the suggest under a category name

xem (approved):

PR #1282, opened 2022-02-22
QMAPS-2479 fix age list

xem:
## Description
- I don't reproduce the Firefox issue anymore (white dropdown)
- This MR includes a fix of the ages list (0-17 instead of 0-16) + no label for the age 0 + correct default param for Field (it caused a React warning)

NB: I'm gonna remove the empty value in the <Field type=select> in the design system. The reservation form is the only one that uses it


pascaloliv (changes requested):
Could you add `setInitialLoading` in useEffect dep array in `Reservation/Reservation.jsx` to avoid warning.

pascaloliv (approved):

PR #1286, opened 2022-02-24
QMAPS-2513 - Deprecate br and ca langs

pascaloliv:
## Description
This PR deprecates BR and CA languages,
removing the unnecessary br.po and ca.po files and compiles thoses langs code with their fallback locale file (respectively fr.po and es.po).


xem (approved):
Nice! with these changes, can you now compile the project on Mac OS without editing constants.yml?

pascaloliv:
> Nice! with these changes, can you now compile the project on Mac OS without editing constants.yml?

Yes it is. I'm not 100% sure we will keep `.po` files for i18n forever but at least, it is working like a charm now.

PR #1287, opened 2022-02-24
QMAPS-2520 - Viewport body height miscalculation with 100vh

pascaloliv:
## Description
Buggy behaviour when using Qmaps on Safari - focusing the top bar and focusing out 

## Screenshots
Issue: 
https://user-images.githubusercontent.com/442681/155574553-aefa5809-4152-4ec5-a398-caf9830128bc.mov

### DIFF
--- a/src/scss/includes/main.scss
+++ b/src/scss/includes/main.scss
@@ -1,5 +1,7 @@
 body {
-  background: url("../images/backgrounds/grid.png");
+  background: url('../images/backgrounds/grid.png');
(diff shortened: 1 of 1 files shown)


xem (approved):
nice, I wonder if this fix should also be present in qwant-ponents (and Phoenix) ?

pascaloliv:
> nice, I wonder if this fix should also be present in qwant-ponents (and Phoenix) ?

This `body` rule should live in Qwantcom & Maps repositories - but not in qwant-ponents in my opinion - which should be project agnostic.

PR #1318, opened 2022-03-25
QMAPS-2623 - Add zustand + refactor Menu and Products drawers logic

pascaloliv:
## Description
- Implement zustand, zustand-persist, zustand dev tools (to log in Redux dev tools)
- Add /store folder to create zustand store and split logic in slices (different files for ui, result, history, favs, etc..) to be scalable in the future.
- Refactor Menu and top bar a bit to use zustand based open/close state. 
- Add qwant-ponents d.ts types on-the-fly (not the best solution but qwant-ponents don't have their own TS types yet)


fatal69100:
I am not sure, I can review correctly the pr due to lack of knowledge :s it would be better to wait xem's review ?

xem (dismissed):

xem:
Hi,
I tested locally and it's fine!
(though, npm run start shows a new warning, but I don't think it's related to this PR)
image

The code is OK too except useStore(store=>store) that can be simplified in useStore()

PR #1319, opened 2022-03-28
QMAPS-1739 - Lazy loading of POI image

pascaloliv:
## Description
Add lazy loading on image in Poi list item + refactor a bit styling + component structure and add TS typings.

## Why
At the moment - we preload all images without specifying a loading strategy.
By default it's `eager` and it is not appropriated.

## Screenshots
https://user-images.githubusercontent.com/442681/160407152-7b4a790e-ba16-44be-a127-e90f86e49554.mp4


xem (approved):
great. thanks for the demo too

PR #1321, opened 2022-03-30
Fix visual regression on top-bar input width

pascaloliv:
## Description
Because input should be full width.

## Screenshots
Before
image
After
image

### DIFF
--- a/src/scss/includes/menu.scss
+++ b/src/scss/includes/menu.scss
@@ -190,6 +190,10 @@ $productDrawerWidth: 744px;
 }
 
 @media (min-width: 641px) {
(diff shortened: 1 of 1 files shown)


xem (approved):

xem:
the tests pass now? cool!

xem (approved):

pascaloliv:
> the tests pass now? cool!

The failing test was catching a regression (which is very good).
The `#react_menu__container` is containing desktop buttons too and they weren't displayed.
I switched from `display: none` to `width: 0` (because desktop buttons are `position: fixed`)

PR #1343, opened 2022-05-28
Add "only_osm" param from Localstorage value

fatal69100:
## Description
If "qmaps_v1_only_osm" value exist in the localstorage, add a parameter in the url to get only osm results for category and intention

## Why
https://github.com/Qwant/qwantmaps/issues/176 and back end dev team agree


xem (approved):
ok, and happy samedi soir :p

pascaloliv (dismissed):

PR #1367, opened 2022-07-08
UI Directions results revamp

pascaloliv:
## Description
- Itinerary results revamp according to Q rebranding


xem (approved):
Code + test locally OK :)

PR #1368, opened 2022-07-12
geocoder.js: fix intention field which was renamed from intentions

remi-dupre:
## Description
The `intentions` list field was renamed into a simpler `intention` value in a recent [refacto of Idunn](https://github.com/Qwant/idunn/pull/333). This led erdapfel to not find a value for this field an fallback to the regex mode.

## Screenshots
| master | HEAD |
|-|-|
|Screen Shot 2022-07-12 at 08 30 44|Screen Shot 2022-07-12 at 08 30 56|

### DIFF
--- a/src/adapters/geocoder.js
+++ b/src/adapters/geocoder.js
@@ -71,7 +71,7 @@ export function getGeocoderSuggestions(term, { focus = {}, useNlu = false } = {}
     }
     suggestsPromise = ajax.get(geocoderUrl, query);
     suggestsPromise
-      .then(({ features, intentions }) => {
+      .then(({ features, intention }) => {
         const pois = features.map((feature, index) => {
           const queryContext = new QueryContext(
             term,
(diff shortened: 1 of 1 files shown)


src/adapters/geocoder.js
```diff
@@ -82,10 +82,11 @@ export function getGeocoderSuggestions(term, { focus = {}, useNlu = false } = {}
           return new BragiPoi(feature, queryContext);
         });
         const bragiResponse = { pois };
-        if (intentions) {
-          bragiResponse.intentions = intentions
-            .map(intention => new Intention(intention))
-            .filter(intention => intention.isValid());
+        if (intention) {
+          const parsed = new Intention(intention);
```
remi-dupre, line 86: BTW I tried to write `const intention = new Intention(intention);`, while I wasn't expecting Javascript to support shadowing, I'm quite surprised it just made the function abort silently, without a warning or crashing the application. Is this the expected behavior, even though I built using the `--mode=development` parameter?

pascaloliv (approved):

PR #1369, opened 2022-07-12
poedit: add xgettext flag `--no-location`

remi-dupre:
Since the bundled JS now contains an hash of the file, its name changes
at every commit which leads to unnecessary long diffs if we specify the
location of a string in the .po files.


language/message/de.po
```diff
@@ -1,2247 +1,1646 @@
...
+#~ msgstr "Bewertungen"
+
+#~ msgid "TripAdvisor user"
+#~ msgstr "TripAdvisor-Nutzer"
+
+#~ msgid "Read the full review"
+#~ msgstr "Komplette Bewertung lesen"
+
+#~ msgid "Read more reviews on TripAdvisor"
+#~ msgstr "Weitere Bewertungen bei TripAdvisor lesen"
```
remi-dupre, line 1646: Not sure why these ones got commented?

Feels like it will probably need more extensive testing before we merge this :sob:

pascaloliv (approved):

PR #1370, opened 2022-07-12
Stringified instructions for walk sections

remi-dupre:
## Description
Add stringified instructions when they are not provided for walking steps on public transport directions.

:warning: This PR is based on #1369 in order to get a shorter diff, make sure it is merged before.

| master | HEAD |
|-|-|
| Screen Shot 2022-07-12 at 14 20 35 | Screen Shot 2022-07-12 at 14 18 13 |

## Languages

 - [x] English
 - [x] French
 - [x] German
 - [x] Italian
 - [x] Spanish

## Bummers

First, in french translations this naïve approach will get us "Tournez à doite sur Avenue de la Paix" instead of "Tournez à droite sur l'Avenue de la Paix". We could maybe try to be smarter but I think we will either have to cover a large range of specific cases or make something more generic that would expose us to even weirder edge cases.

Second, on some cases there are several consecutive instructions with no road name, which lead to show several consecutive lines of "Walk" to the user. However I don't think it is a topic for this PR, the API should probably be responsible of merging consecutive irrelevant parts.

### DIFF
--- a/src/panel/direction/RoutesList/Route/RoadMap/WalkLeg/index.jsx
+++ b/src/panel/direction/RoutesList/Route/RoadMap/WalkLeg/index.jsx
@@ -7,6 +7,26 @@ import LegLine from '../LegLine';
 import classnames from 'classnames';
 import { Chevron } from 'src/components/ui';
 
+const stringifyManeuver = maneuver => {
+  const stringifyModifier = {
+    'sharp left': _('Turn left', 'direction'),
+    left: _('Turn left', 'direction'),
+    'slight left': _('Keep left', 'direction'),
+    straight: _('Walk', 'direction'),
+    'slight right': _('Keep right', 'direction'),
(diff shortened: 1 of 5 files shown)


src/panel/direction/RoutesList/Route/RoadMap/WalkLeg/index.jsx
```diff
@@ -7,6 +7,26 @@ import LegLine from '../LegLine';
 import classnames from 'classnames';
 import { Chevron } from 'src/components/ui';
 
+const stringifyManeuver = maneuver => {
+  const stringifyModifier = {
+    'sharp left': _('Turn left', 'direction'),
```
pascaloliv, line 12: We will grab the object keys from Directions appTypes enum.
So the prerequisite is to port this one in TSX.

pascaloliv (approved):

PR #1374, opened 2022-07-18
QMAPS-2749 save preferred itinerary vehicle in localStorage

xem:
## Description

- first visit: default directions vehicle is car (driving)
- if the user changes vehicle when they compute their route, the last selected vehicle is saved in localStorage
- next visit: default vehicle is the one present in localStorage
- if the user visits an url that contains "?mode={vehicle}", then this vehicle is used instead of the value present in localStorage.

image

### DIFF
--- a/src/store/slices/ui.ts
+++ b/src/store/slices/ui.ts
@@ -21,7 +21,7 @@ export const createUiSlice = (set: NamedSet<AppState>, get: GetState<AppState>):
   isMenuDrawerOpen: false,
   isProductsDrawerOpen: false,
   isSearchInputTyping: false,
-  defaultVehicle: modes.WALKING,
+  defaultVehicle: modes.DRIVING,
   setMenuDrawerOpen: isOpen =>
     set(
       () => {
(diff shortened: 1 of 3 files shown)


src/store/slices/ui.ts
```diff
@@ -3,21 +3,25 @@ import { onDrawerChange } from 'src/libs/url_utils';
...
   setProductsDrawerOpen: (isOpen: boolean) => void;
   setSearchInputTyping: (isSearchInputTyping: boolean) => void;
+  setDefaultVehicle: (defaultVehicle: string) => void;
 }
 
 export const createUiSlice = (set: NamedSet<AppState>, get: GetState<AppState>): UiSlice => ({
   isMenuDrawerOpen: false,
   isProductsDrawerOpen: false,
   isSearchInputTyping: false,
+  defaultVehicle: modes.WALKING,
```
pascaloliv, line 24: I think the `defaultVehicle` should be `modes.DRIVING` instead of `modes.WALKING`

pascaloliv (approved):

PR #1384, opened 2022-07-29
Revamp Github Actions

pascaloliv:
## Description

- [x] Rename some steps + split build steps
- [x] Add coverage 
- [x] Update Quality steps output to display with a Github Markdown output

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,16 +1,19 @@
-name: Erdapfel CI
+name: Qwant Maps - CI
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
-  build:
+  Build:
     runs-on: ubuntu-18.04
 
     steps:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -1,16 +1,19 @@
-name: Erdapfel CI
+name: Qwant Maps - CI
 
-on: [push, pull_request]
```
remi-dupre, line 3: Don't we want to keep the *pull_request* trigger? I think it will run the CI on a merged version of the branch, which is often equivalent to the branch itself but can eventually catch bugs  that would happen once merged on master.

This is not the first time it comes up [this week](https://github.com/Qwant/mimirsbrunn/pull/16#discussion_r929758900), so I'm starting to wonder how relevant my comment is :sweat_smile:

fatal69100 (approved):

PR #1386, opened 2022-08-03
add category airport

fatal69100:
## Description
Add the airport category to display airport category label


remi-dupre (approved):
This needs https://github.com/Qwant/idunn/pull/374 to be deployed, but I think it is not a big deal if looking for "airport" leads to no results instead of bad results in dev for a bit of time :sweat_smile:

PR #1394, opened 2022-08-24
Make front compatible with intention api field

fatal69100:
## Description
API return `intention` instead of `intentions` field. Need to be merge for new Idunn api version that return a `intention` field

## Why
Because there is a maximum of 1 intention return by the back end


xem (approved):
OK, a noter que la mep front devra être synchronisée avec celle de l'API du coup

PR #1397, opened 2022-08-29
QMAPS-2909 fix PT dezoom

xem:
## Description
Map zoom is not computed correctly for some public transport itineraries
This is due to a npm dependency which sometimes reduces the route to a single point.
I copied this npm module in local_modules and made the fix there: ensure that at least 2 points are present.
image


pascaloliv:
@xem - Please, could you keep just the index.js + remove properly the previous dep using `npm uninstall alt-route-labeller` ?

pascaloliv (approved):

PR #1400, opened 2022-09-06
Disable nlu switch and use intention field

fatal69100:
## Description
We use only one intention, use intention instead of "intentions"
Try to remove NLU switch complexity

### DIFF
--- a/tests/integration/tests/autocomplete.js
+++ b/tests/integration/tests/autocomplete.js
@@ -196,26 +196,27 @@ test('check template', async () => {
   expect(lines[3][1]).toEqualCaseInsensitive("Alpes-Maritimes, Provence-Alpes-Côte d'Azur, France");
 });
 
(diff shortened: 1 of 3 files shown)


tests/integration/tests/autocomplete.js
```diff
@@ -196,26 +196,27 @@ test('check template', async () => {
...
+//       document.querySelector(
+//         '.autocomplete_suggestion:first-child .autocomplete_suggestion__first_line'
+//       ).innerText,
+//       document.querySelector('.autocomplete_suggestion:first-child').getAttribute('data-id'),
+//     ];
+//   });
+//
+//   expect(firstLine).toEqual('test result 1');
+//   expect(suggestionId).toEqual('category:restaurant');
+// });
```
remi-dupre, line 219: Shouldn't we just remove this instead of keeping track of a dirty comment?

xem (approved):

PR #1408, opened 2022-10-03
Update lat/lon precision according to the zoom level

fatal69100:
## Description
Update latitude & longitude precision according to the zoom level. More the user zoom, bigger the precision is.

## Why
In order to have closer suggestion result of the zoomed bounding box

### DIFF
--- a/@types/config.ts
+++ b/@types/config.ts
@@ -3,7 +3,7 @@ export interface Geocoder {
   useLang: boolean;
   maxItems: number;
   useFocus: boolean;
-  focusPrecision: string;
+  focusPrecision: Array<[string, string]>;
   focusZoomPrecision: string;
   focusMinZoom: number;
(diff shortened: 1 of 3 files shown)


remi-dupre (changes requested):
Why is it that the day I am picky happen to be the day I'm reviewing Javascript :sob:

@types/config.ts
```diff
@@ -3,9 +3,8 @@ export interface Geocoder {
   useLang: boolean;
   maxItems: number;
   useFocus: boolean;
-  focusPrecision: string;
+  focusPrecision: Array<[string, string]>;
```
remi-dupre, line 6: I think instead of `[string, string]`, the array sould contain an object with a *zoom* and a *precision* field (according to the default in *default_config.yml*).

PR #1434, opened 2022-12-08
add knowledge claims in POI pannel

remi-dupre:
A new `claims` field was recently added in the description block of Idunn which lists a few key informations about a place.

Idunn returns the values in the same format as Knowledge, which may include some markdown links. I wanted to avoid to add a complex dependency for the very basic task of handling those links so I implemented a very naïve parser based on regexps. This sounds like a very ":confused:" idea, so **I'll leave the decision of me changing this up to my dear reviewers**. (Note that on worst case this "parser" will just display URLs with escaped "]" as full text.

Also, some links are filtered out because they contain the sticky param of phoenix, we could make use of this information to redirect to qwant.com in the future and it is still relevant in the API result as we could naturally use it in the instant answer.

image

### DIFF
--- a/src/panel/poi/blocks/Description/index.tsx
+++ b/src/panel/poi/blocks/Description/index.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { components } from 'appTypes/idunn';
+import { parseText } from '../../../../libs/miniMarkdown';
 
 export type PoiDescriptionBlockProps = {
   block: components['schemas']['DescriptionBlock'];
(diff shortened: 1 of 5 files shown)


src/panel/poi/blocks/Description/index.tsx
```diff
@@ -11,20 +11,69 @@ export type PoiDescriptionBlockProps = {
...
+    <>
+      <div className="block-description u-mb-m">
+        {block?.description && <p>{block?.description}</p>}
+        {block?.url && (
+          <a rel="noopener noreferrer" target="_blank" href={block?.url} onClick={onClick}>
+            {texts[block.source] || texts.readMore}
+          </a>
+        )}
+      </div>
+      <ul className="block-description-extra">
```
pascaloliv, line 69: Could you wrap `<ul>` with a `{block.claims?.length && <ul></ul>}` to prevent unnecessary rendering.

src/panel/poi/blocks/Description/index.tsx
```diff
@@ -11,20 +11,69 @@ export type PoiDescriptionBlockProps = {
...
+
+  // Put the elements from texts and links in order to reflect original content.
+  const content = Array.from(Array(texts.length + links.length).keys()).map(i =>
+    // Note that `texts.lengths == links.length + 1` because links are
+    // separators of text sections.
+    i % 2 === 0 ? texts[i / 2] : links[(i - 1) / 2]
+  );
+
+  return <>{content}</>;
+}
```
pascaloliv, line 52: Could you extract the mini-parser in separate file in `src/libs` for example. 
Will be easier to re-use and to test it. And it would be ideal if it were unit tested 👍

remi-dupre: I just updated this, I did 85% of the work then managed to forget about this :sob:

pascaloliv (approved):

PR #1472, opened 2023-11-05
Add Czech language

Fjuro:
## Description
I've made a Czech translation for Qwant Maps.

## Why
Because I think this is the best OSM frontend and I really like it, so I wanted to make it available to more people.

## Screenshots
Currently unable to test locally, sorry.

### DIFF
--- a/language/message/cs.po
+++ b/language/message/cs.po
@@ -1040,7 +1040,7 @@ msgid "In partnership with"
 msgstr "V partnerství s"
 
 msgid "View"
-msgstr "Zobrazit"
+msgstr "Prohlédnout"
 
 msgid "Edit"
 msgstr "Upravit"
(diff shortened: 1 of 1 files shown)


pascaloliv:
First off, big thanks for diving into our code and adding the Czech language file! :rocket: 
However, after careful consideration with the team, we're gonna pass on merging your branch.
Let me explain you why.

Our goal is to maintain a consistent user experience across both websites: Qwant, and Qwant Maps.
The language change mechanism flows from Qwant to Qwant Maps site.
As Czech interface language is not currently supported on Qwant, we believe it's essential to ensure harmony in language availability.

Please understand that this decision is not a reflection of the quality of your work but rather a strategic choice to uphold overall coherence.
Once again, thank you for your valuable contribution.

Fjuro:
Okay, no problem, thanks for the response!

